PR #25, opened 2014-12-12
Add coffeelint to check coffeescript

Tranquility:
Currently it displays the header line of the ouput as first item in the loclist(|| path,lineNumber,lineNumberEnd,level,message). Syntastic probably does something clever to get rid of this line? Otherwise it works as expected.


regedarek:
looks like my coffeelint is triggered without '-f coffeelint.json' with options or I am wrong.

I have file with configuration and without vim I don't see any warnings.

benekastah:
You can add your own args:

```
let g:neomake_coffee_coffeelint_args = ['--reporter=csv', '-f', 'coffeelint.json']
```

PR #34, opened 2015-01-16
add section on maker name conventions

natemara:
Add a paragraph to the readme detailing the naming conventions required for makers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ configuration:
         \ }
     let g:neomake_javascript_enabled_makers = ['jshint']
 
-It is critical that makers follow this naming convention:
+For use with the `:Neomake` command (makers that run on an individual file), it
+is critical that makers follow this naming convention:
 
     g:neomake_{ language }_{ makername }_maker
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,6 +44,15 @@ configuration:
         \ }
     let g:neomake_javascript_enabled_makers = ['jshint']
 
+It is critical that makers follow this naming convention:
```
benekastah: Your addition is accurate if you are using `:Neomake`, but not if you use `:Neomake!`. (As an aside, I think it is unfortunate that the `!` makes such a big difference here, because I think it makes it hard to understand the difference. Probably `:Neomake` should be called `:NeomakeFile` and `:Neomake!` should be called `:Neomake`. But that change would make an annoying and confusing update for current users.)

If you can add a mention of this distinction to your change, I'll merge. I think most people are probably using neomake for this use case, so it's great to call it out better in the readme!

natemara:
I just added a sentance clarifying that this convention is for the `Neomke` command exclusively.

benekastah:
Thanks!

PR #40, opened 2015-01-18
Add flake8 maker for python files

natemara:
It seemed like there weren't enough Python makers, so I thought I'd add a maker for flake8.

### DIFF
--- a/autoload/neomake/makers/python.vim
+++ b/autoload/neomake/makers/python.vim
@@ -7,11 +7,11 @@ function! neomake#makers#python#EnabledMakers()
 
     let makers = ['python']
     if neomake#utils#Exists('flake8')
-        add(makers, 'flake8')
+        call add(makers, 'flake8')
     else
-        extend(makers, ['pep8', 'pyflakes'])
+        call extend(makers, ['pep8', 'pyflakes'])
     endif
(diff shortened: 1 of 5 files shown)


autoload/neomake/makers/python.vim
```diff
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#python#EnabledMakers()
-    return ['python', 'pep8', 'pyflakes', 'pylint']
+    return ['python', 'pep8', 'pyflakes', 'pylint', 'flake8']
```
benekastah: Since flake8 runs pep8 and pyflakes, it seems that if all three are in the list of enabled makers we're going to get some redundant runs. It's more complicated, but I think it would be better if we did something like this:

```
if exists('s:python_makers')
    return s:python_makers
endif
let makers = ['python']
if neomake#utils#Exists('flake8')
    add(makers, 'flake8')
else
    extend(makers, ['pep8', 'pyflakes'])
endif
add(makers, 'pylint')  " Last because it is the slowest
let s:python_makers = makers
return makers
```

natemara:
I changed the python enabled makers function to the one that you gave, and added the `shellcheck

benekastah:
Great, thanks!

PR #41, opened 2015-01-19
add gcc/g++ makers for c/c++ files

natemara:
Not everybody has clang. Since clang is a bit faster (I think) clang is the prefered maker, but I added gcc as a fallback maker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,12 @@ Ruby:
 C:
 
 - clang
+- gcc
 
 C++:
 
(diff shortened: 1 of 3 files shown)


natemara:
I guess I was just trying to leave room for more makers that did more than compile, like static analysis tools. I updated those functions to that pattern for now.

benekastah:
This is great! Keep 'em coming!

PR #45, opened 2015-01-19
Add Frosted maker for python.

renstrom:
Mostly taken from Syntactic, but prepended an extra row to `errorformat` to try and correctly identify warnings vs. errors and tag accordingly.

However I'm not sure where to place it in the list of checkers.

>Frosted is a fork of pyflakes (originally created by Phil Frost) that aims at more open contribution from the outside public, a smaller more maintainable code base, and a better Python checker for all.

https://github.com/timothycrosley/frosted

### DIFF
--- a/autoload/neomake/makers/python.vim
+++ b/autoload/neomake/makers/python.vim
@@ -5,11 +5,11 @@ function! neomake#makers#python#EnabledMakers()
         return s:python_makers
     endif
 
-    let makers = ['python']
+    let makers = ['python', 'frosted']
     if neomake#utils#Exists('flake8')
         call add(makers, 'flake8')
     else
-        call extend(makers, ['pep8', 'frosted', 'pyflakes'])
+        call extend(makers, ['pep8', 'pyflakes'])
     endif
     call add(makers, 'pylint')  " Last because it is the slowest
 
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/python.vim
```diff
@@ -75,3 +75,17 @@ function! neomake#makers#python#python()
...
         \ }
 endfunction
+
+function! neomake#makers#python#frosted()
+    return {
+        \ 'args': [
+            \ '-vb'
+        \ ],
+        \ 'errorformat':
+            \ '%f:%l:%c:%t%n:%s:%m,' .
```
renstrom: This is the row I added.

autoload/neomake/makers/python.vim
```diff
@@ -9,7 +9,7 @@ function! neomake#makers#python#EnabledMakers()
     if neomake#utils#Exists('flake8')
         call add(makers, 'flake8')
     else
-        call extend(makers, ['pep8', 'pyflakes'])
+        call extend(makers, ['pep8', 'frosted', 'pyflakes'])
     endif
     call add(makers, 'pylint')  " Last because it is the slowest
```
benekastah: I think we could change this line to:

```
call extend(makers, ['frosted', 'pylint'])  " Pylint is last because it is the slowest
```

And remove frosted above. The reason flake8 has that conditional craziness is because it has pep8 and pyflakes as dependencies, and adding them all into the list would result in pep8 and pyflakes running twice for no reason if flake8 was installed. Unless flake8 might run frosted (I'm guessing it won't since it relies on pyflakes), I think it can simply go in the list.

renstrom:
Updated the `makers` list according to your suggestions.

benekastah:
Thanks!

Issue #47, opened 2015-01-19
Find a way to share makers for :Neomake! like we can share makers for :Neomake


benekastah:
#43 brought this up. Currently filetype makers are found in autoload/neomake/makers/{filetype}.vim. This is nice because we can have makers for many different filetypes that get loaded on demand as more filetypes are used. Possibly we need to put these in autoload/neomake/filetype_makers/{filetype}.vim. Then we can have non-filetype makers in autoload/neomake/makers/{tool}.vim.

benekastah:
Fixed by b19be85e321478faf15608922f43dbf1604ae463 and 293e2c8adc36bc2bb8d078395eb72d9fbf4b2af9

Issue #46, opened 2015-01-19
:Neomake! should be able to set signs and echo current error like :Neomake


benekastah:
#43 brought this up

maurizi:
Wait, has this really been fixed?  It looks like it hasn't been yet...

PR #49, opened 2015-01-20
add maker for rust and fix gcc maker exe

natemara:
The `gcc` maker had its exe field set to `clang`. This PR fixes that.

I also added a basic maker for the Rust programming language.

### DIFF
--- a/autoload/neomake/makers/c.vim
+++ b/autoload/neomake/makers/c.vim
@@ -25,7 +25,7 @@ endfunction
 
 function! neomake#makers#c#gcc()
     return {
-        \ 'exe': 'clang',
+        \ 'exe': 'gcc',
         \ 'args': ['-fsyntax-only'],
         \ 'errorformat':
             \ '%-G%f:%s:,' .
--- a/autoload/neomake/makers/c.vim
+++ b/autoload/neomake/makers/c.vim
@@ -10,7 +10,6 @@ endfunction
 
 function! neomake#makers#c#clang()
     return {
-        \ 'exe': 'clang',
         \ 'args': ['-fsyntax-only'],
         \ 'errorformat':
             \ '%-G%f:%s:,' .
@@ -25,7 +24,6 @@ endfunction
 
 function! neomake#makers#c#gcc()
     return {
-        \ 'exe': 'gcc',
         \ 'args': ['-fsyntax-only'],
         \ 'errorformat':
             \ '%-G%f:%s:,' .
--- a/autoload/neomake/makers/rust.vim
+++ b/autoload/neomake/makers/rust.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#rust#EnabledMakers()
-    return ['cargo']
+    return ['rustc']
 endfunction
 
 function! neomake#makers#rust#cargo()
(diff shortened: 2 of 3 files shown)


autoload/neomake/makers/c.vim
```diff
@@ -25,7 +25,7 @@ endfunction
 
 function! neomake#makers#c#gcc()
     return {
-        \ 'exe': 'clang',
+        \ 'exe': 'gcc',
```
benekastah: Actually the exe field can be omitted when it matches the maker name.

natemara:
There you go

autoload/neomake/makers/rust.vim
```diff
@@ -0,0 +1,19 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#rust#EnabledMakers()
+    return ['rustc']
+endfunction
+
+function! neomake#makers#rust#rustc()
+    return {
+        \ 'args': ['-o /dev/null'],
```
benekastah: Each arg should take its own entry in the list. Also, /dev/null is env specific, so we have a utility for that. Here is a corrected version:

```
\ 'args': ['-o', neomake#utils#DevNull()],
```

natemara:
Fixed the `rustc` maker, added a `cargo` maker. Had a question about this though. `cargo` is a project/package manager, and is to `rustc

benekastah:
Good question. There isn't currently a good place to put project-level makers. Issue #47 was opened to address this. Your cargo maker will actually work if you do this:

```
let g:neomake_cargo_maker = neomake#makers#rust#cargo()
```

Then `:Neomake! cargo` will run `cargo build` and put the results in the quickfix list. This isn't quite as good as the file-level makers yet. See #46 for that.

For now, perhaps you could add a TODO note referencing #47 by the cargo maker and leave it out of the enabled makers list.

natemara:
how's this?

natemara:
This actually doesn't work. When I run `cargo build`, it only runs `cargo`.

natemara:
Okay, I got it working by putting `let g:neomake_cargo_maker = neomake#makers#rust#cargo()` in my nvimrc, but the cargo maker doesn't populate the location list. Is there a bigger difference between `Neomake` and `Neomake!`?

benekastah:
I should have been more clear. The `g:neomake_cargo_maker` shouldn't be defined in any file in the autoload directory, since it may not be available (depending on whether that code has loaded). I was imagining individual users would configure that themselves until we make project-level makers easier to share.

`:Neomake!` is different from `:Neomake` in that it doesn't pass the current file as the last argument. Additionally, it puts its results in the quickfix list, not the location list. The reason for this is that the location list is associated with a window (a better fit when getting info on an individual file), whereas the quickfix list is global (better when running something at the project level). `:cwindow` should show you what you're looking for.

natemara:
`cwindow` does show the compiler output, but the lines with errors aren't highlighed. Is this to be expected?

benekastah:
Yup, #46 is tracking that issue. Once you remove the g:neomake_cargo_maker bit from this PR, I'll happily merge. Thanks for doing this!

natemara:
done.

benekastah:
#47 is fixed now. `let g:neomake_cargo_maker = neomake#makers#rust#cargo()` will no longer work. Instead, you can just do `:Neomake! cargo`. You can also do:

```
let g:neomake_enabled_makers = ['cargo']
```

Then, just running `:Neomake!` will run cargo for you.

PR #50, opened 2015-01-20
add some checkers for tex

natemara:
I added `lacheck` and `chktex` makers for checking tex/latex files.

### DIFF
--- a/autoload/neomake/makers/tex.vim
+++ b/autoload/neomake/makers/tex.vim
@@ -1,11 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#tex#EnabledMakers()
-    if neomake#utils#Exists('chktex')
-        return ['chktex']
-    else
-        return ['lacheck']
-    end
+    return ['chktex', 'lacheck']
 endfunction
 
 function! neomake#makers#tex#chktex()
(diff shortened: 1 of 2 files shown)


autoload/neomake/makers/tex.vim
```diff
@@ -0,0 +1,24 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#tex#EnabledMakers()
+    return ['chktex']
```
benekastah: Is there a reason lacheck is missing from this list?

natemara:
better?

autoload/neomake/makers/tex.vim
```diff
@@ -0,0 +1,28 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#tex#EnabledMakers()
+    if neomake#utils#Exists('chktex')
```
benekastah: This would work but probably isn't necessary (unless I'm missing something). Makers that fail the `neomake#utils#Exists` check are filtered out before they are invoked. The only reason we had this function in the `neomake#makers#python#EnabledMakers` function is just because flake8 depended on pep8 and pyflakes, which would have resulted in redundant runs.

Normally, all that is required is to return a straight list of all the checkers that are defined, and only the makers the user has installed will run.

benekastah:
Thanks!

PR #54, opened 2015-02-19
Add ruby checker to neomake

cartolari:
This add the ruby checker the way syntastic checks, so if you do not have rubocop you can at least have basic checks from the ruby interpreter.
The syntax was not build by me, I borrowed from syntastic.


benekastah:
Thanks!

PR #57, opened 2015-03-11
Use Vim builtin 'strftime' to create timestamps in logs.

Neki:
Just stumbled upon the TODO while reading the code, so here is a solution.

Disclaimer: I have not actually tested the code on other platforms, but it did not break anything on my Linux and according to https://msdn.microsoft.com/en-us/library/fe06s4ak.aspx all the format characters used here are supported on Windows.

Edit: just realized the call to `cat` below is not cross-platform, but I am not aware of an easy way in Vimscript to append a string to a file (without reading the entire file first).


benekastah:
Thanks. Every little step toward cross-platform is a good one!

PR #59, opened 2015-03-12
Do not use 'cat' to create a temp file.

Neki:
Just found another "TODO Make this cross platform". 

I am a Vimscript newbie, so I am not sure why this trick with `cat` was used in the first place. My tests did not reveal a difference between the two versions.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -207,8 +207,9 @@ function! neomake#Make(options) abort
                 let tempfile = 1
                 let tempsuffix = '.'.neomake#utils#Random().'.neomake.tmp'
                 let makepath .= tempsuffix
-                " TODO Make this cross platform
-                silent exe 'w !cat > '.shellescape(makepath)
+                let escapedpath = fnameescape(makepath)
+                noautocmd silent exe 'keepalt w! '.escapedpath
+                noautocmd silent exe 'bwipeout '.escapedpath
                 call neomake#utils#LoudMessage('Neomake: wrote temp file '.makepath)
             endif
         endif


benekastah:
Sorry, I should have left a comment for why this `cat` trick was used. The code actually used to be like your change. The reason I didn't keep it that way was that it adds all the temporary files to your buffer list, which messes up things like toggling between two buffers with `CTRL-6` or `:b#`.

Neki:
Thanks for the explanation, I appreciate it (I have much to learn regarding Vim).

Would something like this work, or are there other side effects I am not aware of?

```vim
let escapedpath = fnameescape(makepath)
silent exe 'keepalt w! '.escapedpath
bwipeout escapedpath
```

benekastah:
Hm, didn't know about `keepalt`. Nice. That seems like it should work. If you test it out and make sure everything is in order, I'd be happy to merge that.

Neki:
After this change the buffer list behaves correctly, however a major difference is that events such as BufAdd are now fired whenever a temporary file is created. It doesn't seem to be an issue with my configuration and set of plugins, but maybe there is a scenario where it would create problems (I can't think of one right now).

This needs more testing.

benekastah:
Ah, yes. I think the autocommands will cause problems (because now that you mention it, I remember having problems with those too). However, we could use `eventignore`:

```
let oldEventignore = &eventignore
let escapedpath = fnameescape(makepath)
set eventignore=all
silent exe 'keepalt w! '.escapedpath
bwipeout escapedpath
let &eventignore = oldEventignore
```

 I haven't tested this, but it *should* work.

Neki:
Thanks a lot for your feedback and helping me on this PR! 

Your solution seems to work. A couple of hours using this branch in a real coding session did not reveal any issue.

I have squashed and rebased the commits.

benekastah:
danielr-minted

benekastah:
Good find on `noautocmd`. Even better.

Issue #31, opened 2015-01-11
g:neomake_make_modified sometimes leaves temp files laying around


benekastah:
Most of the time it cleans them up, but sometimes leaves them around. Probably need a good way to log errors to a file before tracking this down.

Neki:
I found another situation that can trigger this bug: calling `:Neomake` (on a modified buffer), then quickly after `:wq` before the maker is finished.

benekastah:
Ah, good point. We should also clean these files up in a `VimLeave` autocmd.

benekastah:
Will close as wontfix if/when #84 gets merged.

benekastah:
The make_modified feature was removed, so this issue is now invalid.

PR #62, opened 2015-03-14
Delete the temporary file even if the working directory has changed.

Neki:
Referring to temporary files by their full, expanded path allows neomake to still find the temporary file if the working directory changes between the moment where the temp file is created and the moment it should be deleted.

This is a corner case, but as I am testing #59 I am forcing neomake to write a temporary file each time it performs a check, and in combination with the plugin [vimfiler](https://github.com/Shougo/vimfiler.vim) which temporarily changes the working directory when performing a rename, I have temp files laying around after a rename.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -202,7 +202,7 @@ function! neomake#Make(options) abort
         let tempsuffix = ''
         let makepath = ''
         if file_mode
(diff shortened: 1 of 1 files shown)


benekastah:
Awesome, maybe this fixes #31.

PR #63, opened 2015-03-17
Wait for the next maker first error (or end) to remove signs.

Neki:
Some makers (e.g. rubocop running on JRuby) may take a few seconds to return the first error (in this case, because of the JVM startup time). Neomake currently removes all signs and errors as soon as `:Neomake` is called, but this is a problem with makers that take a while to start.

For instance, I have bound `:Neomake` to `Bufwrite`. Let's assume that a maker gave two errors, one on line 1 and one on line 2. If I fix the first error, then (out of habit) type `:w`, I have to wait a few seconds before the error on line 2 appears again.

With this PR, signs and errors are not cleared when `:Neomake` is called, but when the first maker starts returning an error or is finished. This helped my workflow with JRuby, and may be useful when #46 is implemented, as building/checking large projects can take some time before the current buffer is processed.

However, this could be an option rather than the default (or may be a bad idea in general :) ). What is your opinion about it?

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -187,13 +187,11 @@ function! neomake#Make(options) abort
 
     if file_mode
         let b:neomake_loclist_nr = 0
-        let b:neomake_errors = {}
-        " Remove any signs we placed before
-        let b:neomake_signs = get(b:, 'neomake_signs', {})
-        for ln in keys(b:neomake_signs)
-            exe 'sign unplace '.b:neomake_signs[ln]
-        endfor
-        let b:neomake_signs = {}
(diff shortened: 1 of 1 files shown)


benekastah:
LGTM. Thanks!

Neki:
Not sure why yet, but sometimes I have errors now that this PR is merged. You may want to revert the merge until I find the time to investigate. Sorry :(

Neki:
And of course, now that I have some time to work on this, I can not reproduce the issue (it was something about `b:neomake_signs_cleared` not being defined - I was in a hurry and did not note the exact message).

Please notify me if you encounter an issue with this PR, I will try to fix it.

PR #65, opened 2015-03-19
Check that b:neomake_signs_cleared exists before reading it.

### LINKED ISSUE #64: Error about undefined variable b:neomake_signs_cleared
I just installed `Neomake` and when I run `:silent Neomake!<CR>` I get this error:

shot-001

Neki:
`b:neomake_signs_cleared` is set in the buffer where `:Neomake` is called, but if a maker returns
errors in another buffer, then this variable won't be defined.

Should fix #64.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -297,7 +297,7 @@ function! s:AddExprCallback(maker) abort
 
             " On the first valid error identified by a maker,
             " clear the existing signs in the buffer
-            if !b:neomake_signs_cleared
+            if !exists("b:neomake_signs_cleared") || !b:neomake_signs_cleared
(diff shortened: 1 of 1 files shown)


benekastah:
Thanks for figuring this out! By the way, I went ahead and made you a contributor. If for some reason you'd rather not be, just ask me and I'll undo that. Thanks again!

Neki:
Fixing the bugs I created myself seems normal to me :)

I still think there may be a deeper issue, because I think I encountered this bug even with a maker that to the best of my knowledge emit errors only for the current file (rubocop). But I can not reproduce it.

Thanks for the access to the repo! However, I don't promise anything about the frequency of my contributions.

benekastah:
That's fine :)

PR #69, opened 2015-03-26
Remove flaky error format from Frosted (Py) checker.

renstrom:
This line broke "jump-to-error-line" from the location window. This is something I added in https://github.com/benekastah/neomake/pull/45 which wasn't included by Syntastic, I thought it would produce a bit more exhaustive match with error type (`%t`) and error number (`%n`), but it seems like it broke more than it added.


benekastah:
Thanks!

PR #70, opened 2015-03-26
Add vim-airline extension.

renstrom:
Add statusline extension for vim-airline. The lines got a bit long, do you want me to wrap them?

It's that `E:4` to the right:
screen shot 2015-03-26 at 13 28 11

### DIFF
--- a/autoload/airline/extensions/neomake.vim
+++ b/autoload/airline/extensions/neomake.vim
@@ -1,3 +1,9 @@
+" vim: ts=4 sw=4 et
+
+if exists('g:neomake_airline') && g:neomake_airline == 0
+    finish
+endif
+
 let s:spc = g:airline_symbols.space
 
 function! airline#extensions#neomake#apply(...)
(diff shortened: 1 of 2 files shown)


benekastah:
This looks fine, but I'd rather we wrap these methods in a neomake setting to prevent adding things to people's statusline who don't want it. Something like:

```
function! airline#extensions#neomake#apply(...) abort
    if (get(g:, 'neomake_airline', 0)
        " All the other stuff here
    endif
endfunction
```

Same with the other function. Also, please copy the modeline from another file into this one. Thanks!

renstrom:
Sure thing. Do you want the option enabled or disabled by default?

benekastah:
My preference is to be pretty conservative with things that are done automatically. However, this feature would be difficult to discover if it was disabled by default. I'll let that be your call. I'm not an airline user myself.

renstrom:
Right, I'll probably go with enabled by default since that's the behaviour you'd get with stock Syntastic and Airline.

Added an option now with some docs, not sure if it's enough though. Also, from what I can tell the `errorformat` must contain `%t` (and possibly others) to populate `neomake#statusline#LoclistStatus()`...?

benekastah:
Yeah, currently `neomake#statusline#LoclistStatus()` isn't very sophisticated. Feel free to improve it if you want something more about it.

Issue #60, opened 2015-03-13
AddExprCallback: not possible to change sign neomake_err


Neki:
I am currently working on a Ruby file (checkers : mri + rubocop on JRuby). If I call `:Neomake` on this file, I get the error:

```
Error detected while processing function neomake#ProcessCurrentBuffer..<SNR>125_ProcessJobOutput..<SNR>125_AddExprCallback:
line   38:
E885: Not possible to change sign neomake_err
```

I then get this error every time I enter the window displaying my buffer. If I restart Vim (without loading a session), and call `:Neomake` again, I get the exact same error.

The corresponding part of the logfile generated by neomake:
```
2015-03-13T21:45:32+0100 Log level 2: Neomake: Starting: ruby -c -T1 -w lib/remotebox/core/world.rb
2015-03-13T21:45:32+0100 Log level 2: Neomake: Starting: rubocop --format emacs lib/remotebox/core/world.rb
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri stderr: SyntaxError in lib/remotebox/core/world.rb:73: syntax error, unexpected kRESCUE
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri stderr:         rescue UnsupportedSystem => e
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri stderr:               ^
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri stderr done.
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri processing 3 lines of output
2015-03-13T21:45:36+0100 Log level 3: Neomake: executing: sign place buffer=5
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri stderr: 
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri stderr: 
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri stderr done.
2015-03-13T21:45:36+0100 Log level 3: Neomake: mri processing 2 lines of output
2015-03-13T21:45:36+0100 Log level 1: Neomake: mri complete
2015-03-13T21:45:41+0100 Log level 3: Neomake: rubocop stdout: /media/docs/projects/remotebox/lib/remotebox/core/world.rb:73:9: E: unexpected token kRESCUE
2015-03-13T21:45:41+0100 Log level 3: Neomake: rubocop stdout: 
2015-03-13T21:45:41+0100 Log level 3: Neomake: rubocop stdout done.
[...]

Issue #48, opened 2015-01-19
Processing lint results (placing signs) takes a long time on large files


benekastah:
I don't know exactly what is slow, so I'll just write what I did so that it will be easy to reproduce later:

```
let g:neomake_c_nvimlint_maker = {
    \ 'exe': './clint.py',
    \ 'errorformat': '%f:%l:  %m [%n]'
    \ }
```

Then edit [eval.c](https://github.com/neovim/neovim/blob/master/src/nvim/eval.c) and run `:Neomake nvimlint`. It gave over 900 results for me and the file is close to 20000 loc. It seems like when I disabled placing of signs with `let g:neomake_place_signs = 0` that it helped a lot. I'm not sure if that's because the file is too large or if my sign placing routine is just too slow for 900 entries.

One idea: Only process the results that are correspond with the visible part of the buffer.

benekastah:
Another idea that may speed things up is to place all the signs as they come, but only move them to the top after the maker is complete. I suspect the moving signs to the top part is the expensive part.

Neki:
Even on small files (~ 100 lines), processing results when there is more than 10 errors has a noticeable delay (between half a second and a second) on my computer (neovim is freezed during the processing). I have started to play with this issue, so far it seems the best gain I was able to achieve was by removing the `redraw!` command at https://github.com/benekastah/neomake/blob/master/autoload/neomake.vim#L338 (but performance is still poor on `eval.c`). What is the goal of this command? Signs appear as normal when it is removed. 

Removing the "move signs on top" routine did not speed things up noticeably.

Then again I did not benchmark the improvements or anything, so take this with a grain of salt.

benekastah:
It's possible that the `redraw!` is no longer needed with recent versions of neovim. When I coded that stuff up at first I would get odd results without it.

There are a lot of issues with the sign placing code (as you can see by the issues list). I'm planning to refactor that stuff this weekend where I hope to knock out a bunch of the major issues. So keep that in mind as you investigate that code and don't sink too much time into it because it will change. I'll try to get that done quickly so that it doesn't block you on looking into issues when you want to.

Neki:
Ok, thanks for the info. There is no need to rush as far as I am concerned, I have plenty of other things to do :)

benekastah:
I said I fixed this in #77, but I'd like to test it again before closing.

Neki:
Some tests on my laptop:

[...]

PR #77, opened 2015-03-31
Signs

### LINKED ISSUE #46: :Neomake! should be able to set signs and echo current error like :Neomake

### LINKED ISSUE #61: If a file-level maker emits errors for a different file, the signs are placed in the other files but never cleaned up
This happens with the clang makers. Should fix at the same time as #46.

benekastah:
Fixes #61, #48, #60

Sets the stage to fix #46. Not sure why it isn't working at the moment, but I wanted to get this merge in because a lot has changed and it seems like nothing new is breaking (after using it personally for about a week).


autoload/neomake/signs.vim
```diff
@@ -0,0 +1,188 @@
...
+    let botline = line('w$')
+    for ln in range(topline, botline)
+        if has_key(s:sign_queue[buf], ln)
+            if !exists('l:signs')
+                let l:signs = neomake#signs#GetSignsInBuffer(buf)
+            endif
+            call neomake#signs#PlaceSign(l:signs, s:sign_queue[buf][ln])
+            unlet s:sign_queue[buf][ln]
+        endif
+    endfor
```
Neki: if `s:sign_queue[buf]` is empty at this point, maybe the key `buf` could be removed from `s:sign_queue` so next calls to this function return faster.

benekastah: Good idea. Done and merging.

PR #80, opened 2015-04-01
Create NeomakeSh to allow executing of shell commands

ViViDboarder:
Create new function and command to allow execution of arbitrary shell
commands by using `:NeomakeSh echo foo bar` the same way one would use
`! echo foo bar` or `:Dispatch echo foo bar`

The function can also be used by other plugins to make leveraging
jobcontrol easy

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -181,13 +181,7 @@ function! neomake#Make(options, ...) abort
     let file_mode = get(a:options, 'file_mode')
 
     let enabled_makers = get(a:options, 'enabled_makers', [])
-    let sh_command = get(a:options, 'sh_command', '')
-    if len(sh_command)
-        let custom_maker = neomake#utils#MakerFromCommand(&shell, sh_command)
-        let custom_maker.name = 'sh: '.sh_command
-        let enabled_makers =  [custom_maker]
-    endif
-    if !len(enabled_makers) 
+    if !len(enabled_makers)
         if file_mode
             call neomake#utils#DebugMessage('Nothing to make: no enabled makers')
             return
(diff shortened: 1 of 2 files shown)


plugin/neomake.vim
```diff
@@ -18,11 +18,17 @@ function! s:NeomakeCommand(file_mode, enabled_makers)
...
+function! s:NeomakeSh(...)
+    call neomake#Make({'sh_command': join(a:000, ' ')})
+endfunction
+
 command! -nargs=* -bang Neomake call s:NeomakeCommand('<bang>' !=# '!', [<f-args>])
 " These commands are available for clarity
 command! -nargs=* NeomakeProject Neomake! <args>
 command! -nargs=* NeomakeFile Neomake <args>
 
+command! -nargs=+ NeomakeSh call s:NeomakeSh(<f-args>)
```
benekastah: I think instead of `<f-args>` you want `<q-args>`, which will just deliver a single string, exactly as they entered it. Since there isn't a lot to the `s:NeomakeSh` function, you could replace this line with:

```
command! -nargs=+ NeomakeSh call neomake#Make({'sh_command': <q-args>})
```

autoload/neomake.vim
```diff
@@ -179,7 +181,13 @@ function! neomake#Make(options, ...) abort
     let file_mode = get(a:options, 'file_mode')
 
     let enabled_makers = get(a:options, 'enabled_makers', [])
-    if !len(enabled_makers)
+    let sh_command = get(a:options, 'sh_command', '')
+    if len(sh_command)
```
benekastah: Instead of having a `sh_command` option, I think we should move the stuff inside this block into `s:NeomakeSh` in the other file.

ViViDboarder:
Good call. That looks much cleaner.

benekastah:
Thanks for doing this!

PR #87, opened 2015-04-08
checker for Sass

alxndr:
[Sass](http://sass-lang.com/) is a langage that extends and compiles into [CSS](http://en.wikipedia.org/wiki/Cascading_Style_Sheets); [`scss-lint`](https://github.com/brigade/scss-lint) is a linter for Sass.

### DIFF
--- a/autoload/neomake/makers/ft/scss.vim
+++ b/autoload/neomake/makers/ft/scss.vim
@@ -1,0 +1,12 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#scss#EnabledMakers()
+    return ['scsslint']
+endfunction
+
+function! neomake#makers#ft#scss#scsslint()
+    return {
+        \ 'exe': 'scss-lint',
+        \ 'errorformat': '%f:%l [%t] %m'
+    \ }
+endfunction


autoload/neomake/makers/ft/scss.vim
```diff
@@ -0,0 +1,13 @@
...
+
+function! neomake#makers#ft#scss#EnabledMakers()
+    return ['scsslint']
+endfunction
+
+function! neomake#makers#ft#scss#scsslint()
+    return {
+        \ 'exe': 'scss-lint',
+        \ 'errorformat': '%f:%l [%t] %m',
+        \ 'returns': [0, 1, 2, 64, 66, 70, 78]
```
benekastah: I'm guessing this was copied from syntastic, but Neomake doesn't use this attribute.

benekastah:
Does this also work on sass files?

autoload/neomake/makers/ft/scss.vim
alxndr: Ah, right you are, thanks

alxndr:
I don't think so; the `scss-lint` project seems to explicitly only target .css and .scss files.

benekastah:
Thanks!

indiesquidge:
Thanks for the PR, @alxndr. I'm a bit confused why this PR and commit is labeled as a checker for SASS. As is discussed here, it only works for CSS and SCSS files. Do you know of any linters for SASS files that may be used?

alxndr:
@indiesquidge do you mean files which have a `.sass` extension? I think I've only ever used `.scss`.

indiesquidge:
@alxndr yeah, `.sass` files. I am working on a project right now that uses SASS, and last night I got tripped up with a ridiculous whitespace error that would cause my build process to seg fault with no information. I learned that it was because of the `gulp` libraries we are using to handle our SASS was silently blowing up because I had written some SASS incorrectly. While a build task shouldn't simply seg fault, I feel like problems such as this would be easier to deal with if I can just catch them with a linter while editing the file (well, in my opinion, the real solution would be to just use SCSS, ha!)

I tried mimicking your PR here with something like the following

```vim
" vim: ts=4 sw=4 et

function! neomake#makers#ft#sass#EnabledMakers()
    return ['sasslint']
endfunction

function! neomake#makers#ft#sass#sasslint()
    return {
        \ 'exe': 'sass-lint',
        \ 'errorformat': '%f:%l [%t] %m'
    \ }
endfunction
```

[`sass-lint`](https://github.com/sasstools/sass-lint) is an NPM library that has a CLI interface, so I figured it would work nicely. However, I can't seem to get it to work properly. I would love some help if anyone can offer it.

PR #91, opened 2015-04-10
add dmd checker for dlang

rsw0x:


autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,50 @@
...
+    let s:d_makers = makers
+    return makers
+endfunction
+
+function! s:UpdateDub()
+    "Add dub directories
+    let l:tmp_file = findfile("dub.json", ".;")
+    if neomake#utils#Exists('dub') && !empty(l:tmp_file)
+        let l:tmp_dir = fnamemodify(l:tmp_file,':p:h')
+        let l:dub = eval(substitute(system("dub describe --annotate --root=" . tmp_dir)
```
benekastah: This statement is confusing. What does this accomplish?

benekastah:
I haven't worked a lot with d, but couldn't you just use `dub` as a maker instead? Then it would resolve all the includes itself and pass that along to `dmd`. The `dmd` checker could run in syntax only mode (which I'm assuming it has. We do this for `clang`) to avoid the complexity of having to resolve includes ourselves.

rsw0x:
Dub isn't distributed with D and not everyone uses it, it also requires that you're editing a dub project(i.e, the D standard library cannot be compiled with dub.)

AFAIK what I did was the canonical way to extract includes from dub - the dub describe command dumps it out in json format which I treat as a dictionary.

benekastah:
Ok, so if a project isn't using dub, can you still use dub to extract includes for it? If so, then I could see something like this being necessary. But if not, then I think we should just have a dub maker and dmd maker, where the dmd maker only checks syntax or something (if possible).

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,50 @@
...
+                call add(tmp_arr, l:path . l:importPath)
+            endfor
+        endfor
+        let s:neomake_d_dub_include_dirs = join(tmp_arr, ":")
+    endif
+endfunction
+
+function! neomake#makers#ft#d#dmd()
+    let l:args = ['-c', '-o-', '-I' . s:neomake_d_dub_include_dirs]
+    if exists("g:neomake_d_dmd_conf")
```
benekastah: It would be good to add this to the docs.

rsw0x:
dmd doesn't have a syntax only mode like rustc, so it would heavily error out without the includes if combined with dub. Is it acceptable to use an autocmd in the maker to check if a dub project is being edited on the BufEnter event, else fall back to a regular D compiler?

benekastah:
Could we just add a dub maker and do something like this?

```
function! neomake#makers#ft#d#EnabledMakers() abort
    if neomake#utils#Exists('dub') && !empty(l:tmp_file)
        return ['dub']
    else
        return ['dmd']
    endif
endfunction
```

Also, probably a dmd maker isn't useful on its own, since it will complain about missing includes for most people, right? I'm guessing most people use something like make to pass all the arguments to dmd, so probably we need an easier way to configure the program that gets executed by a given maker.

rsw0x:
Yep, that's pretty much what I was going to change it to but I have to be able to detect the root project folder containing the dub project file because it doesn't find it on its own, plus I'd have to detect if a new dub project has been loaded.

Dub doesn't operate on a per-file basis like cargo, and AFAICT can't be put in the filetype folder. Does the user have to manually enable the dub checker?

benekastah:
Ah, that's right. It seems like your approach is about right then. I have a couple more comments to make in source and then we can tie this up :)

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,50 @@
...
+        let l:dub = eval(substitute(system("dub describe --annotate --root=" . tmp_dir)
+                    \ ,'\v[\n\r]','','g'))
+        let l:tmp_arr = []
+        for l:package in l:dub.packages
+            let l:path = l:package.path
+            for l:importPath in l:package.importPaths
+                call add(tmp_arr, l:path . l:importPath)
+            endfor
+        endfor
+        let s:neomake_d_dub_include_dirs = join(tmp_arr, ":")
```
benekastah: Could this value be returned instead? Then we could call this function in `neomake#makers#ft#d#dmd` to get a fresh result every time we build a new dmd maker (which is every time the user runs `:Neomake`).

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,50 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#d#EnabledMakers()
```
benekastah: If it makes sense to follow my comment in `s:UpdateDub`, this function could just be:

```
function! neomake#makers#ft#d#EnabledMakers()
    return ['dmd']
endfunction
```

benekastah:
Oh, and yes the user would have to manually enable the `dub` checker if it existed, either by running `:Neomake! dub` or doing `:let g:neomake_enabled_makers = ['dub']` and then `:Neomake!` after that.

rsw0x:
Added in some statements to check for the GNU D compiler and the LLVM D compiler, some linux distros don't supply dmd in their repos(i.e, Fedora)

I'll get the docs added tomorrow probably.

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,55 @@
...
+    "dmd, ldmd, and gdmd all share a common CLI.
+    "Ordered in efficiency of compiler
+    let l:makers = []
+    if neomake#utils#Exists('dmd')
+        call add(l:makers, 'dmd')
+    elseif neomake#utils#Exists('ldmd')
+        call add(l:makers, 'ldmd')
```
benekastah: For ldmd and gdmd you're going to have to add functions like:

```
function! neomake#makers#ft#d#ldmd()
    return neomake#makers#ft#d#dmd()
endfunction

function! neomake#makers#ft#d#gdmd()
    return neomake#makers#ft#d#dmd()
endfunction
```

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,55 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#d#EnabledMakers()
+    "dmd, ldmd, and gdmd all share a common CLI.
+    "Ordered in efficiency of compiler
+    let l:makers = []
+    if neomake#utils#Exists('dmd')
```
benekastah: You don't have to do exists checks yourself. Neomake checks that the program exists before it starts the job. This function could just be: `return ['dmd', 'ldmd', 'gdmd']`

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,55 @@
...
+                call add(tmp_arr, l:path . l:importPath)
+            endfor
+        endfor
+        let l:dub_incs = join(tmp_arr, ":")
+    endif
+    return l:dub_incs
+endfunction
+
+function! neomake#makers#ft#d#dmd()
+    let l:args = ['-c', '-o-', s:UpdateDub()]
```
benekastah: if you're using `s:UpdateDub()` more than once, it makes sense to save the result in a local variable, especially since it results in two system calls (one for the exists check on 'dub' and the other to get the includes).

autoload/neomake/makers/ft/d.vim
rsw0x: Wouldn't all three run on a make then?
They all use the same frontend code so it offers no benefit to run more than one.

benekastah: Ah, fair enough.

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,62 @@
...
+    "Updating dub paths each make might be slow?
```
benekastah: If you are noticing latency here, then I think caching is the way to go (as you were doing before). But if it's fast enough for you, then getting fresh results is probably better.

autoload/neomake/makers/ft/d.vim
```diff
@@ -0,0 +1,63 @@
...
+        endfor
+        let l:dub_incs .= join(tmp_arr, ":")
+    endif
+    return l:dub_incs
+endfunction
+
+function! neomake#makers#ft#d#dmd()
+    let l:args = ['-c', '-o-', s:UpdateDub()]
+    "Updating dub paths each make might be slow?
+    if exists("g:neomake_d_dmd_conf")
```
benekastah: Could this be renamed to `g:neomake_d_dmd_args_conf`? That would mesh better with the changes made in 002f8250a627ece42280afcc5789ecf46884aca5.

benekastah:
almost ready to merge! thanks for going through all this :)

rsw0x:
Sorry for the delay, had no free time. I think that should be everything.

benekastah:
Thanks!

PR #92, opened 2015-04-13
basic elixir checker

alxndr:
...using the system's `elixir` executable. This will catch more basic syntax errors.

### DIFF
--- a/autoload/neomake/makers/ft/elixir.vim
+++ b/autoload/neomake/makers/ft/elixir.vim
@@ -1,0 +1,13 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#elixir#EnabledMakers()
+    return ['elixir']
+endfunction
+
+function! neomake#makers#ft#elixir#elixir()
+    return {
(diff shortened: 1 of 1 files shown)


benekastah:
Thanks!

johnhamelink:
It would be good if the elixir checker could also load in the mix context - otherwise this PR isn't useful for anything other than scripts which aren't using mix. Or am I mistaken?

derekprior:
@johnhamelink here is what I'm using to try out compiling via mix instead:

```
let g:neomake_elixir_enabled_makers = ['mix']
let g:neomake_elixir_mix_maker = {
  \ 'args': ['compile'],
  \ 'errorformat':
  \ '** %s %f:%l: %m,' .
  \ '%f:%l: warning: %m'
  \ }
```

It's pretty slow compared to `elixir` but it seems to be working. If it works okay for me over this week, I'd like to PR it, but I think ideally there'd be a way to dynamically choose between `elixir` or `mix` depending on whether we can determin if you are in a `mix`-ified project or not.

PR #93, opened 2015-04-13
Processing lint results (placing signs) takes a long time on large files

### LINKED ISSUE #48: Processing lint results (placing signs) takes a long time on large files

benekastah:
Fixes #48

I'm going to go ahead and open this PR. Looks like statusline integration is broken, so I'll fix that first before I merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ things may break and change often until they do. That said, I'm using it daily
 doesn't work for you!
 
 The minimum Neovim version supported by Neomake is `NVIM 0.0.0-alpha+201503292107` (commit `960b9108c`).
+The minimum Vim version supported by Neomake is 7.4.503 (although if you don't
+use `g:neomake_logfile` older versions will probably work fine as well).
 
 ## How to use (basic)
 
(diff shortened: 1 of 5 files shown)


Neki:
To fix statusline integration, you need [this commit](https://github.com/Neki/neomake/commit/619f7e7f008b04660a1bfb124cbfeabc9456b62b) (disclaimer: not fully tested).

benekastah:
@Neki Yeah, looks like I just forgot to add the function calls.

PR #98, opened 2015-04-16
Use :language messages instead of :language

yyt16384:
#97 uses `:language` to change locale to `en_US.utf8`, which is not available on Windows. In this case we only need to set language for messages, so `:language messages` should work fine (and it works on Windows).


benekastah:
Thanks for the fix!

PR #102, opened 2015-04-16
set language to generic C

jspricke:
en_US.UTF-8 doesn't need to bee available, resulting in an error
message.


benekastah:
In ddb52872addd762daa2e8da72f964269a5343707 I made changes to this line as well. Please try that and see if it works for you. Does `C.UTF-8` set the language to english (which will be required to parse the signs text properly)?

jspricke:
Seems to work for me, Closing this.

Issue #94, opened 2015-04-14
The Rubocop maker does not emit error types recognized by neomake.


Neki:
Rubocop emits error types such as "C" which are not properly parsed (for instance in the statusline, which only recognizes "W" and "E"). Syntastic has special handling for rubocop, `errorformat` is not enough.

Neki:
Closed by #103.

PR #103, opened 2015-04-18
Add a way to alter quickfix or location list entries.

Neki:
This is a possible solution to #94: allow makers to change entries in the quickfix or location list before they are further processed by neomake (in the rubocop case, normalize the error type to 'W' or 'E').

`mapexpr` is not really adapted for this case, since to accurately find out what to change in rubocop output we would need to parse its output like `errorformat` does. I find it more robust to work on the entry objects.

The post processing function for rubocop comes from Syntastic.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -240,6 +240,11 @@ function! s:AddExprCallback(maker) abort
         let entry = list[s:neomake_list_nr]
         let s:neomake_list_nr += 1
 
+        if has_key(a:maker, 'postprocess')
+            let Func = a:maker.postprocess
+            call Func(entry)
+        end
+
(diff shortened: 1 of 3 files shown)


benekastah:
Can we rename this just to postprocess? I'd like to rename mapexpr to preprocess to make the distinction between these two properties clear (but not in this PR).

Neki:
Yes, this name was a tad lengthy :)

Do you want me to `rebase -i` the commits?

benekastah:
That would be nice in this case to make things like `git bisect` easier down the road.

Neki:
Done, thanks for your comments.

benekastah:
No problem. Thanks for doing this!

PR #106, opened 2015-04-19
Fix availability check

Tranquility:
The neomake#GetMaker interface changed (in f29e7e1de76ca7c9e3f3ae2e6a2e384c400a8436) and therefore only default makers
were returned. This broke the check for makers that changed the exe name
(ie. ruby#mri).


benekastah:
Thanks for catching this!

PR #107, opened 2015-04-20
Remove s:last_modified_signs.

Neki:
Does that variable have a purpose, or is it just a leftover?

This commit removes all occurrences of this variable, since it can only have the value `{}`, but maybe there is a detail I am not aware of.

Edit: same remark for s:modified_signs.


benekastah:
good catch!

PR #108, opened 2015-04-20
Keep state between two subsequent neomake runs.

### LINKED ISSUE #100: State is lost between two subsequent neomake runs
If you run neomake in a buffer in one window, and then again in another buffer in another window, the signs and statusline information in the first buffer get destroyed. Ideally those would stick around.

Neki:
Fixes #100. This is a work in progress, not completely tested yet. The first two commits here belong to the PR #107 but are necessary for this one (I will clean the git history once the status of #107 is decided).

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -199,13 +199,13 @@ function! neomake#Make(options) abort
     if !get(a:options, 'continuation')
         " Only do this if we have one or more enabled makers
         if file_mode
-            let bufnr = bufnr('%')
-            let winnr = winnr()
-            call neomake#signs#Reset(bufnr)
+            let buf = bufnr('%')
+            let win = winnr()
+            call neomake#signs#Reset(buf)
             let s:cleared_current_errors = get(s:, 'cleared_current_errors', {})
-            let s:cleared_current_errors[bufnr] = 0
+            let s:cleared_current_errors[buf] = 0
             let s:loclist_nr = get(s:, 'loclist_nr', {})
-            let s:loclist_nr[winnr] = 0
+            let s:loclist_nr[win] = 0
         else
             call neomake#signs#ResetAllBuffers()
             let s:cleared_current_errors = {}
@@ -461,8 +461,8 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
 endfunction
 
 function! s:CleanAllSignsAndErrors()
-    for bufnr in keys(s:cleared_current_errors)
-        call s:CleanSignsAndErrors(bufnr)
+    for buf in keys(s:cleared_current_errors)
+        call s:CleanSignsAndErrors(buf)
     endfor
 endfunction
 
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -202,13 +202,10 @@ function! neomake#Make(options) abort
             let buf = bufnr('%')
             let win = winnr()
             call neomake#signs#Reset(buf)
-            let s:cleared_current_errors = get(s:, 'cleared_current_errors', {})
-            let s:cleared_current_errors[buf] = 0
             let s:loclist_nr = get(s:, 'loclist_nr', {})
             let s:loclist_nr[win] = 0
         else
             call neomake#signs#ResetAllBuffers()
-            let s:cleared_current_errors = {}
             let s:qflist_nr = 0
         endif
     endif
@@ -461,19 +458,16 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
 endfunction
 
 function! s:CleanAllSignsAndErrors()
-    for buf in keys(s:cleared_current_errors)
+    for buf in keys(s:current_errors)
         call s:CleanSignsAndErrors(buf)
     endfor
 endfunction
 
 function! s:CleanSignsAndErrors(bufnr)
     call neomake#signs#CleanOldSigns(a:bufnr)
-    if has_key(s:cleared_current_errors, a:bufnr) && !s:cleared_current_errors[a:bufnr]
-        let errors = get(s:, 'current_errors', {})
-        if has_key(errors, a:bufnr)
-            unlet s:current_errors[a:bufnr]
-        endif
-        let s:cleared_current_errors[a:bufnr] = 1
+    let current_errors = get(s:, 'current_errors', {})
+    if has_key(current_errors, a:bufnr)
+        unlet s:current_errors[a:bufnr]
     endif
 endfunction
 
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -9,6 +9,10 @@ let s:current_errors = {
     \ 'project': {},
     \ 'file': {}
     \ }
+let s:need_errors_cleaning = {
+    \ 'project': 1,
+    \ 'file': {}
+    \ }
 
 function! neomake#ListJobs() abort
     call neomake#utils#DebugMessage('call neomake#ListJobs()')
@@ -206,10 +210,12 @@ function! neomake#Make(options) abort
             let buf = bufnr('%')
             let win = winnr()
             call neomake#signs#ResetFile(buf)
+            let s:need_errors_cleaning['file'][buf] = 1
             let s:loclist_nr = get(s:, 'loclist_nr', {})
             let s:loclist_nr[win] = 0
         else
             call neomake#signs#ResetProject()
+            let s:need_errors_cleaning['project'] = 1
             let s:qflist_nr = 0
         endif
     endif
@@ -287,9 +293,9 @@ function! s:AddExprCallback(maker) abort
         " On the first valid error identified by a maker,
         " clear the existing signs
         if file_mode
-            call neomake#CleanFileSignsAndErrors(entry.bufnr)
+            call neomake#CleanOldFileSignsAndErrors(entry.bufnr)
         else
-            call neomake#CleanProjectSignsAndErrors()
+            call neomake#CleanOldProjectSignsAndErrors()
         endif
 
         " Track all errors by buffer and line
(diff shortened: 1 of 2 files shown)


autoload/neomake.vim
```diff
@@ -236,9 +243,9 @@ function! s:AddExprCallback(maker) abort
     let place_signs = get(g:, 'neomake_place_signs', 1)
     let list = file_mode ? getloclist(a:maker.winnr) : getqflist()
 
-    while s:neomake_list_nr < len(list)
-        let entry = list[s:neomake_list_nr]
-        let s:neomake_list_nr += 1
+    while b:neomake_list_nr < len(list)
```
benekastah: In order for the `b:` prefix to work, we will need to be sure that this function can never be called when the user is not in the buffer that initiated `:Neomake`. I'm not sure if that's the case right now.

Neki: Good point, this probably applies to the other place where I used `b:`.
There is another reason this does not work with `Neomake!`: entries in the quickfix list are going to be processed several times.

autoload/neomake/signs.vim
```diff
@@ -46,25 +56,19 @@ function! neomake#signs#PlaceSign(entry) abort
     endif
 endfunction
 
-function! neomake#signs#CleanOldSigns() abort
-    if !empty(s:last_placed_signs)
-        call neomake#utils#DebugObject('Cleaning old signs:', s:last_placed_signs)
+function! neomake#signs#CleanOldSigns(bufnr) abort
+    if !has_key(s:last_placed_signs, a:bufnr)
```
benekastah: Could change to:

```
if !has_key(s:last_placed_signs, a:bufnr) || empty(s:last_placed_signs[a:bufnr])
    return
endif
```

Then you wouldn't have to wrap the DebugObject call in the if statement.

Neki:
Thanks for your feedback. So far it seems to work with file makers. Next step: make it work with project makers (eliminate the `:bufdo` placeholder + a couple of other things).

I do not have paid attention to memory usage ; I think there is a few places where `unlet` could be used.

Neki:
This is ready for review. I don't currently have a project on which I can test this in real world conditions (holidays...), but manual testing has not revealed any major issue.

autoload/neomake.vim
```diff
@@ -195,9 +195,19 @@ function! neomake#Make(options) abort
 
     if !get(a:options, 'continuation')
         " Only do this if we have one or more enabled makers
-        call neomake#signs#Reset()
-        let s:neomake_list_nr = 0
-        let s:cleared_current_errors = 0
+        if file_mode
+            let bufnr = bufnr('%')
+            let winnr = winnr()
```
benekastah: Can this be:

```
let buf = bufnr('%')
let win = winnr()
```

so we aren't overwriting those functions in this scope.

autoload/neomake.vim
```diff
@@ -260,9 +271,10 @@ function! s:AddExprCallback(maker) abort
 
         " On the first valid error identified by a maker,
         " clear the existing signs in the buffer
-        call s:CleanSignsAndErrors()
+        call s:CleanSignsAndErrors(entry.bufnr)
```
benekastah: It seems like this means that a buffer can't have signs/errors for a `:Neomake` run and a `:Neomake!` run concurrently. Is that right?

autoload/neomake.vim
```diff
@@ -421,11 +444,20 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
...
-        let s:cleared_current_errors = 1
+function! s:CleanAllSignsAndErrors()
+    for bufnr in keys(s:cleared_current_errors)
+        call s:CleanSignsAndErrors(bufnr)
+    endfor
+endfunction
+
+function! s:CleanSignsAndErrors(bufnr)
+    call neomake#signs#CleanOldSigns(a:bufnr)
+    if has_key(s:cleared_current_errors, a:bufnr) && !s:cleared_current_errors[a:bufnr]
```
benekastah: Not sure we need this `s:cleared_current_errors` variable. It seems like if `s:current_errors` doesn't have the `a:bufnr` key, then clearing the errors is essentially a no-op anyway.

autoload/neomake/signs.vim
```diff
@@ -44,18 +59,17 @@ function! neomake#signs#PlaceSign(entry) abort
     endif
 endfunction
 
-function! neomake#signs#CleanOldSigns() abort
-    if !empty(s:last_placed_signs)
-        call neomake#utils#DebugObject('Cleaning old signs:', s:last_placed_signs)
+function! neomake#signs#CleanOldSigns(bufnr) abort
+    if !has_key(s:last_placed_signs, a:bufnr) || empty(s:last_placed_signs[a:bufnr])
```
benekastah: I think I advised you to move the `empty` call here, but now I think we don't need it at all since we are `unlet`ing at the end of this function call. We'll free a bit more memory without it, too.

autoload/neomake.vim
Neki: :+1:

autoload/neomake.vim
Neki: :+1:

autoload/neomake/signs.vim
Neki: :+1:

autoload/neomake.vim
Neki: I am not sure yet about how `s:last_placed_signs` works. At first glance my answer is yes, it can't (I have to test it to be really sure), but that would really be a shame, since a useful use case is to run a linter with `:Neomake` on every write while a long build with `:Neomake!` is done in the background.

I do not see the difference between `:Neomake`  and `:Neomake!` when it comes to interfering though (what if I run concurrently `:Neomake ghcmod` and `:Neomake hlint`?)

Since there is a guarantee that there will be only one job by maker, we could track signs by maker instead, and remove only the signs corresponding to a given maker.

Neki:
Just to say that I have not abandoned this, but I do not have a lot of free time in this moment :)

Neki:
Sorry for the delay.

I tried to handle signs and errors on a per-maker basis, but the resulting code was way too unwieldy for the benefits. So now signs and errors are handled separately for 'file' and 'project' makers (no other distinction is made).

`:Neomake` and `:Neomake!` calls no longer interfere, so calling `:Neomake` will never clean signs set by `:Neomake!`.

Maybe we should now add commands such as `:NeomakeClearAll`, `:NeomakeClearProject`, :`NeomakeClearFile`, but this PR is big enough as it is and I think it is better to do that in another PR. We will also need to display in EchoCurrentMessage the maker who produced the error (so the user can know whether it comes from `Neomake!` or `Neomake`).

I also rebased this branch on the lastest master.

Neki:
Actually do not merge this just yet, it breaks EchoCurrentError.

benekastah:
Awesome stuff here. When that last issue is fixed I'm excited to merge. Thanks for doing this!

Neki:
The last commit fixes it. I haven't run into more issues, but I have not tested this during a real work session.

benekastah:
Feel free to merge this once you feel you have tested it properly. Tomorrow I can test this branch during my workday as well to help out with that.

benekastah:
Not 100% sure if it's this branch's fault or not, but I got this error:

```
Error detected while processing function neomake#MakeHandler..<SNR>91_RegisterJobOutput..neomake#ProcessCurrentBuffer..neomake#signs#PlaceVisibleSigns..neomake#signs#PlaceSign:
line   12:
E121: Undefined variable: type
Error detected while processing function neomake#MakeHandler..<SNR>91_RegisterJobOutput..neomake#ProcessCurrentBuffer..neomake#signs#PlaceVisibleSigns..neomake#signs#PlaceSign:
line   12:
E15: Invalid expression: type ==# 'neomake_err'
Error detected while processing function neomake#MakeHandler..<SNR>91_RegisterJobOutput..neomake#ProcessCurrentBuffer..neomake#signs#PlaceVisibleSigns..neomake#signs#PlaceSign:
line   12:
E121: Undefined variable: type
Error detected while processing function neomake#MakeHandler..<SNR>91_RegisterJobOutput..neomake#ProcessCurrentBuffer..neomake#signs#PlaceVisibleSigns..neomake#signs#PlaceSign:
line   12:
E15: Invalid expression: type ==# 'neomake_err'
```

Unfortunately I didn't record the logfile, so that's all I have.

benekastah:
Also, EchoCurrentError doesn't appear to be working for me.

Neki:
The first error should be solved now (I realized I did not test this with makers that produce errors instead of warnings). 

EchoCurrentError works for me using the default configuration for Haskell (both with `:Neomake` and `:Neomake! cabal`) ; could you give me more details?

I found another issue: sometimes (I don't know how to reproduce yet) signs are not placed, even if errors are correctly displayed by EchoCurrentError.

benekastah:
EchoCurrentError still not working most of the time for me. It does work occasionally, though.

Neki:
I got the problem with EchoCurrentError not working today (and I have the logs), I will look into this.

Neki:
The last two commits should fix the issues identified so far. I am going to test this tomorrow at work.

benekastah:
Cool, I'll test it as well

benekastah:
Could you merge master back into your branch? I'm getting errors that have been fixed on master (unrelated to your changes).

Neki:
It's done :)

I did not spot any issue today, looking good so far. That said, I have not tested `:Neomake!` enough to be fully confident yet.

Neki:
This behaved correctly for me in the past few days, you can merge this when you think you have tested it enough too.

benekastah:
cool, I'll make sure I test it out on Monday

hpurmann:
Great! This works really well. I appreciate your work on this plugin, @benekastah, @Neki, @natemara and all the others. It was the reason I finally switched to neovim.

Do you know of other plugins which are only possible/improved a lot by using new features like the job control API of neovim?

Neki:
Thanks!

To be fair, it would possible for Syntastic to achieve something similar to neomake (even in regular Vim). However,  it would require a lot of work given the Syntastic established code base (@benekastah opened a discussion on this subject at https://github.com/scrooloose/syntastic/issues/1253). Neomake has the advantage of being a fresh start.

Some of the plugins that use Neovim features are listed in the [Neovim wiki](https://github.com/neovim/neovim/wiki/Related-projects). I don't use any of them (apart from vim-plug, which also works in regular Vim if you have Python or Ruby) so I can't really comment here.

PR #110, opened 2015-04-22
Fix NeomakeSh maker generation

### LINKED ISSUE #109: NeomakeSh broken
`:NeomakeSh ls`
gives
```
Error detected while processing function <SNR>57_NeomakeSh..neomake#Make..neomake#GetMaker:
line   40:
E15: Invalid expression: 'neomake_'.a:name.'_'.key
```

I think I have a fix...

ViViDboarder:
There are 3 changes:

 * Change `GetMaker` parameter to `maker_id` so it's clear that it's not always a name
 * Set default `errorformat` to something that is usable
 * use `maker.name` for the config vars instead of `a:name` since it could be a dictionary

Fixes #109

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -84,29 +84,29 @@ function! neomake#MakeJob(maker) abort
     endif
 endfunction
 
-function! neomake#GetMaker(maker_id, ...) abort
+function! neomake#GetMaker(name_or_maker, ...) abort
     if a:0
         let ft = a:1
     else
         let ft = ''
     endif
-    if type(a:maker_id) == type({})
-        let maker = a:maker_id
+    if type(a:name_or_maker) == type({})
+        let maker = a:name_or_maker
     else
-        if a:maker_id ==# 'makeprg'
+        if a:name_or_maker ==# 'makeprg'
             let maker = neomake#utils#MakerFromCommand(&shell, &makeprg)
         elseif len(ft)
-            let maker = get(g:, 'neomake_'.ft.'_'.a:maker_id.'_maker')
+            let maker = get(g:, 'neomake_'.ft.'_'.a:name_or_maker.'_maker')
         else
-            let maker = get(g:, 'neomake_'.a:maker_id.'_maker')
+            let maker = get(g:, 'neomake_'.a:name_or_maker.'_maker')
         endif
         if type(maker) == type(0)
             unlet maker
             try
                 if len(ft)
-                    let maker = eval('neomake#makers#ft#'.ft.'#'.a:maker_id.'()')
+                    let maker = eval('neomake#makers#ft#'.ft.'#'.a:name_or_maker.'()')
                 else
-                    let maker = eval('neomake#makers#'.a:maker_id.'#'.a:maker_id.'()')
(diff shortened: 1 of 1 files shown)


autoload/neomake.vim
```diff
@@ -132,7 +132,9 @@ function! neomake#GetMaker(name, ...) abort
             let maker[key] = defaults[key]
         endif
     endfor
-    let maker.name = a:name
+    if !has_key(maker, 'name')
```
benekastah: This bit should go right after line 116, so that `maker.name` definitely exists before we try to use it.

autoload/neomake.vim
```diff
@@ -115,16 +115,16 @@ function! neomake#GetMaker(name, ...) abort
     endif
     let maker = copy(maker)
     let defaults = {
-        \ 'exe': a:name,
+        \ 'exe': a:maker_id,
         \ 'args': [],
-        \ 'errorformat': '',
```
benekastah: I don't think this change is right. Instead, in `s:ProcessJobOutput` there is a line `if has_key(a:maker, 'errorformat')` and then it sets up the errorformat in the `if` body. I think we should just change that to `if len(a:maker.errorformat)` and the behavior should then be the same as before my change.

autoload/neomake.vim
```diff
@@ -84,29 +84,29 @@ function! neomake#MakeJob(maker) abort
     endif
 endfunction
 
-function! neomake#GetMaker(name, ...) abort
+function! neomake#GetMaker(maker_id, ...) abort
```
benekastah: I don't think this argument's name needed to change, since as far as I can tell it is the same thing as it was before.

benekastah:
Thanks for fixing a couple bugs I introduced! Let me know if any of my feedback doesn't make sense.

autoload/neomake.vim
ViViDboarder: Good point. I'll move that.

autoload/neomake.vim
ViViDboarder: Yea, that should work.

autoload/neomake.vim
ViViDboarder: Yea, it doesn't need to change. The reason I changed it is because the `a:name` is not really a name in some cases. It's easy for someone new to the project (or even heavily involved) to forget that `a:name` isn't just a string and then end up with errors like the one I'm attempting to fix.

Maybe another name besides `a:maker_id` or `name_or_maker`?

benekastah: I think `name_or_maker` is a good choice.

autoload/neomake.vim
```diff
@@ -114,17 +114,20 @@ function! neomake#GetMaker(name, ...) abort
         endif
     endif
     let maker = copy(maker)
+    if !has_key(maker, 'name')
+        let maker.name = a:maker_id
+    endif
     let defaults = {
-        \ 'exe': a:name,
+        \ 'exe': a:maker_id,
```
benekastah: This could be `'exe': maker.name,` as well

ViViDboarder:
Got caught up. I saw there was no other changes to this file, so no rebase should be needed.

benekastah:
Thanks!

PR #113, opened 2015-04-23
Added maker for php and phpcs

tthbellcom:
This adds maker recipes for PHP lint (php -l) and PHP_Codesniffer.

phpcs is configurable with a standard, which can change depending on each projects conventions. I have added the posibility to set `g:neomake_php_phpcs_args_standard` depending on the needs. This seems to be the standards used in the D recipe.


benekastah:
Thanks for the contribution!

tthbellcom:
Thank you for the module, in the first place.

Issue #112, opened 2015-04-23
Display the maker progress (output) to the user.


Neki:
I was thinking about completely removing the dispatch plugin to keep only neomake. There is still one feature neomake lacks to do this: being able to see the maker output (seeing it in the logs is a poor replacement) in real time. This is important for long build cycles with `:Neomake!`.

benekastah:
Wouldn't `:Neomake | copen` work?

Neki:
It does as long as #27 is not implemented and the location/quickfix list contains all the maker output.

However, if one wants to use `:Neomake!` to compile a reasonably large project, the build output will contain a lot of lines that are just progress information, and ultimately these lines should not appear in the list (for instance, building neovim with `:Neomake! make`).

badeip:
I am also very interested in real time monitoring of `:Neomake!` progress.
Doing as @benekastah suggest works fairly well, but I'd like the quickfix window to always display the latest output (as in `tail -f`)

aliev:
+1 for this feature. it will be very useful

sooth-sayer:
👍

Determinant:
+1 for this feature, which is quite common in modern IDEs.

mqudsi:
For me, I just needed to know when the build was done so I created this script that adds a spinning progress indicator to my status line that starts and stops with the neomake build:

```vim
let s:spinner_index = 0
let s:active_spinners = 0
let s:spinner_states = ['|', '/', '--', '\', '|', '/', '--', '\']
let s:spinner_states = ['┤', '┘', '┴', '└', '├', '┌', '┬', '┐']
let s:spinner_states = ['←', '↖', '↑', '↗', '→', '↘', '↓', '↙']
let s:spinner_states = ['←', '↑', '→', '↓']
let s:spinner_states = ['d', 'q', 'p', 'b']
let s:spinner_states = ['.', 'o', 'O', '°', 'O', 'o', '.']
let s:spinner_states = ['■', '□', '▪', '▫', '▪', '□', '■']

function! StartSpinner()
    let b:show_spinner = 1
    let s:active_spinners += 1
    if s:active_spinners == 1
        let s:spinner_timer = timer_start(1000 / len(s:spinner_states), 'SpinSpinner', {'repeat': -1})
    endif
endfunction

function! StopSpinner()
    let b:show_spinner = 0
    let s:active_spinners -= 1
    if s:active_spinners == 0
        :call timer_stop(s:spinner_timer)
    endif
endfunction

[...]

PR #114, opened 2015-04-25
Add an option to remove invalid entries from the location/quickfix list.

### LINKED ISSUE #27: Add option to get rid of junk lines in maker output
Quoting @Tranquility in #25:

> Currently it displays the header line of the ouput as first item in the loclist(|| path,lineNumber,lineNumberEnd,level,message). Syntastic probably does something clever to get rid of this line? Otherwise it works as expected.

Neki:
Fix #27.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -122,6 +122,7 @@ function! neomake#GetMaker(name_or_maker, ...) abort
         \ 'args': [],
         \ 'errorformat': &errorformat,
         \ 'buffer_output': 0,
+        \ 'remove_invalid_entries': 1
         \ }
     for key in keys(defaults)
         if len(ft)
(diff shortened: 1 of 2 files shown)


benekastah:
Perhaps there could also be a way to override this per maker for the usecase you described in #112?

Neki:
This is not the perfect solution to #112, but it is a good idea.
In order to keep things simple, I removed the global option and made it a per-maker option. I think the interaction between the global option and the per-maker option could be confusing.

autoload/neomake.vim
```diff
@@ -247,11 +249,20 @@ function! s:AddExprCallback(maker) abort
             call Func(entry)
         end
 
-        if entry.valid && !file_mode
+        if !entry.valid
+            if get(a:maker, "remove_invalid_entries", 0)
```
benekastah: I think the default should be 1, since the most common use case is syntax checking.

benekastah: Also, you shouldn't need to use `get` here. Since you added that property to the list of defaults, that property will always be available on the maker.

benekastah:
Apart from my comments, looks good to me. Thanks for doing this! This ended up being simpler than I thought it would be.

Neki:
I have implemented your comments and squashed/rebased the commits.

benekastah:
Looks good! Thanks, excited to try this out.

PR #119, opened 2015-05-05
remember makeprg errorformat

edma2:
My `errorformat` isn't being respected by Neomake when used with `makeprg`.

Here's what I think is happening: I set a new `errorformat`, call `Neomake!`,  then restore the old `errorformat`. The job finishes and Neomake parses the output with the *current* `errorformat` (which is the old `errorformat`) rather than the one I set before calling Neomake! (the new `errorformat`).

Solution is to add an extra `errorformat` argument to MakerFromCommand so that Neomake "remembers" the initial `errorformat`.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -94,7 +94,7 @@ function! neomake#GetMaker(name_or_maker, ...) abort
         let maker = a:name_or_maker
     else
         if a:name_or_maker ==# 'makeprg'
-            let maker = neomake#utils#MakerFromCommand(&shell, &makeprg)
+            let maker = neomake#utils#MakerFromCommand(&shell, &makeprg, &errorformat)
         elseif len(ft)
             let maker = get(g:, 'neomake_'.ft.'_'.a:name_or_maker.'_maker')
         else
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -121,7 +121,7 @@ function! neomake#utils#Random() abort
     return answer
 endfunction
 
-function! neomake#utils#MakerFromCommand(shell, command) abort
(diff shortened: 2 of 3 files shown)


benekastah:
If there is no errorformat on the maker, we don't try to set it. See https://github.com/benekastah/neomake/blob/master/autoload/neomake.vim#L292. However, this was broken a bit ago and fixed in #110. Have you verified that the latest neomake without this commit is broken? If so, the problem should be fixed with the logic that sets errorformat instead of this change.

edma2:
If I'm reading the code correctly, the makeprg maker has no errorformat. See
https://github.com/edma2/neomake/blob/master/autoload/neomake.vim#L97 and
https://github.com/edma2/neomake/blob/master/autoload/neomake/utils.vim#L135

This means Neomake will use the errorformat at the time the job finishes (because we don't touch errorformat if it isn't set in the maker, as you pointed out), rather than the erroformat at the time you invoked Neomake. Is this by design?

My use case is that I have a script that sets a custom errorformat, calls make, then restores the old errorformat. With synchronous make (e.g. :make), this problem doesn't exist because the errorformat is the same before and after invocation. Asynchronous execution complicates things because the errorformat might change while make is running.

The question is, should Neomake use the errorformat at dispatch time, or when the job finishes. IMO the former case is least surprising.

benekastah:
Ah, I see. Yes, you have a good point. I went ahead and fixed it in a more general way here: 2c4701aa9eb320235e0c5513060196327ab3a074. Thanks for raising this issue and pull request!

benekastah:
And of course, let me know if my fix is not sufficient for you and we can revisit this PR if needed.

edma2:
:+1:  thanks for the fix!

PR #123, opened 2015-05-06
Add maker for Haskell's hdevtools.

nathyong:
This is almost a copy of ghc-mod's maker, but without the null byte check. ghc-mod doesn't quite build with the new version of GHC (7.10.1), at least on my machine, but hdevtools does almost the same thing in a daemonised form and builds nicely.


benekastah:
Great, thanks!

PR #125, opened 2015-05-06
Deepcopy returned maker in GetMaker.

Sean1708:
The issue is explained in the commit message but the long and short of it is that when `copy()` was used instead of `deepcopy()` the filename would be appended to the args list again each time `MakeJob()` was called.

This becomes an issue if the executable takes a set number of arguments or if passing the filename more than once has undesirable consequences.


benekastah:
Good catch, thanks!

Sean1708:
No problem, thanks for the amazing plugin :)

PR #132, opened 2015-05-18
Makers: support for phpmd

winged:
This adds (very rudimentary) support for the [PHPMD](http://phpmd.org/) linter.


benekastah:
Thanks!

PR #136, opened 2015-05-24
Fix enabled makers setting

### LINKED ISSUE #135: g:neomake_<ft>_enabled_makers broken
see comment by @dnhgff on https://github.com/benekastah/neomake/pull/133

thomasboyt:
I'm a vimscript novice, so apologies in advance if this isn't the ideal solution, but this seems to fix #135. Looks like it was checking the existence of `neomake_foo_enabled_makers` instead of `g:neomake_foo_enabled_makers`.


benekastah:
This looks right. Thanks!!

PR #137, opened 2015-05-27
add column support for D.

rsw0x:
Just a small change to output(and parse) the column number.

### DIFF
--- a/autoload/neomake/makers/ft/d.vim
+++ b/autoload/neomake/makers/ft/d.vim
@@ -38,25 +38,45 @@ function! s:UpdateDub()
     return l:dub_incs
 endfunction
 
-function! neomake#makers#ft#d#dmd()
-    let l:args = ['-c', '-o-', '-vcolumns', s:UpdateDub()]
+"GDC does not adhere to dmd's flags or output, but to GCC's.
+"This is for LDC and dmd only.
+function! s:DmdStyleMaker(...)
     "Updating dub paths each make might be slow?
-    if exists("g:neomake_d_dmd_args_conf")
-        call add(l:args, '-conf=' . expand(g:neomake_d_dmd_args_conf))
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/d.vim
```diff
@@ -49,8 +49,7 @@ function! neomake#makers#ft#d#dmd()
     return {
         \ 'args': l:args,
         \ 'errorformat':
-        \     '%-G%f:%s:,%f(%l): %m,' .
-        \     '%f:%l: %m',
+        \     '%f(%l\,%c): %m,'
```
benekastah: You've removed 3 errorformat entries and only added one. Are you sure you didn't mean to just replace the last one? Also, if the program may not always give a column number (even with the `-vcolums` argument - some errors may not be column specific), you should just add your errorformat entry instead of repalacing them all.

rsw0x: The error format that was there was shamelessly copied from syntastic, which appears to be taken from a very old forum post dating ~2004 after a bit of googling.

The error format I replaced it with works just fine(actually, a bit better - the original had an issue parsing template errors.)

But you are right on the vcolumns point, I was under the impression that the other compilers adhered strictly to the reference compiler's flags, but was wrong. It appears that the GCC-based compiler doesn't even output the same format(I didn't have it installed when I first wrote it,) but in GCC format.

Going to push a few changes to fix all of this.

benekastah: Changes look good to me. If not 100% sure you're done, but if you're comfortable with it I'll merge.

rsw0x: Hi, sorry for the late reply. I'm done with it, it seems to be working just fine now.

benekastah:
Thanks!

PR #139, opened 2015-06-05
Rust: faster checks by skipping codegen

Grimy:
rustc has a '-Z no-trans' flag that, when on, skips the costly LLVM code
generation phase, while still reporting all errors and warnings.

Enabling this flag makes linting much faster. It also renders "-o /dev/null"
unnecessary.


benekastah:
Thanks!

PR #141, opened 2015-06-08
Escape special characters in filetypes

### LINKED ISSUE #140: Neomake chokes on filetypes containing special characters
Running `setf post-receive | Neomake` yields the following errors:

```
Error detected while processing function <SNR>13_NeomakeCommand..neomake#GetEnabledMakers:
line 17:
E121: Undefined variable: neomake#makers#ft#post
E15: Invalid expression: neomake#makers#ft#post-receive#EnabledMakers()
Error detected while processing function <SNR>13_NeomakeCommand:
line 2:
E171: Missing :endif
```

I’m working on a patch for this issue.

Grimy:
Filetypes are treated as parts of variable names (using eval()). However, some
characters are legal in a filetype but not in an identifier. This caused the
Neomake command to completely fail when the current filename contained such
characters.

This commit fixes that by replacing all such characters with underscores ('_').

fixes #140

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -170,6 +170,7 @@ function! neomake#GetEnabledMakers(...) abort
     let union = {}
     let fts = neomake#utils#GetSortedFiletypes(a:1)
     for ft in fts
+        let ft = substitute(ft, '\W', '_', 'g')
         let varname = 'g:neomake_'.ft.'_enabled_makers'
         let fnname = 'neomake#makers#ft#'.ft.'#EnabledMakers'
         if exists(varname)


Grimy:
This change is crude, undocumented and poses some usability problems (it’s impossible to distinguish filetypes `c--` and `c++`, since they’re both converted to `c__`). However, I think it’s still way better than the current behavior (aborting with errors).

Note that the impact is minimal: while non-word characters are technically possible in filetypes, all builtin filetypes are legal identifiers (matching `/[a-z][a-z0-9]*/`). Thus, this change only affects users with a custom filetype scheme.

Neki:
I guess this one of the reasons why Syntastic uses an explicit "register checker" function.

Your change looks good to me.

benekastah:
Looks like we should plan a better way for the future. This is fine for now, though.

PR #142, opened 2015-06-11
Add support for pylama

gmacon:
This adds a maker definition for pylama, which is a meta-checker that covers pep8, pep257, pyflakes, mccabe, pylint, and gjslint.


benekastah:
Looks good, thanks!

PR #144, opened 2015-06-12
Show command output in quickfix list with :NeomakeSh.

Neki:
Without this change, :NeomakeSh does not give any visible output unless the user manually set its errorformat before running the command.


Neki:
Merging this since it is a really small change.

PR #148, opened 2015-06-17
[WIP] Add java/javac maker

brcolow:
Stolen from syntastic meta-stolen from @tarruda

=)

Just exercising my (super new) git skills (hopefully this is an okay contribution).

Todo:
- [ ] Figure out why my system is producing bogus results.
- [ ] Flesh out the javac maker with all args needed (-cp ?)
- [ ] Implement mvn maker (mvn compile ?)
- [ ] Gradle?

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -280,6 +280,7 @@ function! s:AddExprCallback(maker) abort
 
     while index < len(list)
         let entry = list[index]
+        let entry.maker_name = has_key(a:maker, 'name') ? a:maker.name : 'makeprg'
         let index += 1
 
         if has_key(a:maker, 'postprocess')
@@ -545,7 +546,8 @@ function! neomake#EchoCurrentError() abort
             break
         endif
     endfor
-    call neomake#utils#WideMessage(s:neomake_last_echoed_error.text)
+    let message = s:neomake_last_echoed_error.maker_name.': '.s:neomake_last_echoed_error.text
+    call neomake#utils#WideMessage(message)
 endfunction
 
 function! neomake#CursorMoved() abort
--- a/autoload/neomake/makers/ft/java.vim
+++ b/autoload/neomake/makers/ft/java.vim
@@ -1,0 +1,33 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#java#EnabledMakers()
+    return ['javac']
+endfunction
+
+function! neomake#makers#ft#java#javac()
+    let args = ['-Xlint']
+    let arg_d = []
+    if exists("g:neomake_java_javac_class_output_folder")
+        let arg_d = ['-d', fnameescape(expand(g:neomake_java_javac_class_output_folder))]
+    else
+        if neomake#utils#IsRunningWindows()
+            let arg_d = ['-d', 'C:/tmp/' . neomake#utils#Random()]
+        else
+            let arg_d = ['-d', '/tmp/' . neomake#utils#Random()]
+            echomsg("this branch executed")
+        endif
+    endif
+    let args += arg_d
+    echo args
+    return {
+        \ 'exe' : 'javac',
+        \ 'args': args,
+        \ 'errorformat':
+            \ '%E%f:%l: error: %m,' .
+            \ '%W%f:%l: warning: %m,' .
+            \ '%A%f:%l: %m,' .
+            \ '%+Z%p^,' .
+            \ '%+C%.%#,' .
+            \ '%-G%.%#',
+         \ }
+endfunction


Neki:
Thanks for the contribution!

Won't this checker clutter the working directory with `.class` files?

Skimming trough `man javac`, I don't see an option to prevent `javac` from outputing class files, but the `-d` option could be used to write class files to a temporary directory. Unfortunately, Neomake does not support yet 'post maker hooks' where deleting the temporary directory could be done.

Until this feature is implemented, I think a good workaround would be having a configuration variable  `g:neomake_java_javac_class_output_folder`. When set, it would be passed to `javac -d` so the user can isolate the class files in a 'classes' folder or in '/tmp' (and maybe '/dev/null' if it is supported). If not set, the `-d` option would not be used.

brcolow:
I gave it a go =)...I am super, super new to vimscript (this is the first code I've ever wrote in it) so there may be some egregious stuff going on.

brcolow:
Thanks for the responses, finally was able to run the script I wrote to see the syntax errors. I will take everything you said into account and revise my work. Thanks for all the help. I'll have a much better PR within 24 hours =)

benekastah:
Cool, I'm excited to try this next time I noodle around with java!

tarruda:
@brcolow could your issues be related to [this](https://github.com/neovim/neovim/issues/2143)?

brcolow:
Thanks very much for that link - will help me to expand this and get it working..this is still a heavy WIP. I need to figure out why, given the current code, the -d argument seems to not be present in the final call to javac at all, and then figure out a good way to support -cp.

brcolow:
I am having an issue...

When I print args (just before it is returned) it says:

['-Xlint', '-d', '/tmp/21477']

But then when neomake is starting, I get:

Neomake: Starting: javac -Xlint -d /Users/brcolow/code/MyFile.java

In other words, the -d argument is gone...it should say:

Neomake: Starting: javac -Xlint -d /tmp/21477 /Users/brcolow/code/MyFile.java

Is this a problem with what is printed...or is it an actual problem (e.g. why is the -d argument disspearing?)

Neki:
I can not reproduce what you describe. What I did (Ubuntu 14.04, neovim 0.0.0-alpha+201504230827 commit 04e098fc3c4):

* checkout the `master` branch from your fork
* `cd /tmp/`
* `nvim Test.java`
* add some Java code, save the file:
```java
public class Test {
}
```
* `:Neomake javac`

When Neomake runs, I get the messages:
```
this branch executed
['-Xlint', '-d', '/tmp/28702']
```

and in the logs I have:
```
2015-06-20T09:01:20+0200 Log level 3: Neomake: Calling neomake#utils#Random
2015-06-20T09:01:20+0200 Log level 2: Neomake: Starting: javac -Xlint -d /tmp/28702 /tmp/Test.java
2015-06-20T09:02:08+0200 Log level 3: Neomake: javac stderr: javac: directory not found: /tmp/28702
2015-06-20T09:02:08+0200 Log level 3: Neomake: javac stderr: 
2015-06-20T09:02:08+0200 Log level 3: Neomake: javac stderr done.
```

Could you post the exact steps to reproduce your issue?

A few notes on your code:
1. according to [Wikipedia](https://en.wikipedia.org/wiki/Temporary_folder#Traditional_locations), getting a temporary folder on Windows in going to be more convoluted than that. That's why I suggested you use `_neomake_classes` as the default value for the folder used to create the classes, it's simpler and platform-independent.
2. Creating a random folder for *each* `:Neomake` run is a bit too much ; if you want to sick with the temp folder approach (which I would discourage), I think it would be better to generate one random name and stick to it for the whole session.

brcolow:
When I do the *exact* same steps as you list, I get a different result.
````
  1 2015-06-20T13:09:07-0700 Log level 3: Neomake: Calling neomake#utils#Random
  2 2015-06-20T13:09:07-0700 Log level 2: Neomake: Starting: javac    /private/tmp/Test.java
2015-06-20T13:09:16-0700 Log level 3: Neomake: javac stderr: javac: invalid flag:
  4 2015-06-20T13:09:16-0700 Log level 3: Neomake: javac stderr:
  5 2015-06-20T13:09:16-0700 Log level 3: Neomake: javac stderr done.
...etc
````

**Comment by Neki** [comment]
What Neovim version (`:version`) and OS are you using?

I corrupted my Ubuntu today (PC issues) so I will not be able to help soon by trying to reproduce, sorry. I have no idea where this could come from, the code that creates the job does not do any magic with the arguments.

Maybe you could check if existing makers using several arguments (such as Ruby or Rubocop) work on your machine. You don't need to install the maker on your system, just putting an executable shell script on your $PATH with the correct name (so that Neomake detects it) should do the trick. For instance, if you don't have Ruby (disclaimer: I have not tested this):

```bash
$ cat ruby
#!/bin/bash
echo "$@"
$ chmod a+x ruby
$ mv ruby /usr/bin/
```

Looking forward to having `javac` integrated in Neomake :)

brcolow:
That was a good suggestion. I tried the ruby maker, specifically mri (which I have on my PATH (with Ruby)). And in fact, I get the same problem!

````
2015-06-20T14:19:11-0700 Log level 2: Neomake: Starting: ruby    /private/tmp/Test.rb
````

This is quite strange indeed. I am on Mac OS X, iTerm2, zsh shell.

Neovim version:

````
NVIM 0.0.0-alpha+201506170758 (compiled Jun 18 2015 16:55:46)
Commit: 7b87dc9299993003ea9cd2e0f0f38b881010501d
````

Now I am very confused.

Neki:
I opened #150 to track this specific issue. I don't have access to a Mac, so I won't be able to reprodue the exact same conditions (I do use zsh though, so it is not the culprit).

PR #149, opened 2015-06-19
Add JavaScript maker jscs

gmacon:
This is a style checker for JavaScript.

I'm wondering if there are rules for ordering multiple checkers for the same type.  In this PR, I went with "reverse alphabetical order" since it matched what was already here, but it would be good to know if there was something else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ Haskell:
 
 Javascript:
 - eslint
+- jscs
 - jshint
 - jsxhint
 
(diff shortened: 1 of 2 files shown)


benekastah:
The order is generally fastest to slowest. If all the makers are relatively quick, the order is arbitrary. I'm assuming this one is fairly quick since I haven't heard of a slow maker for javascript. This looks fine to me :)

Thanks for the contribution!

Neki:
@benekastah What about also ordering them by severity? Given that checkers that appear first in the list have their messages displayed in priority, I'd rather see a warning from clang instead of a warning from a style checker if both apply a warning to the same line.

benekastah:
Yeah, that seems pretty reasonable. So then the order could be severity first and speed second.

PR #152, opened 2015-06-22
Add vimscript maker.

beardedprojamz:
This adds a Vimscript maker file using Python-based linter, [vint](https://github.com/Kuniwak/vint).


benekastah:
Nice, excited to try this!

PR #157, opened 2015-07-16
fix long messages

nerdrew:
looks like this code was copied from syntastic and missed the `s:_width` definition: https://github.com/scrooloose/syntastic/blob/31cba018b381feb99c75a67abde7acc69daf2c50/autoload/syntastic/util.vim#L173


Neki:
LGTM, thanks for the contribution!

PR #159, opened 2015-07-18
Add completion to Neomake and NeomakeSh.

lucc:
I had a look around the code to find a way to retrieve all possible makers but am not sure if I found the bast way.  Maybe you can tell me if there is a better way.


Neki:
`neomake#GetEnabledMakers` returns the makers that are run when the user does not explicitly give a maker name. In theory there can be more available makers (that are not to be run when the user calls `:Neomake`, but may be called with `:Neomake mymaker`). There is currently no way to get a list of "available makers".

That said, I think your contribution makes perfect sense, since in practice I think "enabled" and "available" are more or less the same. Thanks!

Issue #161, opened 2015-07-25
Callback on maker exit


critiqjo:
I'm adding a feature to the plugin I'm working on [lldb.nvim](https://github.com/critiqjo/lldb.nvim) which will allow users to compile their code before launching the debug session. And I want to delegate command handling to Neomake, which will take care of displaying the output asynchronously (in quickfix), displaying error/warning signs etc.

The problem is lldb.nvim can't know when Neomake has completed a command it issued, and its exit status. It needs a callback on exit. So I am working on implementing basic callbacks for Neomake. I will submit a PR when ready.

benekastah:
This sounds like a good idea. I can think of two different callbacks that could be nice to have:

- A callback when an individual maker completes
- A callback when all the makers spawned by the :Neomake command complete.

petobens:
Hi @critiqjo  and @benekastah. Dispatch fires a `QuickFixCmdPost cgetfile` event each time a build finishes. Does neomake do the same? Thanks in advance.

critiqjo:
No, I guess it doesn't. Would that be useful to a user? The only use I can think of is for opening the quickfix window right after it completes... @petobens Do you use that event for anything?

petobens:
Hi @critiqjo. I find it useful because by checking that event I know when Dispatch finishes a build and I can therefore do:

```vim
au QuickFixCmdPost cgetfile call s:SomeRandomFunction()
```
I believe your implementation of exit callbacks generalizes this simple approach?

critiqjo:
Yea, I got that... I was asking if you could provide an example use-case...

Anyway, this is not a generalization of it. Because users won't be able to catch *all* exits.
Only the caller can request a callback. And the caller could be another plugin -- and the user will not be notified by any means. Notifying the user of all events would be another independent feature.

petobens:
I''ll try to provide a concrete a example. Thanks for your explanation.

gibiansky:
Isn't "The only use I can think of is for opening the quickfix window right after it completes..." an example usecase? (I came to this issue trying to do exactly that...)

petobens:
@gibiansky  sorry for the late reply. I also think its valid usecase (at least for me).

evindor:
Ran into that while trying to add neovim support via neomake to https://github.com/wincent/ferret

```
        autocmd QuickfixCmdPost cgetfile call ferret#private#post('qf')
```

jamischarles:
[...]

PR #162, opened 2015-07-25
[RFC] Implement exit callbacks

critiqjo:
Motivation: #161
Implemented callback for every individual maker exit event.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -271,6 +271,9 @@ function! neomake#Make(options) abort
             let next_opts.continuation = 1
             let maker.next = next_opts
         endif
+        if has_key(a:options, 'exit_callback')
+            let maker.exit_callback = a:options.exit_callback
+        endif
         call neomake#MakeJob(maker)
         " If we are serializing makers, stop after the first one. The
         " remaining makers will be processed in turn when this one is done.
@@ -476,6 +479,19 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
             endif
         endif
         let status = a:data
+        if has_key(maker, 'exit_callback')
+            try
+                let callback_dict = { 'status': status,
+                                    \ 'name': maker.name,
+                                    \ 'has_next': has_key(maker, 'next') }
+                if type(maker.exit_callback) == type('')
+                    let ExitCallback = function(maker.exit_callback)
+                else
+                    let ExitCallback = maker.exit_callback
+                endif
+                call ExitCallback(callback_dict)
+            endtry
+        endif
         call s:CleanJobinfo(jobinfo)
         if has('nvim')
             " Only report completion for neovim, since it is asynchronous
@@ -577,4 +593,29 @@ function! neomake#CompleteMakers(ArgLead, CmdLine, CursorPos)
         return filter(neomake#GetEnabledMakers(&ft),
                     \ "v:val =~? '^".a:ArgLead."'")
     endif
+endfunction
+
+function! neomake#WithCallback(file_mode, enabled_makers, ...)
+    let options = a:0 ? { 'exit_callback': a:1 } : {}
+    if a:file_mode
+        let options.enabled_makers = len(a:enabled_makers) ?
+                    \ a:enabled_makers :
+                    \ neomake#GetEnabledMakers(&ft)
+        let options.ft = &ft
+        let options.file_mode = 1
+    else
+        let options.enabled_makers = len(a:enabled_makers) ?
+                    \ a:enabled_makers :
+                    \ neomake#GetEnabledMakers()
+    endif
+    call neomake#Make(options)
+endfunction
+
+function! neomake#ShWithCallback(sh_command, ...)
+    let options = a:0 ? { 'exit_callback': a:1 } : {}
+    let custom_maker = neomake#utils#MakerFromCommand(&shell, a:sh_command)
+    let custom_maker.name = 'sh: '.a:sh_command
+    let custom_maker.remove_invalid_entries = 0
+    let options.enabled_makers =  [custom_maker]
+    call neomake#Make(options)
 endfunction
--- a/plugin/neomake.vim
+++ b/plugin/neomake.vim
@@ -1,37 +1,11 @@
 " vim: ts=4 sw=4 et
 
-function! s:NeomakeCommand(file_mode, enabled_makers)
-    if a:file_mode
-        call neomake#Make({
-            \ 'enabled_makers': len(a:enabled_makers) ?
-                \ a:enabled_makers :
-                \ neomake#GetEnabledMakers(&ft),
-            \ 'ft': &ft,
-            \ 'file_mode': 1,
-            \ })
-    else
-        call neomake#Make({
-            \ 'enabled_makers': len(a:enabled_makers) ?
-                \ a:enabled_makers :
-                \ neomake#GetEnabledMakers()
-            \ })
-    endif
-endfunction
-
-function! s:NeomakeSh(sh_command)
-    let custom_maker = neomake#utils#MakerFromCommand(&shell, a:sh_command)
-    let custom_maker.name = 'sh: '.a:sh_command
-    let custom_maker.remove_invalid_entries = 0
-    let enabled_makers =  [custom_maker]
-    call neomake#Make({'enabled_makers': enabled_makers})
-endfunction
-
-command! -nargs=* -bang -bar -complete=customlist,neomake#CompleteMakers Neomake call s:NeomakeCommand(<bang>1, [<f-args>])
+command! -nargs=* -bang -bar -complete=customlist,neomake#CompleteMakers Neomake call neomake#WithCallback(<bang>1, [<f-args>])
 " These commands are available for clarity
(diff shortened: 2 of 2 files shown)


autoload/neomake.vim
```diff
@@ -578,3 +591,28 @@ function! neomake#CompleteMakers(ArgLead, CmdLine, CursorPos)
                     \ "v:val =~? '^".a:ArgLead."'")
     endif
 endfunction
+
+function! neomake#WithCallback(file_mode, enabled_makers, callback)
```
critiqjo: Shall I make this callback argument optional?

critiqjo:
@benekastah how does it look?

critiqjo:
I've made it cleaner now... On each maker exit, Neomake will callback with a dict having keys `status`, `name` and `has_next`. Please check and comment.

autoload/neomake.vim
```diff
@@ -476,6 +485,14 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
             endif
         endif
         let status = a:data
+        if has_key(jobinfo, 'exit_callback')
+            try
+                let callback_dict = { 'status': status,
+                                    \ 'name': maker.name,
+                                    \ 'has_next': has_key(maker, 'next') }
```
benekastah: `has_next` is an internal property. I don't think it should be exposed. It's only used in certain cases (mainly when serializing multiple jobs, which isn't done by default).

autoload/neomake.vim
```diff
@@ -63,6 +63,9 @@ function! neomake#MakeJob(maker) abort
         let jobinfo.id = make_id
         " Assign this before neomake#MakeHandler gets run synchronously
         let s:jobs[make_id] = jobinfo
+        if has_key(a:maker, 'exit_callback')
```
benekastah: Since the `maker` is attached to `jobinfo`, there is no need to add `exit_callback` to `jobinfo`. Just get `exit_callback` from `jobinfo.maker` when you're ready to call it.

autoload/neomake.vim
```diff
@@ -476,6 +485,14 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
             endif
         endif
         let status = a:data
+        if has_key(jobinfo, 'exit_callback')
+            try
+                let callback_dict = { 'status': status,
+                                    \ 'name': maker.name,
+                                    \ 'has_next': has_key(maker, 'next') }
+                exe 'call '.(jobinfo.exit_callback).'(callback_dict)'
```
benekastah: It would be nice to support functions too.

```
if type(jobinfo.exit_callback) == type('')
    let ExitCallback = function(jobinfo.exit_callback)
else
    let ExitCallback = jobinfo.exit_callback
endif
call ExitCallback(callback_dict)
```

plugin/neomake.vim
```diff
@@ -1,37 +1,11 @@
...
-function! s:NeomakeSh(sh_command)
-    let custom_maker = neomake#utils#MakerFromCommand(&shell, a:sh_command)
-    let custom_maker.name = 'sh: '.a:sh_command
-    let custom_maker.remove_invalid_entries = 0
-    let enabled_makers =  [custom_maker]
-    call neomake#Make({'enabled_makers': enabled_makers})
-endfunction
-
-command! -nargs=* -bang -bar -complete=customlist,neomake#CompleteMakers Neomake call s:NeomakeCommand(<bang>1, [<f-args>])
+command! -nargs=* -bang -bar -complete=customlist,neomake#CompleteMakers Neomake call neomake#WithCallback(<bang>1, [<f-args>])
```
benekastah: I think it won't work to pass the callback in the command here. Right now, `:Neomake` only takes a list of commands as arguments and we shouldn't break that expectation. Instead, make it configurable like this:

```
let g:neomake_python_pylint_exit_callback = 'MyExitCallback'
```

Add `exit_callback` to the `defaults` dict to get this working: https://github.com/benekastah/neomake/pull/162/files#diff-1ceaba5126b0cf477e72f87cb2767af0L150

plugin/neomake.vim
```diff
@@ -1,37 +1,11 @@
...
 " These commands are available for clarity
 command! -nargs=* -bar -complete=customlist,neomake#CompleteMakers NeomakeProject Neomake! <args>
 command! -nargs=* -bar -complete=customlist,neomake#CompleteMakers NeomakeFile Neomake <args>
 
-command! -nargs=+ -complete=shellcmd NeomakeSh call s:NeomakeSh(<q-args>)
+command! -nargs=+ -complete=shellcmd NeomakeSh call neomake#ShWithCallback(<q-args>)
```
benekastah: I don't think we can add a new kind of argument to this command, unfortunately. How would we know if we should interpret an argument as part of the command or as a callback?

benekastah:
Sorry for making you wait. Just added a bunch of comments. Thanks for putting this PR up! Happy to keep looking at it as you make updates.

plugin/neomake.vim
critiqjo: Callbacks are never meant to be passed through the commands. Since `<q-args>` creates only a single argument, and `<f-args>` are surrounded by square brackets, the third argument will always be empty when invoked from the commands `:Neomake` and `:NeomakeSh`.

Callbacks can only be requested by directly calling `neomake#WithCallback` and `neomake#ShWithCallback`.

critiqjo:
>`has_next` is an internal property. I don't think it should be exposed. It's only used in certain cases (mainly when serializing multiple jobs, which isn't done by default).

Correct me if I am wrong -- I assumed there is a possibility for a single `:Neomake` command (or a call to `neomake#WithCallback`) to spawn multiple maker jobs in series (eg. `gcc -c foo.c`, followed by `gcc -o out foo.o bar.o`)

If there is no such possibility, I will remove the `has_next` option altogether.

critiqjo:
I think there is another confusion that I need to clear. This feature will allow another plugin to use Neomake, without wanting the user to call `:Neomake` manually.

In case of lldb.nvim, when the user starts the "debug session" by issuing, say `:LLsession debug`, it will (if preconfigured by the user) invoke `neomake#WithCallback` and wait for it complete. If it gets status `0`, it will go to debug mode (when a new tab with a lot of windows may get created) and launch the target.

autoload/neomake.vim
```diff
@@ -578,3 +594,28 @@ function! neomake#CompleteMakers(ArgLead, CmdLine, CursorPos)
                     \ "v:val =~? '^".a:ArgLead."'")
     endif
 endfunction
+
+function! neomake#WithCallback(file_mode, enabled_makers, ...)
+    let options = a:0 ? { 'exit_callback': a:1 } : {}
```
benekastah: Won't this apply `exit_callback` to every maker in `enabled_makers`. So if you do: `call neomake#WithCallback(0, ['a', 'b', 'c'])` then `exit_callback` will be called three times, right?

benekastah:
Ok, I think I understand a bit better. Can we not rename the functions though? I'm not sure it makes sense to name the function after an optional argument -- it makes it seem as if that argument were its main purpose. Also, instead of passing a callback as the optional argument, why not a base options dict? That would be more flexible.

benekastah:
I think this should be good to go after we sort the naming issue out. Thanks for doing this!

critiqjo:
You are right about the naming issue. I named it so because it reads "neomake with callback" and "neomake sh with callback", and the caller of those functions (except from the command definitions, which is part of the plugin itself) will almost surely want a callback, otherwise they could've just invoked the command, right?

If you find better names, I'll gladly change them.

benekastah:
It occurs to me that everything you want to do can be done with the already exposed `neomake#Make`. Since `neomake#WithCallback` and `neomake#ShWithCallback` were originally meant only to serve as utility functions to construct a command, maybe we could leave those functions unchanged and instead use `neomake#Make`. Would that be acceptable? That simplifies this pull request since you only have to add the callback functionality at that point.

critiqjo:
Yes, that is indeed a possibility, but if someone needed the exact same functionality as the commands does, wouldn't that mean a duplication of effort? I'm cool with either way...

benekastah:
A fair point. How about we do this: Rename `neomake#Make` to `s:Make` (we need to verify that it is never called outside of its source file) and then rename `neomake#WithCallback` to `neomake#Make`. Then we can rename `neomake#ShWithCallback` to `neomake#Sh`.

The final thing we'll want to do is document `neomake#Make` and `neomake#Sh` well so that plugin authors don't have to dig around the source to figure out how to use neomake as a library.

critiqjo:
Sounds good!

But there's still a concern -- cancellation of jobs.
What I would suggest is -- when either of those calls is issued, a unique id should be returned to the caller, using which a cancellation request can be issued.

`:NeomakeListJobs` should be more verbose and should show this id as well, which can be passed to, say, `:NeomakeCancelJob` to manually cancel it.

When cancelled, the callback should be invoked specifying so.

PS: How about neomake#r (read neomaker :smile:!!) instead of neomake#Make ?!

PPS: I can't continue working on this since classes started this week. :disappointed: I'm barely getting time to work on extra-curriculars. Sorry...

critiqjo:
Closing in favor of #227

PR #164, opened 2015-08-04
Implement jsxhint maker

rstacruz:
Adds support for https://github.com/STRML/JSXHint. It can be used via:

```vim
let g:neomake_javascript_enabled_makers = ['jsxhint']
```

### DIFF
--- a/autoload/neomake/makers/ft/javascript.vim
+++ b/autoload/neomake/makers/ft/javascript.vim
@@ -11,6 +11,13 @@ function! neomake#makers#ft#javascript#jshint()
         \ }
 endfunction
 
+function! neomake#makers#ft#javascript#jsxhint()
+    return {
+        \ 'args': ['--verbose'],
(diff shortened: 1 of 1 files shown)


rstacruz:
hmm, maybe this is better implemented using `let g:neomake_javascript_jsxhint_maker = { .. }`, not sure.

rstacruz:
oh, actually it looks like there's built in support when the filetype is `jsx`!

benekastah:
Yup :)  If you want, you can add this to the javascript file:

```
function! neomake#makers#ft#javascript#jsxhint()
    return neomake#makers#ft#jsx#jsxhint()
endfunction
```

rstacruz:
hm. you can't really define `function! neomake#makers#ft#javascript#jsxhint()` without modifying the `neomake` bundle though (ie, can't just put it in my vimrc).

benekastah:
Ah. What I meant is that would be an acceptable change to add to neomake proper.

benekastah:
Tracking that in #170

PR #169, opened 2015-08-22
Add scalastyle checker

xanderdunn:
This simply adds the scalastyle checker as an option.  It's working on my setup.


benekastah:
Thanks!

PR #176, opened 2015-09-12
fixing unterminated if statements & scala improvements

vito-c:
Current scala checker is broken. This will fix it :)

### DIFF
--- a/autoload/neomake/makers/ft/scala.vim
+++ b/autoload/neomake/makers/ft/scala.vim
@@ -1,12 +1,23 @@
 " vim: ts=4 sw=4 et
 function! neomake#makers#ft#scala#EnabledMakers()
-    let makers = ['scalac']
-    if neomake#utils#Exists('scalastyle')
+    let makers = get(g:, 'neomake_scala_fast', 0) ? ['fsc'] : ['scalac']
+    if neomake#utils#Exists('scalastyle') && get(g:, 'neomake_scala_style', 1)
         call add(makers, 'scalastyle')
     endif
     return makers
 endfunction
 
+function! neomake#makers#ft#scala#fsc()
+    return {
+        \ 'args': [
+            \ '-Ystop-after:parser'
+        \ ],
+        \ 'errorformat':
+            \ '%E%f:%l: %trror: %m,' .
+            \ '%Z%p^,' .
+            \ '%-G%.%#'
+        \ }
+endfunction
 function! neomake#makers#ft#scala#scalac()
     return {
         \ 'args': [
--- a/autoload/neomake/makers/ft/scala.vim
+++ b/autoload/neomake/makers/ft/scala.vim
@@ -1,7 +1,8 @@
 " vim: ts=4 sw=4 et
 function! neomake#makers#ft#scala#EnabledMakers()
-    let makers = get(g:, 'neomake_scala_fast', 0) ? ['fsc'] : ['scalac']
-    if neomake#utils#Exists('scalastyle') && get(g:, 'neomake_scala_style', 1)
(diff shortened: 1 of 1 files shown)


vito-c:
Added ability to disable scalastyle and a flag to use either scalac or the fast compiler

autoload/neomake/makers/ft/scala.vim
```diff
@@ -1,11 +1,23 @@
 " vim: ts=4 sw=4 et
 function! neomake#makers#ft#scala#EnabledMakers()
-    let makers = ['scalac']
-    if neomake#utils#Exists('scalastyle')
+    let makers = get(g:, 'neomake_scala_fast', 0) ? ['fsc'] : ['scalac']
+    if neomake#utils#Exists('scalastyle') && get(g:, 'neomake_scala_style', 1)
```
benekastah: Currently disabling scalastyle is already possible by setting your enabled makers manually:

```
let g:neomake_scala_enabled_makers = ['scalac']
```

I prefer this method over adding another configuration variable, which must then be documented. I already think there's too many configuration variables and it makes the docs a little hard to read through quickly.

autoload/neomake/makers/ft/scala.vim
```diff
@@ -1,11 +1,23 @@
 " vim: ts=4 sw=4 et
 function! neomake#makers#ft#scala#EnabledMakers()
-    let makers = ['scalac']
-    if neomake#utils#Exists('scalastyle')
+    let makers = get(g:, 'neomake_scala_fast', 0) ? ['fsc'] : ['scalac']
```
benekastah: Instead of adding a configuration variable, how about something like this:

```
if neomake#utils#Exists('fsc')
    let makers = ['fsc']
else
    let makers = ['scalac']
endif
```

If this is not to a user's liking, they always have the ability to set the enabled makers themselves (see other comment).

autoload/neomake/makers/ft/scala.vim
```diff
@@ -1,11 +1,23 @@
...
+    return {
+        \ 'args': [
+            \ '-Ystop-after:parser'
+        \ ],
+        \ 'errorformat':
+            \ '%E%f:%l: %trror: %m,' .
+            \ '%Z%p^,' .
+            \ '%-G%.%#'
+        \ }
+endfunction
```
benekastah: It's nice to have a single line of whitespace between functions for better readability.

benekastah:
Sorry for making you wait. I added some comments. Once those are addressed, I'd be happy to merge!

autoload/neomake/makers/ft/scala.vim
vito-c: I believe when you install scala you will have both on your path. So you'd have to `let g:neomake_scala_enabled_makers = ['fsc']` the default should be scalac.

autoload/neomake/makers/ft/scala.vim
vito-c: without the config variable it makes it a little less apparent that fsc is available

benekastah: Hmm. Any harm in running both? If that won't work, then a configuration variable is ok, but be sure to add an entry for it in the docs.

benekastah: On second thought, I don't think it's the right thing to add a configuration option for every checker that's not enabled by default. Perhaps it's a good time to set up a wiki that explains how to get set up for each file type so this info will be googleable.

vito-c:
anymore feedback? :)

benekastah:
Thanks for the push there.

PR #178, opened 2015-09-13
add missing :endif

euclio:


vito-c:
this is a dupe of https://github.com/benekastah/neomake/pull/176 I also added some features you might want to check out.

benekastah:
thanks!

PR #184, opened 2015-09-21
Fixed tslint maker args

mkusher:
Fixes this issue:
```
-f option is no longer available. Supply files directly to the tslint command instead.
```


benekastah:
Thanks!

PR #191, opened 2015-09-30
Add jade-lint support

elentok:


autoload/neomake/makers/ft/jade.vim
```diff
@@ -0,0 +1,18 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#jade#EnabledMakers()
+    let makers = []
+    if neomake#utils#Exists('jade-lint')
```
benekastah: No need to call this yourself. Neomake won't try to invoke this maker if it doesn't exist. This function is only needed in special circumstances.

benekastah: See https://github.com/benekastah/neomake/blob/master/autoload/neomake/makers/ft/javascript.vim for a typical example of this function.

elentok: does it check for the name of the maker or the name of the "exe" arg?
(because the maker is named "jadelint" while the command is "jade-lint")

benekastah: It checks for the "exe" arg.

elentok: done (sorry it took so long)

benekastah:
Thanks!

Issue #196, opened 2015-10-08
Allow file to be excluded from file maker


jason0x43:
It seems like project level builders such as `go build` and `tsc` would generally work fine as filetype builders if the builder definition had a way of excluding the filename. Consider the `go` plugin. It currently includes filetype builders for `go` and `golint`. It's nice to use the filetype mode with go files since you can see both syntax errors (from `go build`) and style errors (from `golint`). However, it doesn't make sense to run `go build` with a filename argument since you'll get spurious errors if the current file uses things defined in other files, so using `Neomake` (not `Neomake!`) with go files as it stands can be kind of annoying.

If the filename could be excluded, either by providing something like a 'nofile' option or by making the ':p' placeholder non-optional for filename inclusion, then filetype-specific project builders like `go build` could run with other filetype builders. I mean, `go build` *is* a filetype builder,it just makes more sense to run it without a file argument.

benekastah:
This is a good idea. As always, not sure when I can get around to fixing this myself, but always willing to review a pull request :)

jason0x43:
I'll submit a PR this weekend. I implemented a `nofile` option locally, which worked fine with Go. Making the ':p' flag a requirement would be cleaner, but it also breaks existing makers.

blueyed:
Fixed/closed by https://github.com/neomake/neomake/pull/198.

PR #198, opened 2015-10-12
Allow file to be excluded in filetype makers

jason0x43:
This PR adds an `append_file` option to the filetype maker specification. If `append_file` is 0, the file path will not be implicitly added to the arguments list. This option is useful for makers such as `go build` and `tsc` that should be run for certain filetypes but are typically run without a file argument. `append_file` is 1 by default to preserve backward compatibility.

The `go` filetype builder has been modified to set `append_file` to 0.

References #196

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -67,7 +67,7 @@ function! neomake#MakeJob(maker) abort
     let jobinfo.maker = a:maker
 
     let args = a:maker.args
-    let append_file = a:maker.file_mode && index(args, '%:p') <= 0
+    let append_file = a:maker.file_mode && index(args, '%:p') <= 0 && get(a:maker, 'append_file', 1)
     if append_file
         call add(args, '%:p')
     endif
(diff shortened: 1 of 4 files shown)


autoload/neomake.vim
```diff
@@ -67,8 +67,7 @@ function! neomake#MakeJob(maker) abort
     let jobinfo.maker = a:maker
 
     let args = a:maker.args
-    let append_file = a:maker.file_mode && index(args, '%:p') <= 0
-    if append_file
+    if a:maker.file_mode && index(args, '%:p') <= 0 && !(has_key(a:maker, 'append_file') && !a:maker.append_file)
```
benekastah: A slightly simpler (and more readable) way to write this is:

```
if a:maker.file_mode && index(args, '%:p') <= 0 && get(a:maker, 'append_file', 1)
```

benekastah:
Thanks! I suggested a couple of cleanups, so once those are taken care of I'll be happy to merge :)

jason0x43:
Ah, I hadn't thought of `get()` (I haven't written a ton of vim script). Regarding the `append_file` variable, I removed it because I felt its presence made the code a bit more difficult to read. I mean, I'm just not used to adding variables solely for the purpose of documentation, so when I saw a variable being assigned but only used once, I thought I must be missing something. :) Now that you mention it, though, I notice that doc variables seem to be a standard pattern in the code. Consistency FTW.

benekastah:
LG! Thanks for the contribution!

PR #205, opened 2015-11-02
Add eslint_d maker

flannelhead:
This PR adds the `eslint_d` JavaScript maker. It wraps `eslint` and avoids delays by preloading the required modules (see [here](https://github.com/mantoni/eslint_d.js)). Its usage is identical to `eslint`.


benekastah:
thanks for the contribution!

PR #206, opened 2015-11-03
Support for rubber latex compilation

languitar:
This adds support for rubber and rubber-info to compile and analyze latex documents.


benekastah:
Cool, thanks!

languitar:
Thanks. Btw, is there any way to decide in which order makers are executed?

benekastah:
Makers will be executed in the order they appear in the enabled makers list. So if you want to customize that for yourself, you can define it:

```
let g:neomake_latex_enabled_makers = ['rubberinfo', 'lacheck']
```

PR #218, opened 2015-11-10
Allow b: variables for overriding maker settings

### LINKED ISSUE #217: Respect b:variables when overriding or augmenting variables (like Syntastic does)
Loving Neomake so far! Thank you for working on it.

I used the following to pass different arguments to `eslint` for `*.es6` files only.

```vim
autocmd BufRead,BufNewFile *.es6 call s:setupES6File()

" For es6 files, run this every time it gets opened, and apply these buffer " variables
function! s:setupES6File()
  setfiletype javascript " Set filetype

  " Set the checkers for the current buffer to eslint

  " This works, but sets it globally for all files :|
  let g:neomake_javascript_eslint_args = ['-c', '~/.eslintes6rc', '-f', 'compact']  
endfunction
```

I really wish I could use this, like I could with syntastic, so it only gets applied for the current file.
```vim
  " What I really want is this, like I used to use with syntastic: 
  let b:neomake_javascript_eslint_args = ['-c', '~/.eslintes6rc', '-f', 'compact']
  " let b:syntastic_javascript_eslint_args = "-c ~/.eslintes6rc"
```

If you're open to this change, I can probably figure it out and issue a PR. Thoughts?

jamischarles:
Closes #217 

Previously setting overrides for just the current buffer wasn't possible:
  `let b:neomake_javascript_eslint_args = ['-c', '.eslintes6rc', '-f', 'compact']`

This now allows us to set different rules easily for a certain buffer
type.

```vim
autocmd BufRead,BufNewFile *.es6 call s:setupES6File()

let g:neomake_javascript_enabled_makers = ['eslint']

let g:neomake_javascript_eslint_args = ['-c', '.eslintrc']

function! s:setupES6File()
  setfiletype javascript " Set filetype
  " For the current buffer, use the special es6 eslint ruleset
  let b:neomake_javascript_eslint_args = ['-c', '.eslintes6rc']
endfunction
```

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -158,14 +158,16 @@ function! neomake#GetMaker(name_or_maker, ...) abort
         if len(fts)
             for ft in fts
                 let config_var = 'neomake_'.ft.'_'.maker.name.'_'.key
-                if has_key(g:, config_var)
+                if has_key(g:, config_var) || has_key(b:, config_var)
                     break
                 endif
             endfor
         else
             let config_var = 'neomake_'.maker.name.'_'.key
         endif
-        if has_key(g:, config_var)
+        if has_key(b:, config_var)
+            let maker[key] = copy(get(b:, config_var))
+        elseif has_key(g:, config_var)
             let maker[key] = copy(get(g:, config_var))
         elseif !has_key(maker, key)
             let maker[key] = defaults[key]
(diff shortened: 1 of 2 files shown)


benekastah:
Cool, thanks for this! Can you add a bit [to the docs](https://github.com/benekastah/neomake/blob/master/doc/neomake.txt#L178) mentioning that this works with `b:` variables now?

jamischarles:
@benekastah Updated the docs. Not sure if you wanted a separate entry, or if this works. Let me know!

benekastah:
Thanks!

PR #219, opened 2015-11-11
Add clang-tidy support

bchretien:
`g:neomake_c_clangtidy_args_conf` needs to be properly set, especially with the `-p` option indicating the build directory containing the compile command database.

### DIFF
--- a/autoload/neomake/makers/ft/c.vim
+++ b/autoload/neomake/makers/ft/c.vim
@@ -46,20 +46,12 @@ function! neomake#makers#ft#c#gcc()
         \ }
 endfunction
 
+" The -p option followed by the path to the build directory should be set in
+" the maker's arguments. That directory should contain the compile command
+" database (compile_commands.json).
 function! neomake#makers#ft#c#clangtidy()
-    " Default arguments.
-    let l:args = []
-
-    " Add user-defined arguments if some are set.
-    " The -p option followed by the path to the build directory is expected.
-    " That directory should contain the compile command database
-    " (compile_commands.json).
-    if exists("g:neomake_c_clangtidy_args_conf")
-        let l:args += g:neomake_c_clangtidy_args_conf
-    endif
     return {
         \ 'exe': 'clang-tidy',
-        \ 'args': l:args,
         \ 'errorformat':
             \ '%E%f:%l:%c: fatal error: %m,' .
             \ '%E%f:%l:%c: error: %m,' .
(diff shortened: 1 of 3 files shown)


autoload/neomake/makers/ft/c.vim
```diff
@@ -1,11 +1,17 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#c#EnabledMakers()
+    let makers = []
```
benekastah: This can just be `let makers = ['clang-tidy']`. You only need to do the exists check yourself if there are special requirements (e.g. only enable maker x if maker y doesn't exist).

autoload/neomake/makers/ft/c.vim
```diff
@@ -39,3 +45,26 @@ function! neomake#makers#ft#c#gcc()
...
+function! neomake#makers#ft#c#clangtidy()
+    " Default arguments.
+    let l:args = []
+
+    " Add user-defined arguments if some are set.
+    " The -p option followed by the path to the build directory is expected.
+    " That directory should contain the compile command database
+    " (compile_commands.json).
+    if exists("g:neomake_c_clangtidy_args_conf")
+        let l:args += g:neomake_c_clangtidy_args_conf
```
benekastah: This isn't needed. Neomake already provides a way to modify maker arguments through a configuration variable:

```
" This already works for any maker, see :help g:neomake_<ft>_<name>_<property>
let g:neomake_c_clangtidy_args = []
```

benekastah:
Added a couple of comments. Once those are resolved I'll be happy to merge!

autoload/neomake/makers/ft/c.vim
bchretien: The special requirements are that `clang-tidy` should only be added if `clang` and `clang-tidy` are available on the system. Since both `clang` and `clang-tidy` makers can coexist (they achieve different objectives), I do the relevant tests with some `neomake#utils#Exists` calls right after this.

autoload/neomake/makers/ft/c.vim
bchretien: Oh, I didn't realize that, thanks!

bchretien: Now that I think about it, I actually tested that when I started using neomake, but I didn't realize that `neomake_c_clangtidy_args` was not propagating to the C++ maker, and I had to set `neomake_cpp_clangtidy_args` as well.

benekastah:
Thanks!

PR #220, opened 2015-11-11
Make ghc-mod's messages errors by default

expipiplus1:
When ghc-mod doesn't include `Warning` in the error message it's an error


benekastah:
Thanks!

Issue #153, opened 2015-06-29
newlines are not parsed correctly


nerdrew:
Newlines in the `makeprg` output are incorrectly parsed.

Example:

```
let g:neomake_makeprg_remove_invalid_entries = 0
let &l:makeprg = 'echo -n .; sleep 0.5; echo -n .; sleep 0.5; echo .; sleep 0.5; echo -n .'
:Neomake!
```

Actual Output:

```
|| .
|| .
|| .
|| 
|| .
```

Expected Output:

```
|| ...
|| .
```

It looks like newlines are passed into `on_std{out,err}` as single character items in the `a:data` list.

Neki:
Thanks for the bug report.

Relevant logs: 

```
2015-06-30T17:31:56+0200 Log level 2: Neomake: Starting: /usr/bin/zsh -c echo -n .; sleep 0.5; echo -n .; sleep 0.5; echo .; sleep 0.5; echo -n .
2015-06-30T17:31:56+0200 Log level 3: Neomake: makeprg stdout: .
2015-06-30T17:31:56+0200 Log level 3: Neomake: makeprg stdout done.
2015-06-30T17:31:56+0200 Log level 3: Neomake: makeprg processing 1 lines of output
2015-06-30T17:31:57+0200 Log level 3: Neomake: makeprg stdout: .
2015-06-30T17:31:57+0200 Log level 3: Neomake: makeprg stdout done.
2015-06-30T17:31:57+0200 Log level 3: Neomake: makeprg processing 1 lines of output
2015-06-30T17:31:57+0200 Log level 3: Neomake: makeprg stdout: .
2015-06-30T17:31:57+0200 Log level 3: Neomake: makeprg stdout: 
2015-06-30T17:31:57+0200 Log level 3: Neomake: makeprg stdout done.
2015-06-30T17:31:57+0200 Log level 3: Neomake: makeprg processing 2 lines of output
2015-06-30T17:31:58+0200 Log level 3: Neomake: makeprg stdout: .
2015-06-30T17:31:58+0200 Log level 3: Neomake: makeprg stdout done.
2015-06-30T17:31:58+0200 Log level 3: Neomake: makeprg processing 1 lines of output
2015-06-30T17:31:58+0200 Log level 1: Neomake: makeprg complete with exit code 0
```

Neovim is not passing Neomake entire lines. I don't know if it is intended or not.

nerdrew:
[...]

PR #221, opened 2015-11-11
Append captured lines from makers correctly

sanmiguel:
Fixes issue #153
As per https://github.com/neovim/neovim/issues/3555

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -443,7 +443,10 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
         let last_event_type = get(jobinfo, 'event_type', a:event_type)
         let jobinfo.event_type = a:event_type
         if has_key(jobinfo, 'lines')
-            call extend(jobinfo.lines, lines)
+            " As per https://github.com/neovim/neovim/issues/3555
+            let jobinfo.lines = jobinfo.lines[:-2]
+                        \ + [jobinfo.lines[-1] . get(lines, 0, '')]
+                        \ + lines[1:]
         else
             let jobinfo.lines = lines
         endif


sanmiguel:
Hi - I am not sure if this is the appropriate way to do this: you were using `extend()` which implies you expect `lines` to sometimes be a dict, but I couldn't find anywhere that might happen.

Anyway - I used the solution @ZyX-I laid out in neovim/neovim#3555 for correctly appending to rebuild lines even if they arrive in partial flushes.

benekastah:
If the last line captured was a full line rather than a partial line, won't this result in two lines being concatenated together and treated as one line? (Forgive me if I'm mistaken, I haven't kept fully abreast of the changes made in neovim since I implemented this).

sanmiguel:
Apologies I have been busy lately and haven't had a chance to reply to this.

This works because `nvim` does something I find a little surprising (IMHO) when handing flushed lines to a jobcontrol callback. @tarruda described it quite succinctly in neovim/neovim#3555 : 

```
It's simple: for every list you receive in a callback, all items except the first represent newlines. 
```

When a complete line (i.e. some text followed by a newline) is flushed to stdout (or stderr) and nvim, it's presented to the callback as `[ 'some text', '' ]`. We can see this in action by running the following `jobcontrol.vim` (again taken from neovim/neovim#3555) :

``` vim
let g:full_stdout = ['']
let g:raw_stdout = []
function! s:JobHandler(job_id, data, event)
  if a:event isnot# 'stdout'
    echo 'g:raw_stdout: ' . string(g:raw_stdout)
    return
  endif
  call add(g:raw_stdout, a:data)
  let g:full_stdout = g:full_stdout[:-2] + [g:full_stdout[-1] . get(a:data, 0, '')] + a:data[1:]
endfunction
let s:callbacks = {
      \ 'on_stdout': function('s:JobHandler'),
      \ 'on_exit': function('s:JobHandler')
      \ }

let job = jobstart(['bash', '-c', 'echo ..; sleep 0.6; echo ..; sleep 0.6; echo ..'], s:callbacks)
```

```
$ nvim -u NONE -S jobcontrol.vim
...
[['..', ''], ['..', ''], ['..', '']]
```

So we can see the raw lines we receive for each full line are actually presented as the text, followed by an empty string representing a newline.

Actually even the way it was suggested to solve in neovim/neovim#3555 is not 100% accurate as, in cases like the example, what we actually end up with in `g:full_stdout` is `[ '..', '..', '..', '' ]` i.e. an extra blank line at end. Mostly this won't cause an issue but if your `maker.errorformat` does something with blank lines it might cause an issue. We could probably do something durty just to lop off a single trailing blank line (if one exists) but something more elegant must be possible....

sanmiguel:
Rock on! Thanks @benekastah !

benekastah:
Thanks for the explanation! Let's see if the trailing newline thing becomes a problem in the real world.

sanmiguel:
You're welcome! It took me so long to answer because when you asked that question I couldn't immediately explain why it worked beyond "because someone else says it does". Had to experiment a bit, which needed time. I'm thinking the erroneous trailing blank line is a small price to pay and will be reasonably unlikely to cause any problems IRL.

PR #225, opened 2015-11-14
remove zsh shellcheck maker (shellcheck doesn't support zsh)

lucy:
It was removed in https://github.com/koalaman/shellcheck/commit/ed56a837c33caba7788b96a70f6c0f7eb5642e27 and never worked well.


benekastah:
Thanks!

PR #227, opened 2015-11-17
[RFC] Finish implementing exit callbacks

aykamko:
Implements exit callbacks as specified in #162 

Finishes what @critiqjo started.

Also adds documentation.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -21,6 +21,13 @@ function! neomake#ListJobs() abort
     endfor
 endfunction
 
+function! neomake#CancelJob(job_id) abort
+    if !has_key(s:jobs, a:job_id)
+        return
+    endif
+    call jobstop(a:job_id)
+endfunction
+
 function! s:JobStart(make_id, exe, ...) abort
     let argv = [a:exe]
     let has_args = a:0 && type(a:1) == type([])
@@ -100,6 +107,8 @@ function! neomake#MakeJob(maker) abort
     if has_key(a:maker, 'cwd')
         exe 'cd' fnameescape(old_wd)
     endif
+
+    return jobinfo.id
 endfunction
 
 function! neomake#GetMaker(name_or_maker, ...) abort
@@ -210,7 +219,7 @@ function! neomake#GetEnabledMakers(...) abort
     return filter(keys(union), 'union[v:val] ==# l')
 endfunction
 
-function! neomake#Make(options) abort
+function! s:Make(options) abort
     call neomake#signs#DefineSigns()
     call neomake#statusline#ResetCounts()
 
@@ -250,6 +259,7 @@ function! neomake#Make(options) abort
     endif
 
     let serialize = get(g:, 'neomake_serialize')
+    let job_ids = []
     for name in enabled_makers
         let maker = neomake#GetMaker(name, ft)
         let maker.file_mode = file_mode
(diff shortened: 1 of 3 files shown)


critiqjo:
Great! :+1:

doc/neomake.txt
```diff
@@ -258,6 +266,43 @@ section of the vim-airline |statusline|. Defaults to 1.
...
 functions useful.
 
+*neomake#Make(filemode,* *makers[,* *callback])*
+This function is called by |:Neomake(!)| command. It runs all the *makers*
+specified, in order. If *filemode* is 1, then the current file is used as input
+to the makers. |neomake#Make| returns an array of the job ids that it creates
+via 'jobstart'; you can potentially cancel these jobs with |neomake#CancelJob|.
+
+It also accepts a third, optional *callback* argument that
+is called when the command exists.  The *callback* is given the following
```
benekastah: "exists" should be "exits"

doc/neomake.txt
```diff
@@ -258,6 +266,43 @@ section of the vim-airline |statusline|. Defaults to 1.
...
+    \ 'has_next': <true if another maker follows, false otherwise> }
+
+*neomake#Sh(command[,* *callback])*
+This function is called by the |:NeomakeSh| command. It runs the specified
+shell *command* according to 'shell'. |neomake#Sh| returns the single job id
+that it creates via 'jobstart'; you can potentially cancel this job with
+|neomake#CancelJob|.
+
+It also accepts a second, optional
+*callback* argument that is called when the command exists. The *callback* is
```
benekastah: "exists" should be "exits" here too :)

doc/neomake.txt
```diff
@@ -258,6 +266,43 @@ section of the vim-airline |statusline|. Defaults to 1.
...
+given the following dictionary as its sole argument: >
+    { 'status': <exit status of command>,
+    \ 'name': 'sh: <command>',
```
benekastah: Instead of adding a "sh: " prefix to the name, why not make another property on the dict like: `'sh': <true if called from :Sh>`?

autoload/neomake.vim
```diff
@@ -210,7 +219,7 @@ function! neomake#GetEnabledMakers(...) abort
     return filter(keys(union), 'union[v:val] ==# l')
 endfunction
 
-function! neomake#Make(options) abort
+function! s:Make(options) abort
```
benekastah: I wish we could think of a better name for this function. I have no suggestions, though, so this will probably stick.

autoload/neomake.vim
```diff
@@ -476,6 +491,19 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
             endif
         endif
         let status = a:data
+        if has_key(maker, 'exit_callback')
+            try
```
benekastah: This try block is too broad. I think we should wrap only the lines that can err in a try block, and then we should only catch the errors we expect. This is a PITA in vimscript, but there are examples in this file that should point you in the right direction.

The danger in keeping this try statement too broad is that we can accidentally hide errors that we actually want to fix, which makes it difficult to track down when they cause a bug.

benekastah:
LG! Once my comments are addressed I'll be happy to merge.

jamischarles:
@aykamko You still working on this? Would love to get his feature.

aykamko:
@benekastah Apologies for the delay, I believe I've fixed your nits! Wasn't 100% about fixing your comment on that try block, but hopefully these changes are better

@jamischarles I'll try to finish this up ASAP.

benekastah:
Thanks for doing this!

jamischarles:
Yay! :tada:. Thanks @aykamko

jamischarles:
@benekastah Does this mean we can now pretty easily do https://github.com/benekastah/neomake/issues/173? Does this commit lay the groundwork for that?

PR #229, opened 2015-11-19
Reek maker for ruby files

neckhair:
[Reek](https://github.com/troessner/reek) detects code smells in Ruby files. This PR adds support to integrate detections in neovim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,9 @@ Python:
 - python
 
 Ruby:
+- mri
 - rubocop
+- reek
 
 C:
 - clang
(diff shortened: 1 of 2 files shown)


benekastah:
Thanks!

neckhair:
Thanks for neomake! And thanks for merging.

zamith:
Is it possible to only run this maker if reek is in the Gemfile? The reasoning is that I would have to consciously opt-in to having all of the warnings from reek, which are both more and more questionable than the one's in rubocop.

neckhair:
I would rather check if the binaries (rubocop or reek) are present. I never have them in any Gemfile but always install them for all ruby versions. Or maybe add some configuration the deactivate reek or rubocop.

zamith:
My suggestion of the Gemfile, was so that it could be controlled on a project basis, not ruby version. Maybe a `.neomakerc` file or something, to configure it per project? Not sure if there's such a thing already.

PR #232, opened 2015-11-21
Add clang-check support

kyrias:
Heavily inspired by the syntastic clang-check support.

First time I've done any VimL scripting, so do point out anything that is iffy.

### DIFF
--- a/autoload/neomake/makers/ft/c.vim
+++ b/autoload/neomake/makers/ft/c.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#c#EnabledMakers()
-    let makers = executable('clang') ? ['clang', 'clangtidy'] : ['gcc']
+    let makers = executable('clang') ? ['clang', 'clangtidy', 'clangcheck'] : ['gcc']
     call add(makers, 'checkpatch')
     return makers
 endfunction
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/c.vim
```diff
@@ -1,7 +1,9 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#c#EnabledMakers()
-    if neomake#utils#Exists('clang')
+    if neomake#utils#Exists('clang-check')
+        return ['clangcheck']
+    elseif neomake#utils#Exists('clang')
         return ['clang']
```
benekastah: I think this could just be

```
return ['clang', 'clangcheck']
```

Unless those checkers are mutually exclusive. Neomake won't run a maker that doesn't exist, so you generally don't need to check for them.

autoload/neomake/makers/ft/c.vim
```diff
@@ -22,6 +24,26 @@ function! neomake#makers#ft#c#clang()
...
 
+function! neomake#makers#ft#c#clangcheck()
+    if !exists('g:neomake_c_clangcheck_config_file')
+        let g:neomake_c_clangcheck_config_file = '.neomake_clangcheck_config'
+    endif
+
+    return {
+        \ 'exe': 'clang-check',
+        \ 'args': ['%:p', '--', '-fdiagnostic-format=clang']
+                  \ + neomake#utils#ReadConfig(g:neomake_c_clangcheck_config_file),
```
benekastah: No need for this config file. See `:help g:neomake_<ft>_<name>_<property>`.

kyrias: Issue with that is that different projects need different flags, and having a config file in the project is the only solution to that that feels sane to me.

benekastah: That's true, but it's not a problem neomake aims to solve. Other makers don't all have their own config file option in neomake, so this is inconsistent with the rest of the project. I have this in my vimrc as a way to do per-project configurations:

```
let g:project_local_vimrc = findfile('.project.vim', '.;')
if filereadable(g:project_local_vimrc)
    execute 'source ' . g:project_local_vimrc
endif
```

blueyed:
@kyrias 
Can you address the issues in this PR and rebase it, please?

blueyed:
@kyrias 
Ping again..

kyrias:
Sorry, had completely forgotten about this, will do.

kyrias:
@blueyed Sorry for taking so long, got distracted by life.  How's this?

blueyed:
Yeah, life's a beach!

That's nice.. assuming that 'clang', 'clangtidy', 'clangcheck' should really all be run if they are all installed.

Thanks a lot!

PR #234, opened 2015-11-23
Add smlnj Maker for Standard ML

prikhi:


autoload/neomake/makers/ft/sml.vim
```diff
@@ -0,0 +1,29 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#sml#EnabledMakers()
```
benekastah: This function can simply be:

```
function! neomake#makers#ft#sml#EnabledMakers()
    return ['smlnj']
endfunction
```

Before we try to use a maker, `neomake#utils#Exists` is called on its `exe` property. We only need to check if things exist directly in functions like this when there are special requirements (e.g. only enable maker x if maker y is not present).

autoload/neomake/makers/ft/sml.vim
```diff
@@ -0,0 +1,29 @@
...
+    if neomake#utils#Exists('sml')
+        call add(makers, 'smlnj')
+    endif
+
+    let s:sml_makers = makers
+    return makers
+endfunction
+
+" This comes straight out of syntastic.
+function! neomake#makers#ft#sml#smlnj
```
benekastah: Any reason why this is `smlnj` instead of `sml`?

prikhi: There are many implementations of standard ml, the sml command is only provided by the SML/NJ implementation(which usually has a package name of `smlnj`).

autoload/neomake/makers/ft/sml.vim
prikhi: OK will fix, just used the python file as a template.

prikhi:
Fixed & rebased

benekastah:
Thanks!

PR #236, opened 2015-11-25
Prevent certain maker args from silently vanishing

DonnieWest:
In some rare cases, args passed to the maker (such as colon separated
classpaths) can be overidden to become an empty string by expand()

This commit fixes this by employing a utility function that is then
passed to map

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -71,7 +71,14 @@ function! neomake#MakeJob(maker) abort
     if append_file
         call add(args, '%:p')
     endif
-    call map(args, 'expand(v:val)')
+    function! s:expandIfValid(arg) abort
+        if expand(a:arg) != ""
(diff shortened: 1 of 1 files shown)


benekastah:
You can escape special characters so `expand()` interprets them literally.

I guess I'm not sure how you're getting your classpath from java into your maker args, so you may need to tweak this, but here's what I'm thinking:

```
let g:neomake_java_javac_args = ['-classpath', 'some\:pie\:here']
```

DonnieWest:
@benekastah makes sense - I encountered this issue when pulling the classpaths from the JavaComplete2 plugin. I'll fiddle with a way to programatically escape the strings - I just wasn't having any luck with built in escaping mechanisms.

The main reason I opened this is that I haven't encountered similar problems with syntastic, but I understand the need for this kind of workaround :)

PR #242, opened 2015-12-07
Add tsx support

ianks:
This PR adds typescript support for TSX. It mostly just leverages the typescript libray for this, but ensures the compile gets the `--jsx preserve` args for tsx files.


benekastah:
Thanks!

PR #248, opened 2015-12-14
Default signs background highlight

eugen0329:
This pull request defines default colors corresponding to the sign column background, thus the default colors will look more natural

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -156,7 +156,7 @@ function! neomake#signs#RedefineSign(name, opts)
 endfunction
 
 function! neomake#signs#RedefineErrorSign(...)
-    let default_opts = {'text': '✖'}
+    let default_opts = {'text': '✖', 'texthl': 'NeomakeErrorDefault'}
     let opts = {}
     if a:0
         call extend(opts, a:1)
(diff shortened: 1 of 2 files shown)


benekastah:
Thanks!

benekastah:
Looks like the `guifg` bit breaks vim 7.4. For practical reasons we try to keep neomake working on vim and neovim. I don't personally have any vim guis, so could you fix this?

demon386:
I got warning after this patch.

```
line   47:
E421: Color name or number not recognized: ctermbg=-1 guibg=
line   48:
E421: Color name or number not recognized: ctermbg=-1 guibg=
line   49:
E421: Color name or number not recognized: ctermbg=-1 guibg=
line   50:
E421: Color name or number not recognized: ctermbg=-1 guibg=
Press ENTER or type command to continue
```

casimir:
@demon386 If you're using GitGutter maybe you can help there #257.

benekastah:
I reverted this, since it's causing problems.

casimir:
@benekastah since it wasn't broken for everyone could it be an opt-in feature instead of nuking it?

eugen0329:
Fish can't be half fresh. If a problem appears at least for one user - it is nothing more than a bug, so this is a bad idea to make this feature optional. Nevertheless, I fix this bugs as soon as possible.

benekastah:
Thanks @eugen0329!

casimir:
Just to clarify by optional I meant guarded by a `has('nvim')` or a Vim version check.

blueyed:
I don't think it has something to do with Vim vs Neovim, but more with the colorscheme being used.

blueyed:
See https://github.com/neomake/neomake/pull/456 for a follow-up.

PR #250, opened 2015-12-15
Added literate haskell support.

dbeecham:
A way to reduce duplicate code would be great. Also, I have not been able to test hdevtools yet, so I've marked it as unsupported in the meantime.

### DIFF
--- a/autoload/neomake/makers/ft/lhaskell.vim
+++ b/autoload/neomake/makers/ft/lhaskell.vim
@@ -1,0 +1,52 @@
+
+function! neomake#makers#ft#lhaskell#EnabledMakers()
+    " Have not tested hdevtools yet!
+    "return ['ghcmod', 'hdevtools', 'hlint']
+    return ['ghcmod', 'hlint'] 
+endfunction
+
+function! neomake#makers#ft#lhaskell#hdevtools()
+    return {
+        \ 'exe': 'hdevtools',
+        \ 'args': ['check', '-g-Wall'],
+        \ 'errorformat':
+            \ '%-Z %#,'.
+            \ '%W%f:%l:%v: Warning: %m,'.
+            \ '%W%f:%l:%v: Warning:,'.
+            \ '%E%f:%l:%v: %m,'.
+            \ '%E%>%f:%l:%v:,'.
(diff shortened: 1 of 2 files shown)


autoload/neomake/makers/ft/lhaskell.vim
```diff
@@ -0,0 +1,52 @@
+
+function! neomake#makers#ft#lhaskell#EnabledMakers()
+    " Have not tested hdevtools yet!
+    "return ['ghcmod', 'hdevtools', 'hlint']
+    return ['ghcmod', 'hlint'] 
+endfunction
+
+function! neomake#makers#ft#lhaskell#hdevtools()
```
benekastah: You could do:

```
function! neomake#makers#ft#lhaskell#hdevtools()
    return neomake#makers#ft#haskell#hdevtools()
endfunction
```

autoload/neomake/makers/ft/lhaskell.vim
```diff
@@ -0,0 +1,52 @@
+
+function! neomake#makers#ft#lhaskell#EnabledMakers()
+    " Have not tested hdevtools yet!
+    "return ['ghcmod', 'hdevtools', 'hlint']
+    return ['ghcmod', 'hlint']
```
benekastah: I think it's fine to enable `hdevtools` as well, as long as it claims to work for literate haskell code.

dbeecham:
Thank you, benekstah... I don't know why I didn't think of that. I added an errorformatline to haskell#hlint which I think is unique to literate haskell, but I figured it does not matter much, it doesn't interfere with the haskell support, and seems like the right place for error format.

There is also apparently an issue with hlint; it does not report correct line numbers in literate haskell. It's an [open issue](https://github.com/ndmitchell/hlint/issues/154). Not sure if neomake should claim hlint support or not...

benekastah:
You could do:

```
function! neomake#makers#ft#lhaskell#hlint()
    let maker = neomake#makers#ft#haskell#hlint()
    let maker.errorformat = 'hlint: In file %f at line %l:%m,' . maker.errorformat
    return maker
endfunction
```

That said, I'm not sure I mind how you

PR #253, opened 2015-12-18
Added postprocessing hook for flake8.

tail:
Flake8 uses the prefix "F" for PyFlakes codes.  These are a mix of
warnings and errors, so we'll just treat them all as errors.


benekastah:
Thanks

PR #264, opened 2016-01-02
Fix Flow Maker

npbee:
The current flow maker errorformat is being flagged as invalid and thus being
removed from the location list (#27).  This results in Vim just saying that the
maker exited with an error code, but not actually listing any errors.  In an
attempt to display errors in the location list, I'm experimenting with different
errorformats.

Flow can output different error strings depending on the arguments passed, one
of which is '--from=vim' which is a little easier to parse for Vim than the
standard output.  However, the resulting message is arguably too cryptic to
understand without more context, hence attempts to try different formats.

### DIFF
--- a/autoload/neomake/makers/ft/javascript.vim
+++ b/autoload/neomake/makers/ft/javascript.vim
@@ -41,15 +41,6 @@ function! neomake#makers#ft#javascript#standard()
 endfunction
 
 function! neomake#makers#ft#javascript#flow()
-    " Multi-line Vim
-    " \ 'errorformat': '%CFile %m%.%#,%AFile "%f"\, line %l\, characters %c-%.%#,%+G%m,%Z%m,%-G%.%#'
-
-    " Single-line Vim - Long
-    " 'errorformat': '%CFile %m%.%#,%AFile "%f"\, line %l\, characters %c-%.%#,%C%m,%Z%m,%-G%.%#'
-
-    " Single-line Vim - Short
-    " \ 'errorformat': '%CFile %.%#,%AFile "%f"\, line %l\, characters %c-%.%#,%C%m,%Z%m,%-G%.%#'
-
     return {
         \ 'args': ['--from=vim'],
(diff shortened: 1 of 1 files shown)


npbee:
Please do not merge yet, just opening this up for discussion for now in case anyone wants to offer some advice on the flow maker output.

autoload/neomake/makers/ft/javascript.vim
```diff
@@ -41,11 +41,17 @@ function! neomake#makers#ft#javascript#standard()
...
+
+    " Single-line Vim - Short
+    " \ 'errorformat': '%CFile %.%#,%AFile "%f"\, line %l\, characters %c-%.%#,%C%m,%Z%m,%-G%.%#'
+
     return {
-        \ 'args': ['check', '--one-line'],
-        \ 'errorformat': '%f:%l:%c\,%n: %m',
-        \ 'mapexpr': mapexpr,
+        \ 'args': ['--from=vim'],
+        \ 'errorformat': '%CFile %.%#,%AFile "%f"\, line %l\, characters %c-%.%#,%C%m,%Z%m'
```
benekastah: I noticed that this errorformat isn't mentioned in the comments above. How does it differ from those? Might want to add it above so that future readers know which is selected and why it was deemed to hold the advantage over the other options. Or, if you are confident in your choice, you could simply remove the comments above :)

npbee:
Yes sorry I just kept those comments in there for ease of switching back and forth, but I was planning on removing them before merging.  I played around with the error formats a bit more and settled on one, so I've removed the comments and this should be ready to merge in if everything else is ok.  Agreed that something is better than nothing at this point so I think this should get things started at least.  Thanks for your work on this!

benekastah:
Sorry, just saw there are conflicts. Once those are fixed I'll merge.

npbee:
Ah so it looks like there was already some discussion about fixing the flow error format and it was solved with a different solution in #265.  The flow maker is working ok now after that patch, so it might be best to not merge this PR and leave things where they are.

benekastah:
Ok, thanks for your effort!

PR #269, opened 2016-01-05
Fix typo in color CLI option for JSCS maker

jfelchner:
The latest JSCS uses `--no-colors` and not `--no-color`


benekastah:
How will this behave with old versions?

jfelchner:
It'll fail. So your choice is: do you want it to fail with previous versions, or do you want it to fail for every new version going forward?

megalithic:
:+1:

PR #276, opened 2016-01-09
fixed rustc command line parameter for rust maker

### LINKED ISSUE #275: rust maker parameter causes wrong error report
when i open my rust source files with neovim, neomake usually have an error on first line if there is an extern crate statement. 

i checked sytastic plugins syntax_checker for rust programs and it uses `-Z parse-only` and does not raise warning for the extern crate lines. 

neomake rust makes does not use this parse-only argument but using `-Z no-trans`. when i checked descriptions for these params: 

```
    -Z                   parse-only -- parse only; do not compile, assemble, or link
    -Z                     no-trans -- run all passes except translation; no output
```

i think to use parse-only for rust make is a better idea.

zekzekus:
fixes #275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 2016-01-10
+- modified compiler parameter for rust maker
+
 ## 2015-04-06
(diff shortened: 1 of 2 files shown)


benekastah:
thanks!

PR #278, opened 2016-01-13
Use case-insensitive compare when placing signs

### LINKED ISSUE #262: Warnings incorrectly identified as errors
neovim(and possibly vim?) can return either upper-case or lower-case types for the qf/loclist type as can be seen here:

https://github.com/neovim/neovim/blob/f0538639c0a58b3ae3a0cd5db49ba24ba8b1a4a3/src/nvim/quickfix.c#L1922

neomake currently does a case sensitive comparison however, which causes everything to be reported as an error, as can be seen here:
https://github.com/benekastah/neomake/blob/6342a7d7e09083a549800a3cdc0ef95358a73ba7/autoload/neomake/signs.vim#L65

I'd submit a PR but I'm currently very busy, just thought I'd leave an issue ticket.

euclio:
Closes #262.

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -63,11 +63,11 @@ endfunction
 
 " type may be either 'file' or 'project'
 function! neomake#signs#PlaceSign(entry, type) abort
-    if a:entry.type ==# 'W'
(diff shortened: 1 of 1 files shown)


benekastah:
Thanks!

PR #284, opened 2016-01-22
Fix :helptags / documentation by removing duplicate tags.

junkblocker:
Remove duplicate help tags


benekastah:
Thanks for the fixes!

PR #293, opened 2016-01-31
Fix entry types for pylint and pep8 checkers

tdryer:
Add postprocess function for pylint and pep8 makers to correctly set
entry types.


lverweijen:
Since flake8 wraps pep8 and other tools, they could share the same post-process function.

PR #299, opened 2016-02-06
[RDY] Add Windows support and Apache Maven builder

brcolow:
This PR adds Windows support to Neomake and adds a maker for Apache Maven (https://maven.apache.org/). Thanks to @equalsraf, this actually appears to be working. Because Maven is a project based builder, it was necessary to introduce the mvn maker outside of the `ft` directory. It can be invoked thusly: `Neomake! mvn`. Because the exact arguments that one uses to invoke Maven vary so widely between projects, it is recommended to override the default `install` argument by specifying:

````
let g:neomake_mvn_args=['package', '-P', 'unitTests', '-T', '2']
````

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -37,7 +37,11 @@ function! s:JobStart(make_id, exe, ...) abort
         return jobstart(argv, opts)
     else
         if has_args
-            let program = a:exe.' '.join(map(a:1, 'shellescape(v:val)'))
+            if neomake#utils#IsRunningWindows()
+                let program = a:exe.' '.join(map(a:1, 'v:val'))
+            else
+                let program = a:exe.' '.join(map(a:1, 'shellescape(v:val)'))
+            endif
         else
             let program = a:exe
         endif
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: autoload/neomake/utils.vim [hunk 1]
```diff
@@ -129,9 +129,12 @@ function! neomake#utils#MakerFromCommand(shell, command) abort
             \shell_name) >= 0
         let args = ['-c', command]
     else
-        " TODO Windows support (at least)
-        throw "Shell not recognized; can't build command"
+        let shell_name = split(a:shell, '\\')[-1]
+        if (shell_name == 'cmd.exe')
+            let args = ['/C', command]
```
equalsraf: &shellcmdflag instead of '/C', you can probably do it for any shell, not just cmd.exe

autoload/neomake.vim
```diff
@@ -71,7 +76,14 @@ function! neomake#MakeJob(maker) abort
     if append_file
         call add(args, '%:p')
     endif
-    call map(args, 'expand(v:val)')
+    
+    echom string(args)
+    if neomake#utils#IsRunningWindows()
+        " Don't expand '/C' argument to cmd.exe
+        call map(args, 'v:val !=? "/C" ? expand(v:val) : v:val')
```
equalsraf: I'll need to have a deeper look but the whole issue with /C being turned into \C is that Neovim quotes the command when it is called using jobstart([...]) - more in https://github.com/neovim/neovim/pull/3812

equalsraf: My mistake its solely a consequence of `expand()`.

equalsraf: &shellcmdflag

benekastah:
Awesome, thanks for this!

brcolow:
@benekastah Thank you.

PR #300, opened 2016-02-07
Add credo maker for Elixir

mikker:
[Credo](https://github.com/rrrene/credo) is a static analysis tool for elixir code.

Thanks for an amazing plugin :+1:

### DIFF
--- a/autoload/neomake/makers/ft/elixir.vim
+++ b/autoload/neomake/makers/ft/elixir.vim
@@ -11,3 +11,11 @@ function! neomake#makers#ft#elixir#elixir()
             \ '%f:%l: warning: %m'
         \ }
 endfunction
+
+function! neomake#makers#ft#elixir#credo()
+    return {
+      \ 'exe': 'mix',
+      \ 'args': ['credo', 'list', '%:p', '--one-line', '-i', 'readability'],
+      \ 'errorformat': '[%t] %. %f:%l:%c %m'
+      \ }
+endfunction


mikker:
Thanks!

carakan:
+1

emilsoman:
Looks like only warnings (those lines starting with [W]) are being displayed at the moment. It'd be nice if we could show "refactoring opportunities" as warnings as well : 

```
[F] → lib/my_app/file1.ex:7 Pipe chain should start with a raw value.
[F] → lib/my_app/file2.ex:12 Function takes too many parameters (arity is 6, max is 5).
``` 

Consistency related warnings are also of interest : 

```
[C] ↗ lib/my_app/file3.ex:50 There is no whitespace around parentheses/brackets most of the time, but here there is.
```

blueyed:
@emilsoman 
Try changing the `errorformat` to:

    'errorformat': '[%t] %. %f:%l:%c %m,[%t] %. %f:%l %m'

Adding the same without the `%c` for column, which is not in those messages.
If that works, please create a PR for it.

emilsoman:
@blueyed In addition to your suggestion about erroformat, I had to add this in my config to get all kinds of warnings show properly (also show stuff on airline): 

```vim
let g:neomake_elixir_enabled_makers = ['mycredo']
function NeomakeCredoErrorType(entry)
    if a:entry.type ==# 'F'      " Refactoring opportunities
        let type = 'W'
    elseif a:entry.type ==# 'D'  " Software design suggestions
        let type = 'I'
    elseif a:entry.type ==# 'W'  " Warnings
        let type = 'W'
    elseif a:entry.type ==# 'R'  " Readability suggestions
        let type = 'I'
    elseif a:entry.type ==# 'C'  " Convention violation
        let type = 'W'
    else
        let type = 'M'           " Everything else is a message
    endif
    let a:entry.type = type
endfunction
let g:neomake_elixir_mycredo_maker = {
      \ 'exe': 'mix',
      \ 'args': ['credo', 'list', '%:p', '--format=oneline'],
      \ 'errorformat': '[%t] %. %f:%l:%c %m,[%t] %. %f:%l %m',
      \ 'postprocess': function('NeomakeCredoErrorType')
      \ }
```

If this makes sense, I can open a PR, thanks.

bitboxer:
@emilsoman I would love to have this. Can you open a PR about this?

PR #305, opened 2016-02-15
Rust adding options

DarwinSenior:
So, for rust, I found that the inside cargo-check, the cargo check is a much thorough from [this discussion](https://github.com/AtomLinter/linter-rust/issues/29). And [cargo check](https://github.com/rsolomo/cargo-check/tree/e602882a3a88ecfc8206ae13714e19ec2da1d5fd) here.

I am sorry that I am not quite familiar with Git and commited a lot of changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 2016-01-10
+- modified compiler parameter for rust maker
+
 ## 2015-04-06
 - Removed `make_modified` feature and corresponding `g:neomake_make_modified` setting.
 - You can now use the `|` to do a command after neomake: `:Neomake | echo 'something else'`.
--- a/README.md
+++ b/README.md
@@ -41,11 +41,13 @@ Taking a page from the book of syntastic, you can configure "makers" (called
 "checkers" in syntastic) for different filetypes. Here is an example
 configuration:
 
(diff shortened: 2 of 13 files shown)


brcolow:
If you want, you can squash your commit by doing:

`git rebase -i HEAD~9`

This will open up the last 9 commits in the editor that your `~/.gitconfig` is configured to use. If you have not yet configured your editor, you can do:

`git config --global core.editor vim`

which, because of the `--global` flag will write that to your `.gitconfig`.

Once opened, for the latest 8 commits, change the first word `pick` to `squash` OR `fixup`. Squash will make it so the final commit message of the now sole squash commit will be made up of all the individual commit messages of the squashed commits, which you can now edit. Fixup will make it so the commit messages of each individual "squashed" commit will be ignored, and you can still change the now-squashed sole commit by changing `pick` to `reword` for the first commit.

Once that is done, save your work in the editor (e.g. `:wq` for vim) and git will rebase commits. Then type:

`git push --force`

and it will be cleaned up and squashed!

DarwinSenior:
I think there are some other changes that is going on in the middle, and it is better to use another branch. I will make a better PR with another branch. Thank you for your instruction!

brcolow:
@DarwinSenior You're welcome :)

PR #308, opened 2016-02-16
added neovim flag for vim's vint maker

sbdchd:


autoload/neomake/makers/ft/vim.vim
```diff
@@ -6,7 +6,7 @@ endfunction
 
 function! neomake#makers#ft#vim#vint() abort
     return {
-        \ 'args': ['--style-problem'],
+        \ 'args': ['--style-problem', '--enable-neovim'],
```
benekastah: We should only do this if `has('nvim') == 1`, right

benekastah:
Thanks for this!

PR #316, opened 2016-02-25
Show maker command in statusline

nerdrew:
This takes advantage of: https://github.com/neovim/neovim/pull/4337

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -348,6 +348,9 @@ function! s:Make(options, ...) abort
             return []
         endif
         let enabled_makers = ['makeprg']
+        let title = 'makeprg'
+    else
+        let title = join(map(copy(enabled_makers), 'v:val.name'), ', ')
     endif
 
     if a:0
@@ -374,9 +377,17 @@ function! s:Make(options, ...) abort
     try
         let &errorformat = '%-G'
         if file_mode
-            lgetexpr ''
+            try
+                call setloclist(0, [], '', title)
+            catch /E118/
+                call setloclist(0, [], '')
+            endtry
         else
-            cgetexpr ''
+            try
+                call setqflist([], '', title)
+            catch /E118/
+                call setqflist([], '')
+            endtry
         endif
     finally
         let &errorformat = l:efm
@@ -527,9 +538,17 @@ function! s:AddExprCallback(jobinfo) abort
 
     if list_modified
         if file_mode
-            call setloclist(maker.winnr, list, 'r')
+            try
+                call setloclist(maker.winnr, list, '')
+            catch /E118/
+                call setloclist(maker.winnr, list, '')
+            endtry
         else
-            call setqflist(list, 'r')
+            try
+                call setqflist(list, '', name)
+            catch /E118/
+                call setqflist(list, '')
+            endtry
         endif
     endif
 endfunction


autoload/neomake.vim
```diff
@@ -368,9 +371,9 @@ function! s:AddExprCallback(maker) abort
 
     if list_modified
         if file_mode
-            call setloclist(a:maker.winnr, list, 'r')
+            call setloclist(a:maker.winnr, list, 'r', name)
```
nerdrew: I'm not sure about this change. I'm not sure how to test it.

mhinz: @nerdrew Hey, I'm the one who added `{title}` to these functions.

> I'm not sure how to test it.

Testing what exactly?

nerdrew: Testing how this is exercised in neomake. There are two places where I changed the code. I'm not sure when `list_modified` is true in this case. The first change gets executed in my `Ag` command:

```
function! s:Ag(file_mode, args)
  let cmd = "ag --vimgrep --smart-case ".substitute(a:args, '\\', '\\\\', 'g')
  let custom_maker = neomake#utils#MakerFromCommand('bash', cmd)
  let custom_maker.name = cmd
  let custom_maker.remove_invalid_entries = 0
  let custom_maker.place_signs = 0
  let custom_maker.errorformat = "%f:%l:%c:%m"
  let enabled_makers =  [custom_maker]
  call neomake#Make(a:file_mode, enabled_makers) | echo "running: " . cmd
endfunction
command! -bang -nargs=* -complete=file Ag call s:Ag(<bang>0, <q-args>)
```

mhinz: Oh, I see. Unfortunately I'm of no help here, since I haven't dug into neomake so far. Hope you figure it out nevertheless.

autoload/neomake.vim
```diff
@@ -248,10 +248,12 @@ function! s:Make(options) abort
         endif
     endif
 
+    let title = join(map(copy(enabled_makers), 'v:val.name'), ', ')
```
blueyed: This should be only the markers that reported something probably?!

autoload/neomake.vim
```diff
@@ -248,10 +248,12 @@ function! s:Make(options) abort
...
+
     if file_mode
-        lgetexpr ''
+        call setloclist([], 'r', title)
     else
-        cgetexpr ''
+        call setqflist([], 'r', title)
```
blueyed: Please handle `E118` for backwards compatibility: https://github.com/neovim/neovim/pull/4337#issuecomment-188891314

autoload/neomake.vim
```diff
@@ -368,9 +371,9 @@ function! s:AddExprCallback(maker) abort
 
     if list_modified
         if file_mode
-            call setloclist(a:maker.winnr, list, 'r')
+            call setloclist(a:maker.winnr, list, 'r', name)
         else
-            call setqflist(list, 'r')
+            call setqflist(list, 'r', name)
```
blueyed: Please handle `E118` for backwards compatibility: https://github.com/neovim/neovim/pull/4337#issuecomment-188891314 (maybe wrap/refactor it in a function, because there are 2/4 places).

nerdrew:
Hmmm. neovim is seg faulting when I use this. cc @mhinz

mhinz:
@nerdrew When _exactly_ is this happening? Given that `qf_jump()` crashed, I guess you were about to switch to a quickfix entry?

nerdrew:
I think I was doing it wrong :)

I was: `setqflist([], 'r', 'my title')`, adding items with `caddexpr blah`, then `setqflist([], 'r', 'my title')` again (this should have been `setqflist([], '', 'my title')`), and add more items with `caddexpr more_blah`. Trying to jump to any of the items in the last list causes the seg fault.

blueyed:
Related: https://github.com/benekastah/neomake/pull/325

mhinz:
@nerdrew It would be nice if someone could [create an issue](https://github.com/neovim/neovim/issues) for it, so we don't lose track. (In the best case with minimal steps to reproduce it _without_ neomake.)

autoload/neomake.vim
benekastah: @nerdrew I think this is a pretty low-risk change, so if you can't get this condition to occur don't worry about it too much. However, in all the places you modified the call to `setloclist` or `setqflist` with what seems to be a neomake-only argument, I'd appreciate if you could make it ordinary-vim friendly. Unfortunately this means doing something like:

```
if has('nvim')
    " Neovim version
else
    " Vim version
endif
```

You're welcome to add functions wrapping this boilerplate in [autoload/neomake/utils.vim](https://github.com/benekastah/neomake/blob/master/autoload/neomake/utils.vim) if you think that's nicer.

autoload/neomake.vim
benekastah: Looks pretty reasonable to me. However, no need to copy `enabled_makers` here since we aren't mutating it.

autoload/neomake.vim
benekastah: @nerdrew you shouldn't need to worry about handling this error if you handle both cases as I suggested earlier. But +1 for thinking about backwards compatibility @blueyed!

blueyed:
@nerdrew 
Can you please address the remaining issues and update/rebase this PR, please?

autoload/neomake.vim
nerdrew: It errors if I remove the copy.

nerdrew:
Ok, updated to add `if has('nvim')`. Please take a look.

blueyed:
I think it should use `try` and handle the mentioned error when the patch is not available, like I have commented before.
Vim will also get the patch eventually and/or you might have it applied locally.

nerdrew:
I updated it with `try`, let me know if it looks right.

blueyed:
@nerdrew 
Thanks, that's what I've meant.  Will give it a try.

autoload/neomake.vim
blueyed: @nerdrew 
Seen this error here:
```
Error detected while processing function neomake#Make[13]..<SNR>194_Make:
line   17:
E121: Undefined variable: name
```
Just by running `Neomake` on `plugin/neomake.vim` on this PR.

nerdrew: Hmmm. I don't get an error. I wonder if it is something in my global config...

nerdrew: Seems like I do need to use `endtry` instead of just `end` though

nerdrew: This is the command that errors? `vi plugin/neomake.vim "+:Neomake"`

nerdrew: Hmmm. I see the error. `enabled_makers` is just an array. Doh!

blueyed:
I'm adopting this.

blueyed:
Closing in favor of https://github.com/neomake/neomake/pull/924.

nerdrew:
Hmmm. Just getting around to looking at this. I still use my change to get my maker name in the quickfix window title. When I run my custom maker (which runs ripgrep), I get the following title:

```
[Quickfix List] :            cgetexpr ''
```

Here's the maker command:

```
function! s:Rg(file_mode, args)
  let cmd = "rg --vimgrep ".a:args
  let custom_maker = neomake#utils#MakerFromCommand(cmd)
  let custom_maker.name = cmd
  let custom_maker.remove_invalid_entries = 0
  let custom_maker.errorformat = "%f:%l:%c:%m"
  let enabled_makers =  [custom_maker]
  call neomake#Make(a:file_mode, enabled_makers) | echo "running: " . cmd
endfunction
command! -bang -nargs=* -complete=file G call s:Rg(<bang>0, <q-args>)
```

Here's the title I get with my branch:

```
[Quickfix List] :rg --vimgrep boom
```

Is there another way to get this title without using my change (https://github.com/nerdrew/neomake/commit/72667abdc777c23ca9a49ddc54b7defc208c0055)?

blueyed:
https://github.com/neomake/neomake/pull/924 is not merged yet, so you can either try it, of keep using your method.

nerdrew:
DOH! Ya, I missed that. I'll be patient then :)

blueyed:
It's worth checking out locally - feedback would certainly be provided over there.
See https://help.github.com/articles/checking-out-pull-requests-locally/.

PR #321, opened 2016-02-29
Added a graphviz maker

teto:
It's not ready yet but I had some questions: how to redirect the output of the maker ? for instance here, the 'dot' binary generates a png and it appears in neomake log. I would like to redirect the output, hence the '> %:p.png' but that does not seem to work.
Also I wonder if errorformats should not be splitted as it seems to be in other makers ?

### DIFF
--- a/autoload/neomake/makers/ft/dot.vim
+++ b/autoload/neomake/makers/ft/dot.vim
@@ -1,0 +1,16 @@
+" vim: ts=4 sw=4 et
+" inspired by wmgraphviz.vim
+function! neomake#makers#ft#dot#EnabledMakers()
+    if neomake#utils#Exists('dot')
+        return ['dot']
+    end
+endfunction
+
+function! neomake#makers#ft#dot#dot()
+    return {
+        \ 'args': ['-Tpng', '%:p', '> %:p.png'],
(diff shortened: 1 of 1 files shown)


benekastah:
You won't be able to do redirecting here for two reasons:

1. Each entry in the arg array will be interpreted as a quoted string, and so no shell syntax will work
2. We can't make assumptions about what shell people are using, and so it could just be a syntax error for some people.

Your best options are:

1. See if `dot` has a flag that tells it where to write the file (instead of to stdout).
2. You can use the `postprocess` maker option. See `:help neomake-makers-processing`. Vim has a cross-platform function for writing to files, see `:help writefile()`.

benekastah:
As for the errorformat, don't worry about splitting it. The benefit for that mainly comes when the errorformat is very long.

blueyed:
#373 might help, where `args` can be a function/callback.

blueyed:
There is `-o`, and you could use `neomake#utils#DevNull()` to not create a file after all.

blueyed:
@teto 
Will you be working on this again?

teto:
I don't think I can finish this anytime soon so feel free to close it up.

PR #324, opened 2016-03-01
Fix python maker for Vim: use \r instead of \n

blueyed:
When using Vim instead of Neovim, the python maker failed like this:

> Neomake: python stdout:   File "<string>", line 2
> Neomake: python stdout:     from sys import argv, exit\
> Neomake: python stdout:        ^
> Neomake: python stdout: SyntaxError: invalid syntax
> Neomake: python stdout:
> Neomake: python stdout done.
> Neomake: python processing 5 lines of output
> Neomake: File-level errors cleaned in buffer 1

It seems like we have to use "\r" for newlines, instead of "\n".

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -128,13 +128,13 @@ endfunction
 function! neomake#makers#ft#python#python()
     return {
         \ 'args': [ '-c',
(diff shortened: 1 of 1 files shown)


benekastah:
Huh, good catch. I thought this had been working, but I tested it out and you're totally right.

PR #325, opened 2016-03-02
[RFC] Annotate list entries with maker name

blueyed:
This is useful (only) with multiple makers, which is not handled
currently.
For this, it should get done in AddExprCallback probably (once `index`
is > 1), but this then needs an extra list to keep track of maker names
per line (it does not seem to be possible to store extra data with the
location/quickfix list itself (`entry.maker_name`)).

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -403,6 +403,12 @@ function! s:ProcessJobOutput(maker, lines) abort
         let olderrformat = &errorformat
         let &errorformat = a:maker.errorformat
 
+        let lines = a:lines
+        if !get(g:, 'neomake_dont_annotate_with_maker_name')
+            let maker_name = has_key(a:maker, 'name') ? a:maker.name : 'makeprg'
+            let lines = map(copy(lines), 'v:val . " [".a:maker.name."]"')
+        endif
+
         if get(a:maker, 'file_mode')
             " Go to window if it's not current.
             if winnr() != a:maker.winnr
@@ -411,7 +417,7 @@ function! s:ProcessJobOutput(maker, lines) abort
                 exec a:maker.winnr.'wincmd w'
             endif
 
-            laddexpr a:lines
+            laddexpr lines
 
             " Restore window.
             if exists('l:prev_window')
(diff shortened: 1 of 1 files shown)


blueyed:
TODO: This will also add it to the errors displayed on the current line, where it's prefixed already (so duplicated).

blueyed:
That does not make sense for the `pylama` checker, which will add it itself (the name of the sub-tool).

benekastah:
The tricky thing here is we could potentially break the errorformat for some makers. I think we should use `setloclist` and `setqflist` instead. One idea on how we could do this (untested code ahead):

```
let idx = len(getqflist())
caddexpr lines
let list = getqflist()
while idx < len(list)
    if list[idx].text !~ maker_name
        list[idx].text .= ' [' . maker_name . ']'
    endif
endwhile
setqflist(list, 'a')
```

I'm a little unsure how this will impact performance for day-to-day usage.

blueyed:
Should this be done in `s:ProcessJobOutput`?

tweekmonster:
I don't think messing with the quickfix/location list is the way to go since its values has meaning to vim itself.  What Neomake can do is create correlating lists.  For quickfix, only one list needs to be maintained.  For location lists, a per-buffer list could be maintained.

I've been messing around with a plugin that alters the quickfix buffer to be less dumb (for lack of a better word).  My personal goal is to remove the filename from the location list since it's the least informative text that takes up the most screen space, especially for files within directories.

Doing this is pretty easy since the quickfix buffer can be altered to display text any way you want as long as the line numbers match with the quickfix list index:



The filename is removed and the `line:col` is condensed.  For Neomake, the message type could be gutter sign instead of actual text.

<details>
<summary>Script</summary>

```vim
function! s:reformat() abort
  echomsg 'reformat'
  let ul = &l:undolevels
  setlocal modifiable nonumber undolevels=-1
  silent % delete _

  let lnum_width = 0
  let col_width = 0

  let loclist = getloclist(0)
  for item in loclist
    let lnum_width = max([len(item.lnum), lnum_width])
    let col_width = max([len(item.col), col_width])
  endfor

  let lines = []
  for item in loclist
    let type = toupper(item.type)
    if empty(type)
      let type = '?'
    endif
    call add(lines, printf(' %*s:%*s [%s] %s', lnum_width, item.lnum, col_width, item.col, type, item.text))
  endfor

  call setline(1, lines)
  let &l:undolevels = ul
  setlocal nomodifiable nomodified
endfunction
```
</details>

---

Comparison of what it typically looks like:

blueyed:
Very nice!

blueyed:
Closing it in favor of #924.

PR #328, opened 2016-03-02
Handle window switching around "lwindow"

blueyed:
This will open the correct location list, in case a buffer is opened in
multiple windows.

TEST CASE:
 - split a buffer into two windows
 - run ":Neomake" in the first window
 - switch to the other window

Results should get displayed with the first window.


benekastah:
Thanks for this!

PR #330, opened 2016-03-02
Keep order of enabled makers

blueyed:
Using `keys()` on the `union` dict does not retain the order, which is
important for the python makers, where `python` should be run first, and
`pylint` last (in serialized mode).


benekastah:
Thanks!

PR #331, opened 2016-03-03
Travis CI build

siddharthist:
The build currently fails with a few errors. Logs [here](https://travis-ci.org/siddharthist/neomake). It uses two different vim linters, `vim-vimlint` and `vint`. It caches build dependencies, for a build time of ~45 seconds (:open_mouth:).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: python
+
+cache:
+  pip: true
+  directories:
+    - /tmp/vimlint
+    - /tmp/vimlparser
+    - /tmp/vint
+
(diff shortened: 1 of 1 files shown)


siddharthist:
What do you think @benekastah? I think the issues are all valid, but I don't know enough about Vimscript to fix them. Maybe we could merge this and follow up quickly with a PR to fix the issues it finds now?

benekastah:
Thanks! Just merged it. For now I'll continue merging pull requests that fail travis until we get the issues fixed (I won't have time to help with that tonight, perhaps tomorrow).

PR #345, opened 2016-03-12
Add "semistandard" JS maker and mark entries as warnings

szymonkaliski:


autoload/neomake/makers/ft/javascript.vim
```diff
@@ -36,10 +36,23 @@ endfunction
...
+        \ 'errorformat': '  %f:%l:%c: %m',
+        \ 'postprocess': function('neomake#makers#ft#javascript#StandardEntryProcess')
+        \ }
+endfunction
+
+function! neomake#makers#ft#javascript#StandardEntryProcess(entry)
+    " (Semi)Standard entries are warnings
+    let a:entry.type = 'W'
+endfunction
+
```
blueyed: I don't think you need postprocessing here, `'  %W%f:%l:%c: %m'` should do.

szymonkaliski: without this semistandard, and standard makers didn't mark entries as warnings for me

blueyed: But you are not using `%W` in the errorformat..

szymonkaliski: ahh, I don't really have time to mess with it in next few days, so if you don't have the time to fix it we could close this and I'll send new PR when I can?

blueyed: We can just leave this PR open, and you can amend the commit.
My point is that you probably only have to change the errorformat, and no post-processing is required.

jcartledge:
Hi,

`'%W  %f:%l:%c: %m'` works - I'm running this patched at the moment.

The same change should also be applied to the standard maker.

PR #347, opened 2016-03-13
Add Stylus linter (stylint)

herrbischoff:
Maker for Stylus linting via `stylint`.


herrbischoff:
@benekastah: Any interest in this? I'm using it in production for a couple of weeks now without any issues.

blueyed:
Thanks, @herrbischoff.

Could you add it to the README also, please? (after rebasing)
I think it makes sense to include an URL / link to the linter's homepage there then, too.

PR #349, opened 2016-03-15
minor: reformat plugin/neomake.vim

blueyed:


blueyed:
Rebased on master, moving to "next".

PR #351, opened 2016-03-20
makers: go.vim - change to `cwd` of file in buffer for govet

gnanderson:


blueyed:
Why is this needed?

blueyed:
Seems to be required for `go`, too: https://github.com/neomake/neomake/issues/358#issuecomment-202449049.

blueyed:
This was included in #410.
Thanks!

PR #352, opened 2016-03-20
Fix arg with call to jobstop in neomake#CancelJob

blueyed:


dojoteef:
This code does not work with my version of neovim. Looking at the [documentation](https://neovim.io/doc/user/eval.html) for jobstart() it states that it returns the job id that should be passed to jobstop() in order to stop the job.

Here you've changed the call to jobstop() to pass in the jobinfo dict rather than the job id. That's the reason I [added this](https://github.com/dojoteef/dotfiles/blob/ecafedff904df58222712ae06eb5a232c450781d/link/.vimrc#L901) in my lint as you type feature for neomake. I was going to make a pull request, but in searching the git log I noticed this change so I figured I would ask why this change was necessary.

blueyed:
@dojoteef 
Thanks for bringing that up.
I've probably broke this because of some misunderstanding.  Fixed in https://github.com/neomake/neomake/pull/548.

I wonder if `neomake#CancelJob` should catch the `E900` like you are doing in your code?! (and return 0 then)

dojoteef:
Probably. I actually got that from [neomake.vim](https://github.com/neomake/neomake/blob/1a051f05d16e33d405ef9ae1132403d6220641b3/autoload/neomake.vim#L431). It does not appear to be documented in the neovim docs, but here is a [neovim issue](https://github.com/neovim/neovim/issues/1598) that mentions it.

PR #353, opened 2016-03-20
Add initial primitive support for Idris.

majjoha:
This pull request adds initial primitive support for checking Idris source files for errors. You need to open the location list to see the full message from the compiler, but at least it highlights the correct lines when issues occur.

### DIFF
--- a/autoload/neomake/makers/ft/idris.vim
+++ b/autoload/neomake/makers/ft/idris.vim
@@ -1,0 +1,12 @@
+function! neomake#makers#ft#idris#EnabledMakers() abort
+    return ['idris']
+endfunction
+
+function! neomake#makers#ft#idris#idris() abort
+    return {
+        \ 'exe': 'idris',
+        \ 'args': ['--check', '--warn', '--total', '--warnpartial', '--warnreach'],
+        \ 'errorformat':
+            \ '%f:%l:%c%.%#:,%m' ,
+        \ }
+endfunction


majjoha:
If you have a chance to review, and/or merge this, I'd greatly appreciate it, @blueyed.

autoload/neomake/makers/ft/idris.vim
```diff
@@ -0,0 +1,13 @@
...
+    return ['idris']
+endfunction
+
+function! neomake#makers#ft#idris#idris()
+    return {
+        \ 'exe': 'idris',
+        \ 'args': ["--check", "--warn"],
+        \ 'errorformat':
+        \ '%f:%l:%c:,' .
+        \ '%m'
```
blueyed: Are you sure about the errorformat?!
It boils down to `'errorformat': '%f:%l:%c:,%m'` - so two entries whereas the first has no message, and the 2nd one is only a message?!

blueyed:
This is my patch against this PR, please include it when updating it:

```diff
diff --git i/autoload/neomake/makers/ft/idris.vim w/autoload/neomake/makers/ft/idris.vim
index a4a5a8f..ebd6b1a 100644
--- i/autoload/neomake/makers/ft/idris.vim
+++ w/autoload/neomake/makers/ft/idris.vim
@@ -1,13 +1,10 @@
-function! neomake#makers#ft#idris#EnabledMakers()
+function! neomake#makers#ft#idris#EnabledMakers() abort
     return ['idris']
 endfunction
 
-function! neomake#makers#ft#idris#idris()
+function! neomake#makers#ft#idris#idris() abort
     return {
-        \ 'exe': 'idris',
-        \ 'args': ["--check", "--warn"],
-        \ 'errorformat':
-        \ '%f:%l:%c:,' .
-        \ '%m'
-        \ }
+                \ 'args': ['--check', '--warn'],
+                \ 'errorformat': '%f:%l:%c: %m'
+                \ }
 endfunction
```

blueyed:
@majjoha 
ping

majjoha:
Thanks for your feedback, @blueyed.

I've been working on updating this pull request, but I've run into an issue where `idris --check` completes with exit code 1 which Neomake sees, but it does not mark the error in the editor.

screen shot 2016-12-10 at 11 14 12

I've submitted an updated pull request with your changes.

blueyed:
So you are saying that it does not work with the error (and multiline errors in general) still?
Looks like the `errorformat` needs to handle long lines.  See https://github.com/neomake/neomake/blob/1a9c6d7c6c63583064e69b7fbf85d987e800fe16/doc/neomake.txt#L507 for hints to develop it from a given output.

I'd rather see code snippets than screenshots btw, then I might have tried it already, but I won't type it from a screenshot, sorry.

majjoha:
Correct, but I've made a bit of progress, and updated the pull request, accordingly. The maker for Idris has been changed to the following:

```vim
function! neomake#makers#ft#idris#EnabledMakers() abort
    return ['idris']
endfunction

function! neomake#makers#ft#idris#idris() abort
    return {
        \ 'exe': 'idris',
        \ 'args': ['--check', '--warn', '--total', '--warnpartial', '--warnreach'],
        \ 'errorformat':
            \ '%f:%l:%c%.%#:,%m' ,
        \ }
endfunction
```

It now understands, and highlights both errors with one column, and a range of columns. Although, it properly highlights the line with the error, the command bar only says `idris:` at the moment. I am not sure if this is fixable due to the multiline error messages.

I am testing against this Idris program:

```idris
module Main

total
xor : Bool -> Bool -> Bool
-- xor False y = 5
xor 5 = False
```

From the terminal, I run this to verify the error format:

```
~/Dropbox/Kode/Idris % idris --check --total --warnpartial --warnreach --warn xor.idr
xor.idr:6:1-4:When checking left hand side of xor:
When checking an application of Main.xor:
        Type mismatch between
                Integer (Type of 5)
        and
                Bool (Expected type)
```

blueyed:
Looked at this a bit, and came up with the following errorformat: `'%E%f:%l:%c:%.%#,%-C  %#%m,%-C%.%#'`.

Check out `:let &efm='%E%f:%l:%c:%.%#,%-C  %#%m,%-C%.%#' | cgetbuffer | echo getqflist()` for testing it, after loading the output into a buffer.

I wonder about the part after the column (`4` in `:6:1-4`).  Is that supposed to be the end position?

I think I will look into this later a bit more - so do not spend too much time yourself digging into it.. ;)

blueyed:
Check out https://github.com/neomake/neomake/pull/890.

blueyed:
Closing it in favor of #890.

@majjoha 
Please answer to my comment above there, thanks.

PR #357, opened 2016-03-26
Include file name during `go build` and `go vet`

anupcshan:
Both `go` and `govet` rules fail as they are right now.

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -10,7 +10,7 @@ function! neomake#makers#ft#go#go()
             \ 'build',
             \ '-o', neomake#utils#DevNull()
         \ ],
-        \ 'append_file': 0,
+        \ 'append_file': 1,
         \ 'errorformat':
             \ '%W%f:%l: warning: %m,' .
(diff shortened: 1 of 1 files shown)


hansrodtang:
How do they fail? `go build` and `go vet` without a file name are perfectly valid commands and run on the current package.

Something like this fails on a per-file basis for `go build`:

test/file.go:
```go
package main

const T = 5
```

test/main.go:
```go
package main

func main(){
  return T
}
```

Error: `./main.go:76: undefined: T`

anupcshan:
With the example you've posted above, opening go files from another directory, like `nvim test/main.go` (instead of `nvim main.go`), doesn't set CWD correctly to `go build` or `go vet`.

Ideally, we'd provide directory associated with current buffer. This pull request causes way more problems than it solves. I'll close it and file a new issue.

hansrodtang:
```vim
augroup auto_change_directory
	autocmd!
	au BufEnter * silent! lcd %:p:h
augroup END
```
Putting this in your `init.vim` is a quick workaround, but yeah, setting the `cwd` property to the correct package for the maker is probably the best choice.

Issue #138, opened 2015-05-27
Custom preprocessing function


kirelagin:
Let’s be honest, `errorformat` is horrible.

It would be great if it was easy to create a maker that passes obtained output to a custom function (probably even written in Python) that transforms the output into something parsable with `errorformat`.

That’s somewhat related to #111 as this is something similar to `mapexpr` (to be honest, I fail to see how `mapexpr` can be useful in presence of multiline error messages).

benekastah:
I don't like errorformats much myself, especially for multiline error messages. But I don't know if I love writing custom code to parse errorformats. I don't want to maintain code like that for one. As terrible as errorformats are, they are well documented and usually don't required my personal attention to fix. Especially since syntastic has tons of great errorformats that we can use if ours break.

If some maker (like `cargo` for example) doesn't produce output that is easy to parse with an errorformat, you can submit a pull request to `cargo` to add an option to output something easy to parse (like CSV) instead. Alternately, you could write a wrapper program that takes `cargo`'s output, parses it, and spits it out in some other format. Then you could add that as its own maker in this project.

kirelagin:
Hm, I am just asking for a hook to let maker specifications authors add a bit of custom preprocessing. This won’t require you to write or maintain any additional code (well, except for the code in neomake that will invoke the preprocessor if it is configured for the maker).

The wrapper idea is a nice one, and it’s probably the way to go. I’ll give it a shot, thanks!

kirelagin:
Well, since the wrapper thing is actually a fully functional replacement for what I was asking for, I guess, this can be closed.

PR #359, opened 2016-03-29
Added preprocess maker option + purescript syntax checker

FrigoEU:
Hey,

I'm maintaining a PureScript syntax checker for Syntastic for some time now, and am very interested to have a maker for Neomake as well. I ran into a few issues along the way though, and did some controversial stuff so I wouldn't be surprised if you don't just merge this PR off the bat. I'll try to explain my decisions + reasons why:
1. The PureScript compiler (called through the standard build tool, "pulp") has two output modes: One text mode for human users, one JSON mode for editors. In Syntastic you have a maker option called "preprocess", in which I do the JSON parsing. I tried to do this with mapexpr but this is just impossible. The JSON came through in small parts, impossible to parse. I even implemented my own buffering system as an experiment but came to the conclusion that returning multiple errors from mapexpr is impossible.
2. I read in issue #138 that you don't want to add this kind of preprocessing function because you don't want to support the code being written in this function. I totally understand! But to be honest, I would never expect you to support my purescript maker. Secondly, Syntastic has this concept of external plugins (which I use for the Purescript Syntastic checker) and that would completely take away the responsability from you, so maybe that's an idea? Thirdly, writing this kind of wrapper around the PureScript compiler/Pulp is really just redundant, since Pulp is already a wrapper around the compiler. 
3. Something different now: The PureScript compiler also emits suggestions for certain warnings/errors. I use these suggestions in my psc-ide-vim plugin. This is optional of course, but there is some integration code for that in this maker. I hope that's not a problem? Again, I'd feel much more comfortable if I could maintain this maker in my own repo so you'd be shielded from whatever crazy stuff I'd like to do with the compiler output...

Simon

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -399,7 +399,14 @@ endfunction
 function! s:ProcessJobOutput(maker, lines) abort
     call neomake#utils#DebugMessage(get(a:maker, 'name', 'makeprg').' processing '.
                                     \ len(a:lines).' lines of output')
-    if len(a:lines) > 0
+
+    let lines = a:lines
+    if has_key(a:maker, 'preprocess')
+        let Func = a:maker.preprocess
+        let lines = Func(copy(a:lines))
+    endif
+
+    if len(lines) > 0
         let olderrformat = &errorformat
         let &errorformat = a:maker.errorformat
 
@@ -411,7 +418,7 @@ function! s:ProcessJobOutput(maker, lines) abort
                 exec a:maker.winnr.'wincmd w'
             endif
 
-            laddexpr a:lines
+            laddexpr lines
 
             " Restore window.
             if exists('l:prev_window')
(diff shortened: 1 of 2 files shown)


blueyed:
> Syntastic has this concept of external plugins (which I use for the Purescript Syntastic checker)

Which works by using `execute 'runtime! syntax_checkers/' . a:filetype . '/*.vim'`, right?
I like that idea.

But then the `preprocess` would still be required/useful, right?

blueyed:
@FrigoEU 
ping

ditadi:
I would very much like to see purescript support in neomake.  Thank you guys for any progress!

blueyed:
@ditadi 
Looks like a good opportunity to help out then.. :)

FrigoEU:
Hey, this PR can be closed. There have been massive performance improvements in the PureScript compiler so checking goes super fast. Syntastic is plenty fast now very every use case.

blueyed:
A purescript maker has been added: https://github.com/neomake/neomake/pull/1194.

PR #360, opened 2016-03-30
Don't do full flowtype check

marcelbeumer:
Removing the (full) check argument for flowtype. It makes the maker very slow and works fine without. I think full checks on project code should be part of build/deploy scripts.


blueyed:
@antoyo 
Do you agree to change this default (since you've added this maker)?

antoyo:
No problem. I don't use flow anymore :) .

blueyed:
Thanks!

Issue #212, opened 2015-11-08
feature request: highlight error line/column, not just gutter


jaredly:
maybe there's a setting for this already? It would be super helpful.

Thanks for an awesome plugin!

SimpleX91:
I'm interested too.

erf:
+1 i find the gutter indentation annoying. would rather highlight line with a color like [wookiehangover/jshint.vim](https://github.com/wookiehangover/jshint.vim)

camagu:
+1

ahmedelgabri:
Yeah this will be great, this is the only thing I miss from [Syntastic](https://github.com/scrooloose/syntastic/blob/b10c59b21492776cb4bd5e9802cae3a766c6b1ab/doc/syntastic.txt#L182-L186)

sebenik:
Yep, this would be really great.

yarray:
+1

mikeys:
Any news about the attached PR? this is really useful.

blueyed:
@mikeys 
Have you tested the PR?

Or anybody else?
Please provide feedback there, and then we can get it merged.

mhartington:
Hmm, curious, is the purpose of the PR and this issue to add a feature like this? 

![](https://raw.githubusercontent.com/jordwalke/VimBox/master/dotVim/images/VimJSXHint.png)

Where errors could be unlined or at least be part of highlight group for themes to target?

jaredly:
that would be great!

mikekreeki:
That would actually be the best thing ever to improve my development workflow 👍

blueyed:
Please see https://github.com/neomake/neomake/pull/361: you could test it already and help out there, although it seems to be waiting for an update by @paretje currently.
And please refrain from +1 comments.

blueyed:
#361 is merged since a while.. 🎆

PR #361, opened 2016-03-30
Highlighting of columns

paretje:
This PR implements #212. If the entry has a column assigned, it is highlighted, if no column is specified, nothing is done, as this would just duplicate information already visualized by the signs.

Writing this PR, I see however a comment in the issue to visualise lines. Maybe I should make it an option to highlight the complete line when no column is known. Then you could use this instead of the signs. It's easy to do, but I'll probably wait for a comment before I do.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -455,6 +455,8 @@ function! s:Make(options, ...) abort
     endif
     call neomake#signs#DefineSigns()
 
+    call neomake#highlights#DefineHighlights()
+
     call neomake#utils#DebugMessage(printf('Running makers: %s',
                 \ string(enabled_makers)), {'make_id': make_id})
 
@@ -542,6 +544,7 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
     let maker = a:jobinfo.maker
     let file_mode = get(maker, 'file_mode')
     let place_signs = get(g:, 'neomake_place_signs', 1)
+    let highlight_columns = get(g:, 'neomake_highlight_columns', 1)
     let list = file_mode ? getloclist(0) : getqflist()
     let list_modified = 0
     let counts_changed = 0
@@ -618,6 +621,9 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
             else
                 call neomake#signs#RegisterSign(entry, maker_type)
             endif
+        endif
+        if highlight_columns
+            call neomake#highlights#AddHighlight(entry, maker_type)
         endif
     endwhile
 
@@ -709,6 +715,7 @@ function! neomake#ProcessCurrentWindow() abort
         endfor
         call neomake#signs#PlaceVisibleSigns()
     endif
+    call neomake#highlights#ShowHighlights()
 endfunction
 
 " Get tabnr and winnr for a given job ID.
@@ -883,6 +890,7 @@ function! neomake#CleanOldProjectSignsAndErrors() abort
     if s:need_errors_cleaning['project']
         for buf in keys(s:current_errors.project)
             unlet s:current_errors['project'][buf]
+            call neomake#highlights#ResetProject(buf)
         endfor
         let s:need_errors_cleaning['project'] = 0
         call neomake#utils#DebugMessage('All project-level errors cleaned.')
@@ -897,6 +905,7 @@ function! neomake#CleanOldFileSignsAndErrors(...) abort
             unlet s:current_errors['file'][bufnr]
         endif
         unlet s:need_errors_cleaning['file'][bufnr]
+        call neomake#highlights#ResetFile(bufnr)
         call neomake#utils#DebugMessage('File-level errors cleaned in buffer '.bufnr)
     endif
     call neomake#signs#CleanOldSigns(bufnr, 'file')
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -891,7 +891,7 @@ function! neomake#CleanOldProjectSignsAndErrors() abort
     if s:need_errors_cleaning['project']
         for buf in keys(s:current_errors.project)
             unlet s:current_errors['project'][buf]
-            call neomake#highlights#ResetProject(buf)
+            call neomake#highlights#ResetProject(buf + 0)
         endfor
         let s:need_errors_cleaning['project'] = 0
         call neomake#utils#DebugMessage('All project-level errors cleaned.')
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -545,6 +545,7 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
     let file_mode = get(maker, 'file_mode')
     let place_signs = get(g:, 'neomake_place_signs', 1)
     let highlight_columns = get(g:, 'neomake_highlight_columns', 1)
+    let highlight_lines = get(g:, 'neomake_highlight_lines', 1)
     let list = file_mode ? getloclist(0) : getqflist()
     let list_modified = 0
     let counts_changed = 0
@@ -622,7 +623,7 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
                 call neomake#signs#RegisterSign(entry, maker_type)
             endif
         endif
-        if highlight_columns
+        if highlight_columns || highlight_lines
             call neomake#highlights#AddHighlight(entry, maker_type)
         endif
     endwhile
--- a/autoload/neomake/highlights.vim
+++ b/autoload/neomake/highlights.vim
@@ -33,7 +33,11 @@ function! neomake#highlights#AddHighlight(entry, type) abort
     endif
     if a:entry.col
         let l:hi = get(s:highlight_types, toupper(a:entry.type), 'NeomakeError')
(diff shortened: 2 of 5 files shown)


blueyed:
@paretje 
Thanks for this!
Can you please rebase this on master (resolving conflics) and then also squash it?

paretje:
@blueyed: done

autoload/neomake/highlights.vim
```diff
@@ -0,0 +1,66 @@
...
+        let l:hi = get(s:highlight_types, toupper(a:entry.type), 'NeomakeError')
+        call add(s:highlights[a:type][a:entry.bufnr][l:hi], [a:entry.lnum, a:entry.col, 1])
+    endif
+endfunction
+
+function! neomake#highlights#ShowHighlights(buf) abort
+    call s:ResetHighlights()
+    for l:type in ['file', 'project']
+        for l:hi in keys(get(s:highlights[l:type], a:buf, {}))
+            call matchaddpos(l:hi, s:highlights[l:type][a:buf][l:hi])
```
blueyed: This seems to return the id, that can be used with `matchdelete` - so there's no need to loop over all matches in `s:ResetHighlights`.

blueyed:
Some feedback:
1. errors are highlighted in the location list window for me, too!  (if there's an error in line 1, col 8, it will be highlighted also in the location list window).  Can you confirm this?  Otherwise it might be related to my location list handling (e.g. auto-resizing it).
2. I think using something like `cterm=underline` for the highlighting groups would be less distracting.
3. It should be reset/hidden for the current line if that gets changed (via `TextChanged/TextChangedI` maybe)?  Because the position will usually get out of sync then.

blueyed:
There are also highlights in a new terminal buffer (via `:terminal` or fzf (which uses a terminal) sometimes).
I still have not looked at the code itself in more depth..

paretje:
Ah, well, highlights are global in vim. They are removed when processing a buffer by `neomake#ProcessCurrentWindow`. I guess it's never called in a terminal buffer. Can be fixed by adding a autocmd to do so.

It currently uses caches by buffer. I know neomake switched to window variables, I guess it could be calculated on the fly using them. What was the exact reason to use window variables, and not ie buffer variables?

blueyed:
> What was the exact reason to use window variables, and not ie buffer variables?

IIRC it was in the context of having the same buffer in multiple windows.

Do you plan to update this PR (needs to be rebased, too).

paretje:
I'll update and rebase the PR next week.

mhartington:
Not sure where @paretje is with the update, but I have a rebased version of this that I've been testing locally. Would you like me to submit a new PR and go from there?

paretje:
I'm sorry for the delay, but I rebased the PR, and fixed the issue with terminal buffers. Id's are now kept track of, as you suggested. I remember I did try it before, but I believe I used a script variable, which of course didn't work, as the highlights are per window.

teto:
> Maybe I should make it an option to highlight the complete line when no column is known. Then you could use this instead of the signs
I wish to see that :) !
I've checked out the PR but don't see any change (latex or python), I guess I should run it on a C++ project as clang does return a column number ?

bounceme:
here is a helper for match creation:
```
function! neomake#highlight#MatchAdd(hi,pos) abort
  if exists('*matchaddpos')
    return matchaddpos(a:hi,a:pos)
  endif
  return matchadd(a:hi,'\%' . a:pos[0] . 'l\%' . a:pos[1] . 'c')
endfunction
```

JelteF:
I just created my own fork with these changes. I have also added support to supply the optional length of key by the maker to highlight more than just the first letter. I implemented this in a basic way for the flake8 maker. https://github.com/JelteF/neomake/tree/highlight (I also rebased it on current master, which could be done automatically)
I've only tested it very shortly, but it seems to work quite well.

Also, could someone with Travis permissions restart the build for this pull request, because I think the current failure might not be correct.

blueyed:
For Neovim there is an API for buffer highlight!
See `:h api-highlights`:

> Nvim allows plugins to add position-based highlights to buffers. This is
similar to |matchaddpos()| but with some key differences. The added highlights
are associated with a buffer and adapts to line insertions and deletions,
similar to signs. It is also possible to manage a set of highlights as a group
and delete or replace all at once.

I think this should be used here.

paretje:
@blueyed: OK, to gather everything, I guess I need to do this:
- [ ] support for new neovim api
- [ ] support to highlight complete line
- [ ] cherry-pick @JelteF's changes to support length in matches

@bounceme: why? `matchaddpos` has been added in 7.4.330, and neomake already requires 7.4.503.

bounceme:
up to you

paretje:
When I added the option to highlight the complete line, I realised I should add documentation for the various options. I added this to the TODO list.

paretje:
@blueyed: I think this PR is ready now. What do you think?

paretje:
Rebased on current master.

Alok:
+1 to this

autoload/neomake.vim
```diff
@@ -885,8 +894,14 @@ function! neomake#CleanOldFileSignsAndErrors(...) abort
         if has_key(s:current_errors['file'], bufnr)
             unlet s:current_errors['file'][bufnr]
         endif
+<<<<<<< HEAD
```
JelteF: You forgot to merge this

autoload/neomake/makers/ft/python.vim
```diff
@@ -83,6 +83,11 @@ function! neomake#makers#ft#python#Flake8EntryProcess(entry) abort
     else
         let type = ''
     endif
+    let l:token = matchstr(a:entry.text, "'.*'")
+    if strlen(l:token)
+        let a:entry.length = strlen(l:token) - 2 " subtract the quotes
+    endif
```
blueyed: @JelteF @paretje 
What is this needed for?
Should be a separate PR probably.

paretje: @paretje: that's an extension of @JelteF to have highlights of variable lengths when makers produce such information. What should be separated? The implementation for python, or the whole feature?

blueyed: Oh, it was not clear to me (and still is not): then length will only be set if the text is in quotes?
How does entry.length get set in general?  (is it only overwritten here?)

blueyed:
Tests are failing on Neovim (https://travis-ci.org/neomake/neomake/jobs/175132001#L779), which might be an issue with the tests though.

I would really like to see new tests for this feature, too.

paretje:
I have encountered the error from Travis before. I investigated it, and for a moment I thought it was a bug in neovim, but apparently the bufnr passed isn't always a number. It's easy to convert it back to a number, but I'll investigate what's happening.

autoload/neomake/makers/ft/python.vim
JelteF: Length will indeed only be set if text is in quotes (indicating a specific word). It's not full proof but it works quite well for flake8 errors. I've been using this branch for a while now and the only error for which it is wrong is the unused import one.

Length can be set optionally, because not all linters/makers will be able to support it. However, for ones that do it can be really useful. I'm mostly working on python projects with flake8 so I added support for that one, allowing possible additions for other makers to be done later.

paretje: Well, by default the length is one column. Some makers contain information which can be used for this, e.g. to mark a non-existing function. The only maker currently setting `length` is `flake8`.

In this case, when you have a file:
```python
#!/usr/bin/python
nonexist
```

`flake8 --format=default test.py`:
```
test.py:2:1: F821 undefined name 'nonexist'
```

paretje:
I found and fixed the issue. @blueyed: would something like https://github.com/neomake/neomake/blob/master/tests/signs.vader do for tests?

blueyed:
@paretje 
Not sure about the tests, but that's one of the later ones, so it might be a good base.  Thanks for considering to write them!

autoload/neomake/makers/ft/python.vim
blueyed: Makes sense, thanks for explaining!
@JelteF 
I have a local branch/stash to improve Python makers btw.. all my recent improvements are ways to get there.. the current one being to fix "serialize": I would like to run the "python" maker serialized, and when it bails out (syntax errors) skip the others.

JelteF:
Is there anything holding this up to be merged? (except for the obvious merge conflict at the moment) I've been using this as my main branch for quite a while now and have not noticed any issues.

paretje:
As far as I know, I just have the write the tests.

blueyed:
Oops.. ;)

I got the impression that @paretje is willing to write tests after the fact instead.
(came here via the [Neovim #7 Newsletter](https://neovim.io/news/2016/11/), which mentioned a feature being used here).

Alok:
=)

autoload/neomake/makers/ft/python.vim
blueyed: Found some issue regarding this: https://github.com/neomake/neomake/issues/824.

PR #362, opened 2016-03-30
Improve shell makers

paretje:
I did the following improvements to the shell makers:
 - fix error type returned by shellcheck
 - add optional `checkbashism` maker
 - add maker using shell

For the last one, I try to detect the correct shell from the first line of the script, otherwise use `/usr/bin/sh`.

### DIFF
--- a/autoload/neomake/makers/ft/sh.vim
+++ b/autoload/neomake/makers/ft/sh.vim
@@ -22,3 +22,17 @@ function! neomake#makers#ft#sh#ShellcheckEntryProcess(entry)
     endif
     return a:entry
 endfunction
+
+function! neomake#makers#ft#sh#checkbashisms()
+    return {
+        \ 'args': ['-fx'],
+        \ 'errorformat':
+            \ '%-Gscript %f is already a bash script; skipping,' .
+            \ '%Eerror: %f: %m\, opened in line %l,' .
+            \ '%Eerror: %f: %m,' .
+            \ '%Ecannot open script %f for reading: %m,' .
+            \ '%Wscript %f %m,%C%.# lines,' .
+            \ '%Wpossible bashism in %f line %l (%m):,%C%.%#,%Z.%#,' .
+            \ '%-G%.%#'
+        \ }
+endfunction
--- a/autoload/neomake/makers/ft/sh.vim
+++ b/autoload/neomake/makers/ft/sh.vim
@@ -10,6 +10,15 @@ function! neomake#makers#ft#sh#shellcheck()
         \ 'errorformat':
             \ '%f:%l:%c: %trror: %m,' .
             \ '%f:%l:%c: %tarning: %m,' .
-            \ '%f:%l:%c: %tote: %m'
+            \ '%f:%l:%c: %tote: %m',
+        \ 'postprocess':
+            \ function('neomake#makers#ft#sh#ShellcheckEntryProcess')
         \ }
+endfunction
+
+function! neomake#makers#ft#sh#ShellcheckEntryProcess(entry)
+    if a:entry.type ==? 'N'
+        let a:entry.type = 'W'
+    endif
+    return a:entry
 endfunction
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/sh.vim
```diff
@@ -10,6 +10,15 @@ function! neomake#makers#ft#sh#shellcheck()
...
+            \ function('neomake#makers#ft#sh#ShellcheckEntryProcess')
         \ }
 endfunction
+
+function! neomake#makers#ft#sh#ShellcheckEntryProcess(entry)
+    if a:entry.type ==? 'N'
+        let a:entry.type = 'W'
+    endif
+    return a:entry
+endfunction
```
blueyed: Can you elaborate on why that's needed, please?

paretje: `N` is not a valid entry type, because of which it is marked as an error. Similar functions are seen ie. in the python makers.

blueyed: `N` seems to come from `Note:` entries?!
Do you have an example for that?

paretje:
```
bin/3gmonitor:7:27: note: Double quote to prevent globbing and word splitting. [SC2086]
```

autoload/neomake/makers/ft/sh.vim
```diff
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#sh#EnabledMakers()
-    return ['shellcheck']
+    return ['shellcheck', 'sh']
```
blueyed: I've thought to have asked already: shouldn't `sh` come first?  (in case you're running the makers successively)

autoload/neomake/makers/ft/sh.vim
blueyed: Ah, I've seen those, but haven't noticed that it's a note and not a warning.

autoload/neomake/makers/ft/sh.vim
paretje: Probably, yes, if only for the sake of consistency. I'll change it.

blueyed:
Thanks!

autoload/neomake/makers/ft/sh.vim
blueyed: @paretje 
> `N` is not a valid entry type, because of which it is marked as an error.

Where is this marked as an error?

I think this should become `I` instead?!

paretje: Ie `autoload/neomake/neomake.vim`

```vim
function! neomake#signs#PlaceSign(entry, type) abort
    if a:entry.type ==? 'W'
        let sign_type = 'neomake_warn'
    elseif a:entry.type ==? 'I'
        let sign_type = 'neomake_info'
    elseif a:entry.type ==? 'M'
        let sign_type = 'neomake_msg'
    else
        let sign_type = 'neomake_err'
    endif
```

paretje: The reason I picked `W` is that it's used in the python maker for notes, and in the shellcheck checker of syntastic.

wsdjeg:
for most file type, we use compiler as maker,for example in java file we use javac, it is a compiler not execute current file.

maybe we should not use sh for sh file check. it is dangerous,if we have dangerous command in the script such as`rm /home/foo`.

and the script should be checked in sandbox

paretje:
This shouldn't really be an issue according to the docs:
```
           -n noexec        If not interactive, read commands but do not execute them.  This is useful for checking the syntax of shell scripts.
```

I believe similar things are done for ruby and python. This should be supported by any POSIX compliant [shell](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html), although it might be a good idea to do sanity checks on the detected shell.

wsdjeg:
sorry,I do not know this feature.

PR #365, opened 2016-03-30
Fix some error type specific code

paretje:
The error type should be checked using case insensitive comparison. At some places, case sensitive comparison is used however.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -646,7 +646,7 @@ function! neomake#EchoCurrentError() abort
 
     let s:neomake_last_echoed_error = ln_errors[0]
     for error in ln_errors
-        if error.type ==# 'E'
+        if error.type ==? 'E'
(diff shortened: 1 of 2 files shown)


blueyed:
> The error type should be checked using case insensitive comparison

Why?

blueyed:
Is this related to https://github.com/neomake/neomake/pull/364 ?
Or some other specific maker?

I think it should be handled/mapped by the maker then instead.

paretje:
I'm sorry, I kind of forgot about this PR. If I recall correctly, I think there were two reasons for this change:
1. at multiple places, case insensitive comparisons are already used, so I think it's even just a matter of consistency
2. neovim and vim didn't behave exactly the same, I believe

blueyed:
Thanks for the PR, but I am closing this for now, until proven otherwise.

PR #366, opened 2016-03-31
added markdownlint maker for markdown

sbdchd:


blueyed:
Thanks.

Does it make sense to use both 'proselint' and 'markdownlint' (by default) when they are installed together, or should one of them get preferred?

sbdchd:
@blueyed I think it makes sense to only have one markdown linter enabled by default, `mdl` or `markdown lint`. 

[Proselint](http://proselint.com) serves a different purpose in that it doesn't lint markdown syntax, instead it lints the contents of a file, the proses.

So I think it is best to have `proselint` enabled along side a markdown linter.

README.md
```diff
@@ -170,7 +170,8 @@ Standard ML:
 - smlnj
 
 Markdown:
-- mdl
+- [markdownlint](https://github.com/igorshubovych/markdownlint-cli)
+- [mdl](https://github.com/mivok/markdownlint)
```
blueyed: Shouldn't proselint be added here then, too?

sbdchd:
updated markdown portion of readme to include proselint

blueyed:
Thanks!

PR #368, opened 2016-03-31
Added xo type checker

ranisalt:
Adds [sindresorhus/xo](https://github.com/sindresorhus/xo) linter, which is heavily based on eslint

### DIFF
--- a/autoload/neomake/makers/ft/javascript.vim
+++ b/autoload/neomake/makers/ft/javascript.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#javascript#EnabledMakers()
-    return ['jshint', 'jscs', 'eslint']
+    return ['jshint', 'jscs', 'eslint', 'xo']
 endfunction
 
 function! neomake#makers#ft#javascript#jshint()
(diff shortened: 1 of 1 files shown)


blueyed:
Does is make sense to run both xo and eslint then by default?
Or should one of them get preferred?
You could check for this with `executable` - see other makers, e.g. the python one.

ranisalt:
I will take a look closely later

aymericbeaumet:
@blueyed eslint is more popular so it should be the default, plus it's not opinionated.

Also the options are slightly different, here is what I use:

```
        let g:neomake_javascript_xo_maker = {
        \   'args': [ '--compact' ],
        \   'errorformat': '%E%f: line %l\, col %c\, Error - %m,' .
        \                  '%W%f: line %l\, col %c\, Warning - %m',
        \ }
```

blueyed:
@aymericbeaumet 
Can you create a PR for that then, please?

This PR here seems to make sense nonetheless, right?

blueyed:
I see now.. if `eslint` is available, should it get used instead of all other makers, i.e. is it a meta-maker, like pylama for Python?

aymericbeaumet:
`eslint` is not a metalinter, it's just a configurable linter.
`xo` has been built on the top of `eslint`, and brings an opinionated configuration.

PR: https://github.com/neomake/neomake/pull/561

PR #383, opened 2016-04-21
Specify format for flake8 that errorformat actually parses.

RyanMcG:
Without this, flake8 may read local configuration and pickup an output format which is incompatible with the errorformat defined in neomake.


blueyed:
Thanks!

RyanMcG:
🙇 
Thanks you!

PR #386, opened 2016-04-24
Enable eslint_d maker for javascript supersets

whilestevego:
Given that `eslint_d` is essentially a drop in replacement for `eslint`, I thought it would be a reasonable addition to add `eslint_d` maker support wherever `eslint` is supported in javascript's supersets.

### DIFF
--- a/autoload/neomake/makers/ft/jsx.vim
+++ b/autoload/neomake/makers/ft/jsx.vim
@@ -21,3 +21,7 @@ endfunction
 function! neomake#makers#ft#jsx#eslint()
     return neomake#makers#ft#javascript#eslint()
 endfunction
+
+function! neomake#makers#ft#jsx#eslint_d()
+    return neomake#makers#ft#javascript#eslint_d()
+endfunction
--- a/autoload/neomake/makers/ft/angular.vim
+++ b/autoload/neomake/makers/ft/angular.vim
@@ -16,6 +16,10 @@ function! neomake#makers#ft#angular#eslint()
     return neomake#makers#ft#javascript#eslint()
 endfunction
 
+function! neomake#makers#ft#angular#eslint_d()
+    return neomake#makers#ft#javascript#eslint_d()
+endfunction
+
 function! neomake#makers#ft#angular#jscs()
     return neomake#makers#ft#javascript#jscs()
 endfunction
--- a/autoload/neomake/makers/ft/jasmine.vim
+++ b/autoload/neomake/makers/ft/jasmine.vim
@@ -16,6 +16,10 @@ function! neomake#makers#ft#jasmine#eslint()
     return neomake#makers#ft#javascript#eslint()
 endfunction
 
+function! neomake#makers#ft#jasmine#eslint_d()
+    return neomake#makers#ft#javascript#eslint_d()
+endfunction
+
 function! neomake#makers#ft#jasmine#jscs()
     return neomake#makers#ft#javascript#jscs()
 endfunction
--- a/autoload/neomake/makers/ft/node.vim
+++ b/autoload/neomake/makers/ft/node.vim
@@ -16,6 +16,10 @@ function! neomake#makers#ft#node#eslint()
     return neomake#makers#ft#javascript#eslint()
 endfunction
 
+function! neomake#makers#ft#node#eslint_d()
+    return neomake#makers#ft#javascript#eslint_d()
+endfunction
+
 function! neomake#makers#ft#node#jscs()
     return neomake#makers#ft#javascript#jscs()
 endfunction


blueyed:
Would 'eslint_d' be preferred in case both are installed?

whilestevego:
Good question, @blueyed. I'm not entirely sure, but I will figure it out and let you know. 😄

whilestevego:
My observations so far.

If `g:neomake_javascript_enabled_makers` isn't set to include `eslint_d`, it will just run `eslint` even if you have both installed given the defaults:
```viml
function! neomake#makers#ft#javascript#EnabledMakers()
    return ['jshint', 'jscs', 'eslint']
endfunction
```
If you happen to set both in `g:neomake_javascript_enabled_makers` (i.e. `let g:neomake_javascript_enabled_makers = ['eslint', 'eslint_d']`), both will run. I would imagine that you would get similar behaviour if you had jshint, jscs and eslint installed. All will run. Makes sense to me.

The section below in the subset appears to be acting as a white list. It will only allow the makers in the array to run for that superset regardless of which makers are enabled for javascript.
```viml
function! neomake#makers#ft#jsx#EnabledMakers()
    return ['jshint', 'eslint', 'eslint_d']
endfunction
```

To recap, neither `eslint` or `eslint_d` will have priority. They will both run if they're set in `g:neomake_javascript_enabled_makers`. But, `eslint_d` is not enabled by default therefore only `eslint` will run unless `eslint_d` is manually enabled in your config file.

blueyed:
This PR adds `eslint_d` to the list of enabled markers, so it will be run by default, no?

I was more thinking in terms of how other makers behave, e.g. the Python one (https://github.com/neomake/neomake/blob/f77f40609d814665475592254782f9259f01d75f/autoload/neomake/makers/ft/python.vim#L8-L20), where certain tools get priority over others (for the default list).

If `eslint` and `eslint_d` are similar, it makes no sense to enable both by default, does it?
See also https://github.com/neomake/neomake/blob/3a76e9cb873659136e8838d5c52f56e909e93707/autoload/neomake/makers/ft/d.vim#L7-L13.

But be aware of https://github.com/neomake/neomake/pull/400, which I'd like to merge soonish - so you should use `executable()` probably directly - although the `Exists` function will still be there.

autoload/neomake/makers/ft/jsx.vim
```diff
@@ -5,7 +5,7 @@ function! neomake#makers#ft#jsx#SupersetOf()
 endfunction
 
 function! neomake#makers#ft#jsx#EnabledMakers()
-    return ['jshint', 'eslint']
+    return ['jshint', 'eslint', 'eslint_d']
```
blueyed: Is there a specific reason to not have `jscs` here?

blueyed:
@whilestevego 
Sorry, I somehow missed the point that you were only adding it to other filetypes (and that it's not enabled for `javascript` currently).
However, you're enabling it for the other filetypes by default, and that should be removed: you should only add the functions so that it will be available.

And then the README should get updated probably, too.

Given that the filetypes are very similar it seems bad to duplicate all the functions however, and maybe it could be refactored somehow to make e.g. `node` derive from `javascript` in a better way?!

autoload/neomake/makers/ft/jsx.vim
whilestevego: `jscs` was merged with `eslint`. Maybe, it should be removed as a default from javascript's enabled makers.

whilestevego:
@blueyed 
>However, you're enabling it for the other filetypes by default, and that should be removed: you should only add the functions so that it will be available.

I agree. I only added it there, because of a misunderstanding on my part.

>Given that the filetypes are very similar it seems bad to duplicate all the functions however, and maybe it could be refactored somehow to make e.g. node derive from javascript in a better way?!

That repetition certainly seems unnecessary. As far as I can tell, only javascript has supersets. Maybe, we can dynamically create the mapping functions from some sort of list.

With these PRs, my intent was to add support for `eslint_d` in javascript supersets and this PR achieves that in the pattern that's already been established. Can we address the function duplication in another task? I'd be happy to work on it although my VimL skills are a bit lacking.

whilestevego:
https://github.com/neomake/neomake/pull/386/commits/f0a1404977139d38a8e99a333fe381c44f10a371 squashed my commits and removed `eslint_d` as a default maker from all supersets.

blueyed:
> Can we address the function duplication in another task?

Of course.

> I'd be happy to work on it although my VimL skills are a bit lacking.

That would be awesome and a good training then.. :)

I've not investigated though if that is really required / useful after all - just some thought that came up.
Maybe it's better to leave it as it is for now.  After all it's more explicit and easy to understand that way.
Especially when it would only affect javascript.

Thanks for your nice contribution and then good discussion!

whilestevego:
Thank you, @blueyed! Thanks for maintaining this fantastic project.

PR #387, opened 2016-04-24
add running status for statusline

chemzqm:
User could use function `neomake#statusline#GetRunningStatus()` to show neomake running status in their statusline setting.

 `...` is returned when neomake is running for this buffer and window, empty string returned if not.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -95,7 +95,7 @@ function! neomake#MakeJob(maker) abort
         let cwd = expand(a:maker.cwd, 1)
         exe 'cd' fnameescape(cwd)
     endif
-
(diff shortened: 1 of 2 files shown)


blueyed:
From a quick skim I think this should take the current tab into account, too?

blueyed:
Needs to be rebased also.

phcerdan:
:+1: for this feature.

blueyed:
@chemzqm 
Why have you closed it?
Are you using it still, or in some other form?
See also https://github.com/neomake/neomake/issues/915#issuecomment-271120716.

chemzqm:
@blueyed Found some issue with it but couldn't figure out the way to fix.
I'm not using it any more.

PR #391, opened 2016-04-30
added luac maker for lua

sbdchd:


autoload/neomake/makers/ft/lua.vim
```diff
@@ -1,5 +1,5 @@
 function! neomake#makers#ft#lua#EnabledMakers()
-    return ['luacheck']
+    return ['luacheck', 'luac']
```
blueyed: Should this come first, because it's faster and does basic syntax checks only?

blueyed:
I don't think both should be enabled by default, since `luacheck` seems to report the same fatal parse errors, doesn't it?

So I suggest to use only `luacheck` by default if it exists.
Do you agree?

sbdchd:
@blueyed sounds like a plan

blueyed:
Thanks!

PR #393, opened 2016-05-01
Implement support for on-the-fly linting

ivalkeen:
This is my attempt to implement  on-the-fly linting using this proposal: https://github.com/benekastah/neomake/issues/190

To actually run `Neomake` on the fly I've added this to my vimrc:

```
autocmd TextChanged,TextChangedI * Neomake
```

Using `CursorHold` is another option.

Rubocop and Eslint makers were changed in order to use this feature.
To use it with different linters, `pipe` option needs to be set to 1
and maker arguments need to be adjusted to handle stdin as source.

Here is an example of how it is configured for rubocop:

```
function! neomake#makers#ft#ruby#rubocop()
    return {
        \ 'pipe': 1,
        \ 'args': ['--format', 'emacs', '--stdin'],
        \ 'errorformat': '%f:%l:%c: %t: %m',
        \ 'postprocess': function('neomake#makers#ft#ruby#RubocopEntryProcess')
        \ }
endfunction
```

Tested on both Neovim (0.1.4) and Vim (7.4.1795)

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -148,6 +148,12 @@ function! s:MakeJob(make_id, maker) abort
                 let s:jobids_by_makeid[a:make_id] = []
             endif
             call add(s:jobids_by_makeid[a:make_id], jobinfo.id)
+
+            if get(a:maker, 'pipe', 0)
+                call jobsend(job, getline(1, '$'))
+                call jobsend(job, "\n")
+                call jobclose(job, 'stdin')
+            endif
         else
             " Vim, synchronously.
             if has_args
@@ -162,7 +168,15 @@ function! s:MakeJob(make_id, maker) abort
             let jobinfo.id = job_id
             let s:jobs[job_id] = jobinfo
             call s:AddJobinfoForCurrentWin(jobinfo.id)
-            call neomake#MakeHandler(job_id, split(system(program), '\r\?\n', 1), 'stdout')
+
+            if get(a:maker, 'pipe', 0)
+                let stdin = join(getline(1,'$'), "\n")."\n"
+                let cmd = system(program, stdin)
+            else
+                let cmd = system(program)
+            endif
+
+            call neomake#MakeHandler(job_id, split(cmd, '\r\?\n', 1), 'stdout')
             call neomake#MakeHandler(job_id, v:shell_error, 'exit')
             let r = 0
         endif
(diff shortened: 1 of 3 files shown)


outpunk:
@benekastah @natemara Guys, please accept this PR :)

natemara:
Sorry, I don't actually have merge access

blueyed:
@outpunk 
Have you tested/reviewed it?

I've skimmed it a while ago and found it hard to understand: doesn't it just use `stdin` instead of a file?
But the whole point here is probably to not having to write the file, right?

What about a more general approach then: writing a temporary file instead? (using `writefile()`)
Will it could problems because of a different filename/path?

ivalkeen:
@blueyed yes, it just uses `stdin` instead of file, changes configuration for linters to read from `stdin` and one important thing here is that it still passes original file name as a parameter to the linter, because linters are smart enough to get both `stdin` and file name and generate errors messages using the file name but `stdin` as the source. I think this last part may be tricky to do in case of writing a temporary file, however I didn't think about this option enough. 

Could you please explain, what was hard to understand in this PR? I don't think it's ready to be merged, it is more like a proof of concept, but I will be happy to finish it and make it ready following you suggestions if you think it makes sense.

blueyed:
> Could you please explain, what was hard to understand in this PR?

Initially I've thought that linters are meant to be run as a service/daemon, constantly piping in the changes only (instead of the whole file).

ivalkeen:
I've rebased this branch against master, because there were some conflicts.

purpleP:
@blueyed @ivalkeen Hi, can anyone say when it would be merged? And why it itsn't merged yet?

Alok:
I'd also like to see this merged soon.

blueyed:
Sadly it conflicts again (needs to be rebased).
As for merging (and maintaining) stuff: feel free to step up and help with triaging and reviewing etc, and become a maintainer yourself.

@ivalkeen 
Given that this PR is rather simple, I will merge it once it's updated.

blueyed:
@ivalkeen 
Ping.
There is #517 now for the rebasing.

I am feeling a bit uncomfortable merging things with an high impact to break stuff without any tests and no response for weeks..

ivalkeen:
I've rebased it again. I agree with @blueyed, there is a hight chance to break stuff with this PR and it should not be merged before tests are added. It's not easy, but I will try to find some time to finish and test it.

ivalkeen:
I think I agree that this approach is too hacky and not all linters support reading from STDIN etc. There are other solutions for the same problem (https://github.com/dojoteef/neomake-autolint) which are probably better. Closing this one.

PR #394, opened 2016-05-01
Only handle CursorMoved when cursor was moved between lines

ivalkeen:
Moving cursor left and right staying on the same line causes lags,
because visible signs and error messages are redrawn on every move.

Looks like there is no need to do it, because the first error message
for the line is shown anyway, and visible area is not changed.

This PR is my naive attempt to avoid this lag by handling `CursorMoved`
only when line under cursor was changed.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -577,6 +577,7 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
 
         " Show the current line's error
         call neomake#CursorMoved()
+        call neomake#EchoCurrentError()
 
         if has_key(maker, 'next')
             let next_makers = '['.join(maker.next.enabled_makers, ', ').']'
@@ -647,8 +648,12 @@ function! neomake#EchoCurrentError() abort
 endfunction
 
 function! neomake#CursorMoved() abort
-    call neomake#signs#PlaceVisibleSigns()
-    call neomake#EchoCurrentError()
+    let l:line = line('.')
+    if get(s:, 'last_cursormoved_line', 0) != l:line
+        let s:last_cursormoved_line = l:line
+        call neomake#signs#PlaceVisibleSigns()
+        call neomake#EchoCurrentError()
+    endif
 endfunction
 
 function! neomake#CompleteMakers(ArgLead, ...)
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -576,7 +576,6 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
         endif
 
         " Show the current line's error
-        call neomake#CursorMoved()
         call neomake#EchoCurrentError()
 
         if has_key(maker, 'next')


autoload/neomake.vim
```diff
@@ -656,8 +656,11 @@ function! neomake#EchoCurrentError() abort
 endfunction
 
 function! neomake#CursorMoved() abort
-    call neomake#signs#PlaceVisibleSigns()
-    call neomake#EchoCurrentError()
+    if !exists("s:recordedLine") || line(".") != s:recordedLine
+        let s:recordedLine = line(".")
```
blueyed: I'd use `s:last_cursormoved_line` instead.
Also single quotes are used around here.

blueyed:
Nice, just a small nitpick.

blueyed:
Looked at this some more.
I think echoing the current error is good to do in general (after all it's `CursorHold`, which is not triggered while moving).

But I am not sure if `neomake#signs#PlaceVisibleSigns` should be called on `CursorHold` at all?!  (and then this PR might be void)

Anyway, there's a call to CursorHold, which did not work with this PR anymore, so the cache needs to be invalidated, or the relevant message being called directly.

Here's a diff I came up with against this PR:

```diff
--- d1	2016-05-15 15:50:59.067849023 +0200
+++ d2	2016-05-15 15:51:12.214545572 +0200
@@ -1,4 +1,4 @@
-commit 17fdee4 (origin/pr/394)
+commit 5aa6af0 (HEAD -> master)
 Author: Ivan Tkalin <itkalin@gmail.com>
 Date:   Thu Mar 24 16:30:15 2016 +0100
 
@@ -12,19 +12,31 @@
     
     This PR is my naive attempt to avoid this lag by handling `CursorMoved`
     only when line under cursor was changed.
+    
+    Closes https://github.com/neomake/neomake/pull/394.
 
 diff --git a/autoload/neomake.vim b/autoload/neomake.vim
-index d81f929..99984b7 100644
+index b5f51cb..ef2e68c 100644
 --- a/autoload/neomake.vim
 +++ b/autoload/neomake.vim
-@@ -656,8 +656,11 @@ function! neomake#EchoCurrentError() abort
+@@ -573,7 +573,7 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
+         endif
+ 
+         " Show the current line's error
+-        call neomake#CursorMoved()
++        call neomake#EchoCurrentError()
+ 
+         if has_key(maker, 'next')
+             let next_makers = '['.join(maker.next.enabled_makers, ', ').']'
+@@ -647,8 +647,12 @@ function! neomake#EchoCurrentError() abort
  endfunction
  
  function! neomake#CursorMoved() abort
 -    call neomake#signs#PlaceVisibleSigns()
 -    call neomake#EchoCurrentError()
-+    if !exists("s:recordedLine") || line(".") != s:recordedLine
-+        let s:recordedLine = line(".")
++    let l:line = line('.')
++    if get(s:, 'last_cursormoved_line', 0) != l:line
++        let s:last_cursormoved_line = l:line
 +        call neomake#signs#PlaceVisibleSigns()
 +        call neomake#EchoCurrentError()
 +    endif
```

blueyed:
@ivalkeen 
Ping.

ivalkeen:
I don't understand your comment, could you please explain? The only change here is to place signs and echo error only when cursor moved between lines (vertically), and it is attached to `CursorMoved` autocommand. So I don't really understand what it has to do with `CursorHold`, which you are referring in your comment.

blueyed:
Oh sorry, I've mixed those up - I think you can ignore most of it.

But apart from that I think the diff still is necessary.

ivalkeen:
@blueyed I've updated it according to your recommendations. Could you, please, take a look again?

blueyed:
Thanks!

The change looks good.. I'll try it for a while, but consider it to be merged soon hopefully.. :)

ivalkeen:
Wunderbar!

autoload/neomake.vim
```diff
@@ -577,6 +577,7 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
 
         " Show the current line's error
         call neomake#CursorMoved()
```
blueyed: I think this line is not really necessary.

Only calling `neomake#EchoCurrentError` (added below) seemed to be enough (see my diff).

autoload/neomake.vim
```diff
@@ -577,6 +577,7 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
...
+        call neomake#EchoCurrentError()
```
blueyed: Or this line should not be added..

autoload/neomake.vim
ivalkeen: Yes, we can remove this one. The next one should stay, otherwise if there is a new error it won't be displayed (because the line number isn't changed).

blueyed:
Merged, with some additional patch to take into account the current buffer:

```diff
diff --git a/autoload/neomake.vim b/autoload/neomake.vim
index 9b619a8..a3ca39b 100644
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -691,10 +691,11 @@ function! neomake#EchoCurrentError() abort
     call neomake#utils#WideMessage(message)
 endfunction
 
+let s:last_cursormoved = [0, 0]
 function! neomake#CursorMoved() abort
     let l:line = line('.')
-    if get(s:, 'last_cursormoved_line', 0) != l:line
-        let s:last_cursormoved_line = l:line
+    if s:last_cursormoved[0] != l:line && s:last_cursormoved[1] != bufnr('%')
+        let s:last_cursormoved = [l:line, bufnr('%')]
         call neomake#signs#PlaceVisibleSigns()
         call neomake#EchoCurrentError()
     endif
```

Thanks!

PR #375, opened 2016-04-09
[WIP] Update java checker

wsdjeg:
TODO:
- [x] auto generate classpath
- [x] auto detect output directry
- [x] maven support
- [x] gradle support
- [x] support eclipse project's `.classpath` file, for this feature,you need python support.
- [x] support checkstyle

PR #398, opened 2016-05-09
Add ruby-lint support

adamors:
Add rubylint maker

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ Ruby:
 - jruby
 - rubocop
 - reek
+- rubylint
 
 C:
 - clang
--- a/autoload/neomake/makers/ft/ruby.vim
+++ b/autoload/neomake/makers/ft/ruby.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#ruby#EnabledMakers()
-    return ['mri', 'rubocop', 'reek']
+    return ['mri', 'rubocop', 'reek', 'rubylint']
 endfunction
 
 function! neomake#makers#ft#ruby#rubocop()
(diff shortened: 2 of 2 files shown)


autoload/neomake/makers/ft/ruby.vim
```diff
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#ruby#EnabledMakers()
-    return ['mri', 'rubocop', 'reek']
+    return ['mri', 'rubocop', 'reek', 'rubylint']
```
wsdjeg: I think we should not return this makers without executable() check;
see #375 
@blueyed 
I think `EnabledMakers()` should only return the effective maker in current os;
for example
```vim
function! neomake#makers#ft#ruby#EnabledMakers()
      let makers = []
      if executable('ruby-lint')
           call add(makers,'ruby-lint')
      endif
      return makers    
endf
```

adamors: Except MRI, every Ruby maker needs an existing executable.

wsdjeg: and also you can provide a g:var for user to set the executable.

blueyed: @wsdjeg 
The "executable" check is done in https://github.com/neomake/neomake/blob/1b585580f2f1d33dc115ebf86466bbb4112ed8da/autoload/neomake/utils.vim#L97-L107 (via `neomake#utils#MakerIsAvailable`) already.  But this should probably use Vim's `executable` however.
Creating a PR for that.  In this PR here it's fine.

README.md
```diff
@@ -113,6 +113,7 @@ Ruby:
 - mri
 - jruby
 - rubocop
+- rubylint
 - reek
```
blueyed: I think the order should reflect the actual one from `neomake#makers#ft#ruby#EnabledMakers`.

autoload/neomake/makers/ft/ruby.vim
blueyed: @wsdjeg 
> and also you can provide a g:var for user to set the executable.

That would be `g:neomake_ruby_rubylint_exe` (already supported).  See https://github.com/neomake/neomake/blob/1b585580f2f1d33dc115ebf86466bbb4112ed8da/doc/neomake.txt#L186-L190.

blueyed:
LGTM.

@adamors 
Please squash it into a single commit, although we can do so when merging (but you have more control over the commit message then).

adamors:
Okay.

blueyed:
Thanks, @adamors!

PR #399, opened 2016-05-09
Add perl's perlcritic maker

cazador481:


dafky2000:
This PR works nicely, I've simply added -4 for my own purpose.

I fought with implementing this in my init.vim config as so:
```
let g:neomake_perl_perlcritic_maker = {
 	\ 'exe': '/usr/bin/vendor_perl/perlcritic',
 	\ 'args': ['--quiet', '--nocolor', '--verbose "%f:%l:%c:%m (%e)\\n"', '-4'],
 	\ 'errorformat': '%f:%l:%c:%m',
 	\ }
```

Would like to know how I can make "my way" work so I can keep it all together in my init.vim until this gets merged in. PerlCritic seems to return exit code 2 naturally so neomake doesn't recognize as an error as far as I found out. Would love an explanation if anyone has one. In the meantime this patch works well!

cazador481:
I have the following in my init.vim, so all you would need to do is add the -4 to the command line. or you can add severity = 4 to a  $HOME/.perlcriticrc

````perl 
let g:neomake_perl_perlcritic_maker = { 
             \'exe': 'perlcritic',
             \ 'args' : ['--quiet', '--nocolor', '--verbose', '\\%f:\\%l:\\%c:(\\%s) \\%m (\\%e)\\n'],
             \ 'errorformat': 
             \ '%f:%l:%c:%m,'
````

dafky2000:
Not sure why I thought the "extra" escaping was specific to this patch, that's likely the issue I am having. Thanks for the perlcriticrc hint too that'll be handy when this is merged in!

blueyed:
Thanks!

PR #400, opened 2016-05-09
Simplify/cleanup neomake#utils#Exists

blueyed:
I'm not sure if not just using `executable()` was intentionally?!

Inspired by https://github.com/neomake/neomake/pull/398#discussion_r62559441.

### DIFF
--- a/autoload/neomake/makers/ft/c.vim
+++ b/autoload/neomake/makers/ft/c.vim
@@ -1,9 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#c#EnabledMakers()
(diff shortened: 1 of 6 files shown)


wsdjeg:
should we use `executable(expand(exe,1))` or `shellescape()`

blueyed:
@wsdjeg 
> should we use `executable(expand(exe,1))` or `shellescape()`

I don't think that makes sense in this context, does it?

blueyed:
@benekastah 
What do you think about this PR?
Am I missing something with this simplification?
I don't have a Windows box to test this, but from looking at other plugins only cygwin could use some special treatment in this regard (which is not done with the current implementation though anyway).

Issue #373, opened 2016-04-07
Use expansion/callback for maker args


hugdru:
I am trying to convert my old make vim configuration to neomake. I converted it directly according to the example in the docs but it does not work, on execution, `:Neomake! makedebug` replies `Neomake: makedebug complete with exit code 2`.

I suspect it has something to do with the way the expansion works in args. Because if I, :cd %:p:h, and create a marker with a args relative path argument ../ it executes perfectly.

** Folder structure **

```
code |
         | bin
         | src
             | file.cpp
             | file.hpp
         | makefile
         | buildtemp
```

**Old Vim**
```
nnoremap <leader>mkf :vsplit %:p:h/../makefile<CR>
inoremap <leader>mkf <ESC>:vsplit %:p:h/../makefile<CR>
vnoremap <leader>mkf <ESC>:vsplit %:p:h/../makefile<CR>
nnoremap <leader>mkc :Make clean -C %:p:h/../<CR>
vnoremap <leader>mkc <ESC>:Make clean -C %:p:h/../<CR>
inoremap <leader>mkc <ESC>:Make clean -C %:p:h/../<CR>
nnoremap <leader>mkd :wa<CR>:Make debug -C %:p:h/../<CR>
vnoremap <leader>mkd <ESC>:wa<CR>:Make debug -C %:p:h/../<CR>
inoremap <leader>mkd <ESC>:wa<CR>:Make debug -C %:p:h/../<CR>
nnoremap <leader>mka :wa<CR>:Make all -C %:p:h/../<CR>
vnoremap <leader>mka <ESC>:wa<CR>:Make all -C %:p:h/../<CR>
inoremap <leader>mka <ESC>:wa<CR>:Make all -C %:p:h/../<CR>
nnoremap <leader>mke :!<C-R>=expand("%:p:h")<CR>/../bin/
vnoremap <leader>mke <ESC>:!<C-R>=expand("%:p:h")<CR>/../bin/
inoremap <leader>mke <ESC>:!<C-R>=expand("%:p:h")<CR>/../bin/
```

**Neovim and neomake**
```
let g:neomake_makedebug_maker = {
    \ 'exe': 'make',
    \ 'args': ['debug', '-C', '%:p:h/../'],
    \ }

let g:neomake_makeall_maker = {
    \ 'exe': 'make',
    \ 'args': ['all', '-C', '%:p:h/../'],
    \ }

let g:neomake_makeclean_maker = {
    \ 'exe': 'make',
[...]

PR #409, opened 2016-05-13
Allow for maker.args to be a callback and use it for shellcheck

blueyed:
This makes shellcheck pass &ft in case there is no shebang.

Fixes https://github.com/neomake/neomake/issues/373.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -76,7 +76,21 @@ function! neomake#MakeJob(maker) abort
     endif
     let jobinfo.maker = a:maker
 
-    let args = a:maker.args
+    " Resolve exe/args, which might be a function or dictionary.
+    if type(a:maker.exe) == type(function('tr'))
+        let exe = call(a:maker.exe, [])
+    elseif type(a:maker.exe) == type({})
+        let exe = call(a:maker.exe.fn, [], a:maker.exe)
+    else
+        let exe = a:maker.exe
+    endif
+    if type(a:maker.args) == type(function('tr'))
+        let args = call(a:maker.args, [])
+    elseif type(a:maker.args) == type({})
+        let args = call(a:maker.args.fn, [], a:maker.args)
+    else
+        let args = a:maker.args
+    endif
     let append_file = a:maker.file_mode && index(args, '%:p') <= 0 && get(a:maker, 'append_file', 1)
     if append_file
         call add(args, '%:p')
@@ -95,7 +109,7 @@ function! neomake#MakeJob(maker) abort
         exe 'cd' fnameescape(cwd)
     endif
 
-    let job = s:JobStart(make_id, a:maker.exe, args)
+    let job = s:JobStart(make_id, exe, args)
     let jobinfo.start = localtime()
     let jobinfo.last_register = 0
 
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -81,6 +81,11 @@ All 'args' will be |expand()|ed: >
         \ 'exe': 'lint',
         \ 'args': ['%:p'],
         \ }
+
+The 'exe' and 'args' entries can be a function (callback) or a dictionary with
+an expected entry 'fn' (which has to be a function).
+The function has to return a string, and receives no arguments.
+
 <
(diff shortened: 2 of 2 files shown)


blueyed:
@hugdru
Have you tested it, does it work for you (as described in #373)?

hugdru:
@blueyed 
Yes it does work very nicely. However I tweaked some of the code so I could use a closure. This avoids having too many functions, for instance one for each make target: debug, clean, etc.
Here are the changes I did:
**autoload/neomake.vim**
```vimscript
    if type(a:maker.args) == type(function("tr"))
        let args = call(a:maker.args, [])
    elseif type(a:maker.args) == type({})
        let args = call(a:maker.args.fn, [], a:maker.args)
    else
        let args = a:maker.args
    endif
```
**init.vim example**
```vimscript
" Make
function! Make_args_factory(target)
  let closure = {'target': a:target}
  function! closure.fn()
    let l:makefileLocation = expand("%:p:h:h") . "/makefile"
    let l:args = [self.target, '-f', l:makefileLocation]
    return l:args
  endfunction
  return closure
endfunction

let g:neomake_makedebug_maker = {
    \ 'exe': 'make',
    \ 'args': Make_args_factory('debug'),
    \ }
let g:neomake_makeall_maker = {
    \ 'exe': 'make',
    \ 'args': Make_args_factory('all'),
    \ }
let g:neomake_makeclean_maker = {
    \ 'exe': 'make',
    \ 'args': Make_args_factory('clean'),
    \ }
```

lverweijen:
I'm not sure what problem you're trying to solve. Why are you doing your own shebang parsing if (neo)vim has already decided the file type is of type sh?

blueyed:
@lverweijen 
It is about (not) passing in the &ft to shellcheck in case there is a shebang (which would be used then).
A file might be `sh` but has no shebang etc.

This is to avoid the warning about a missing shebang from shellcheck.

lverweijen:
Why not pass '-s sh' or '-s &ft' to shellcheck and leave the shebang parsing to vim?
Doing so should not give a warning.

When would the shebang and &ft be different anyway?

blueyed:
I think the shebang should have priority (current behavior).
But if there is none, we can help by passing in the Vim filetype.
But the &ft should not get passed in if there's a shebang.

> When would the shebang and &ft be different anyway

You could always manually set it.

> and should the shebang even get priority in such a case?

Yes (but not too confident), because that's what gets used in the end.

lverweijen:
I think this approach might lead to somewhat inconsistent behaviour. For example, having ft set to 'sh' and shebang set to 'bash', would mean that:

- 'sh -n' treats your file as _sh_
- checkbashims treats your file as _sh_
- shellcheck treats your file as _bash_

By the way, checkbashisms is currently not in the list of enabled makers.

blueyed:
Very good point!
Do the other makers have a way to pass in the shell-to-be-used?

Given that we should maybe really use &ft unconditionally?!

hugdru:
Well, maybe it is better to make a pull request with support just for callbacks, and if possible closures as well. Since a couple of people need that for their workflow.

blueyed:
@hugdru 
Agreed.
So for your use case I would add the following there, right?

```vim
    elseif type(a:maker.args) == type({})
        let args = call(a:maker.args.fn, [], a:maker.args)
```

The factory itself does not make sense to include as-is, does it?

hugdru:
@blueyed 
Yes for my use case only that elseif line is needed. However for other people to use that they must do something similar to what I did in the init.vim example. Emulate a closure with a dictionary. Define the variables in the dictionary and add a function called fn with no arguments.

blueyed:
@lverweijen 
Just for reference: there is already shebang parsing for `sh` by now (initially via #362).

blueyed:
Amended.  I've removed the shellcheck part, and made it apply also for `exe`, which would be handy/useful for `neomake#makers#ft#sh#sh` then.

blueyed:
@hugdru 
What do you think?

hugdru:
@blueyed Perfect :+1:

PR #410, opened 2016-05-13
Go cwd

mediocregopher:
There's two fixes in here:

* use `go test` instead of `go build` for the main syntax check. `go build` only actually works on a `main` package, so if you're working on a library package it's not very useful. `go test` with `-c` will only compile the tests without running them. Even if there's no tests it will still make sure the package compiles. Otherwise it's virtually identical to `go build`.

* use `mapexpr` to fix the output paths from things erroring. Since `cwd` puts the command inside an inner-path, it will break the output paths on its own.

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -31,6 +31,7 @@ endfunction
 function! neomake#makers#ft#go#govet()
     return {
         \ 'exe': 'go',
+        \ 'cwd': '%:p:h',
         \ 'args': ['vet'],
         \ 'append_file': 0,
         \ 'errorformat':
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/go.vim
```diff
@@ -1,16 +1,22 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#go#EnabledMakers()
-    return ['go', 'golint', 'govet']
+    return ['gotest', 'golint', 'govet']
```
blueyed: I'd say to not change the name, but keep it as "go".
Or at least still provide the old maker function for backward compatibility.

mediocregopher:
Fixed the name

blueyed:
Thanks!

PR #411, opened 2016-05-13
suppress trailing newline

nerdrew:


nerdrew:
screen shot 2016-05-13 at 4 50 38 pm

nerdrew:
Without this change: 

screen shot 2016-05-13 at 4 51 01 pm

nerdrew:
Function that creates this output:

```
function! s:Ag(file_mode, args)
  let cmd = "ag --vimgrep --smart-case ".substitute(a:args, '\\', '\\\\', 'g')
  let custom_maker = neomake#utils#MakerFromCommand('bash', cmd)
  let custom_maker.name = cmd
  let custom_maker.remove_invalid_entries = 0
  let custom_maker.place_signs = 0
  let custom_maker.errorformat = "%f:%l:%c:%m"
  let enabled_makers =  [custom_maker]
  call neomake#Make(a:file_mode, enabled_makers) | echo "running: " . cmd
endfunction
command! -bang -nargs=* -complete=file Ag call s:Ag(<bang>0, <q-args>)
```

autoload/neomake.vim
```diff
@@ -504,41 +510,45 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
...
-                        \ + lines[1:]
+                        \ + [jobinfo.lines[-1] . get(a:data, 0, '')]
+                        \ + a:data[1:]
         else
-            let jobinfo.lines = lines
+            let jobinfo.lines = a:data
+        endif
+
+        if len(a:data) < 2
+            return
```
blueyed: What is this doing?
Does it refer to the "Each element *after* the first element represents a newline" comment?

blueyed:
This seems to help in general (using the `vint` maker) to skip the list being reset (via `setloclist`) because of an empty line at the end.

Where are those lines coming from though?

wsdjeg:
what is the different between the two pictures shown above?different corsor position?

blueyed:
@wsdjeg 
The empty line at the end.

autoload/neomake.vim
nerdrew: I'm embarrassed to admit I don't remember. I think it is just an optimization for the following change. If there are fewer than 2 lines in `a:data` then calling `sRegisterJobOutput(jobinfo, maker, jobinfo.lines[:-2])` won't do anything since `jobinfo.lines[:-2]` will be empty. Though, it looks like there's a bug since it should be checking `len(jobinfo.lines) < 2` instead :/

I think it can just get removed. Thoughts?

blueyed: > I think it can just get removed. Thoughts?

Yes, probably.

blueyed: @nerdrew 
It's still in.  Do you plan to remove it still?

nerdrew: Doh! I updated it on my local master but forgot to update this PR. Sorry for the delay!

autoload/neomake.vim
```diff
@@ -525,41 +531,41 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
...
+            let jobinfo.lines = jobinfo.lines[-1:]
             let jobinfo.last_register = now
         endif
     elseif a:event_type ==# 'exit'
         if has_key(jobinfo, 'lines')
+            if jobinfo.lines[-1] == ''
+                call remove(jobinfo.lines, -1)
+            endif
             call s:RegisterJobOutput(jobinfo, maker, jobinfo.lines)
         endif
```
blueyed: @nerdrew 
I think this is wrong and there should be no call to `RegisterJobOutput` at all on "exit": exit is not meant to receive new data.
Also the change/slicing to lines above feels weird.

nerdrew: I agree. It seems like the whole

```
        if has_key(jobinfo, 'lines')
            if jobinfo.lines[-1] == ''
                call remove(jobinfo.lines, -1)
            endif
            call s:RegisterJobOutput(jobinfo, maker, jobinfo.lines)
        endif
```

block can be deleted.

blueyed: Thanks, done in https://github.com/neomake/neomake/pull/488.

blueyed: JFI: I was wrong.. this is meant to handle unfinished lines.
Fixed in https://github.com/neomake/neomake/pull/512/commits/a56ad99d8c10eed2ab43008a2c67d6679d1e5e3b.

PR #392, opened 2016-04-30
Add swiftlint linter

brendanjerwin:
https://github.com/realm/SwiftLint

PR #412, opened 2016-05-14
add swift maker

dhcmrlchtdj:


blueyed:
There is also https://github.com/neomake/neomake/pull/392.
Does it make sense to enable both at once?

brendanjerwin:
I bet it does. Linters and compilers like peanut butter and chocolate... I'll try it out and see if it rocks.

VDKA:
It maybe nice to use SwiftPM if the directory contains a `Package.swift` file.

blueyed:
There we go, thanks!

keith:
I've created a PR for SwiftPM support https://github.com/neomake/neomake/pull/1890

Issue #339, opened 2016-03-07
Issue with the hlint checker


antoyo:
Hello.
When using the `hlint` checker, sometimes the quickfix window does not show up (and sometimes it does).
When running with `neomake_verbose = 3`, here is what I see:

```
"src/Escape.hs" 460L, 21114C written
Neomake: Starting: hlint […]/src/Escape.hs
Press ENTER or type command to continue
```

And typing `:messages` shows me the following:

```
"src/Escape.hs" 460L, 21114C written
Neomake: Starting: hlint […]/src/Escape.hs
Neomake: hlint stdout: […]/src/Escape.hs:336:13: Suggestion: Redundant bracket
Neomake: hlint stdout: Found:
Neomake: hlint stdout:   (traceShow lastTrueStatement) $
Neomake: hlint stdout:     statementEscapes lastTrueStatement || lastFalseStatementEscapes
Neomake: hlint stdout: Why not:
Neomake: hlint stdout:   traceShow lastTrueStatement $
Neomake: hlint stdout:     statementEscapes lastTrueStatement || lastFalseStatementEscapes
Neomake: hlint stdout:
Neomake: hlint stdout: 1 hint
Neomake: hlint stdout:
Neomake: hlint stdout done.
Neomake: hlint processing 10 lines of output
Neomake: hlint complete with exit code 1
Neomake: File-level errors cleaned in buffer 16
```

It seems related to issue #286.

Thanks to fix this issue.

lwm:
Oh great, this was fixed :+1: :beers:

PR #414, opened 2016-05-14
Include hlint suggestions in error format

etcinit:
`hlint` has three message types: [Errors, Warnings, and Suggestions](https://github.com/ndmitchell/hlint#what-is-the-difference-between-errorwarningsuggestion). Currently, neomake's default hlint format only parses Errors and Warnings. This PR adds Suggestions to the error format as informational messages.

This should address issues like #339 where users mentioned the quickfix window not showing on all occasions (most likely due to missing Suggestions).


blueyed:
Thanks!

PR #416, opened 2016-05-15
fixed typo in function def for sml maker

sbdchd:


sbdchd:
Oops. Didn't see #367

PR #420, opened 2016-05-16
vulture: a new python maker added

jabbalaci:
A new Python maker called vulture is added (https://bitbucket.org/jendrikseipp/vulture). Vulture finds unused classes, functions and variables in your code.

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -19,10 +19,6 @@ function! neomake#makers#ft#python#EnabledMakers()
         call add(makers, 'pylint')  " Last because it is the slowest
     endif
 
-    if neomake#utils#Exists('vulture')
-        call add(makers, 'vulture')
-    endif
-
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/python.vim
```diff
@@ -19,6 +19,10 @@ function! neomake#makers#ft#python#EnabledMakers()
         call add(makers, 'pylint')  " Last because it is the slowest
     endif
 
+    if neomake#utils#Exists('vulture')
```
blueyed: The exists check is done automatically.

Here it is used to e.g. only use `pylama` if it exists.
I think you want to just add it unconditionally.

blueyed: But is probably better to not enable it by default, given it's false positives, e.g. on class attributes / module functions etc.

jabbalaci:
The explicit Exists call is removed. Vulture's errorformat is added.

PR #423, opened 2016-05-17
Add zsh support

lverweijen:
Uses checkshell and the native zsh checker

### DIFF
--- a/autoload/neomake/makers/ft/zsh.vim
+++ b/autoload/neomake/makers/ft/zsh.vim
@@ -1,0 +1,31 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#zsh#EnabledMakers()
+    return ['zsh', 'shellcheck']
+endfunction
+
+function! neomake#makers#ft#zsh#shellcheck()
+    return {
+        \ 'args': ['-fgcc', '--shell', 'zsh'],
+        \ 'errorformat':
+            \ '%f:%l:%c: %trror: %m,' .
+            \ '%f:%l:%c: %tarning: %m,' .
+            \ '%f:%l:%c: %tote: %m',
+        \ 'postprocess':
+            \ function('neomake#makers#ft#zsh#ShellcheckEntryProcess')
+        \ }
+endfunction
+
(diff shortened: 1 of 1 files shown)


blueyed:
I think the shellcheck function from `sh` should be used?!

See https://github.com/neomake/neomake/pull/409 for an improvement to it.

blueyed:
@lverweijen 
Do you plan to update it based on my comment?
Otherwise I might adopt it.

lverweijen:
I don't have time to update and test it now and the current version fulfils all my short term needs. If I decide to update it in the future, I don't mind making a new pull request.

autoload/neomake/makers/ft/zsh.vim
```diff
@@ -0,0 +1,31 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#zsh#EnabledMakers()
+    return ['zsh', 'shellcheck']
+endfunction
+
+function! neomake#makers#ft#zsh#shellcheck()
+    return {
+        \ 'args': ['-fgcc', '--shell', 'zsh'],
```
blueyed: That does not work:

    % shellcheck --shell zsh
    Unknown shell: zsh

blueyed: FWIW: It was removed in ac3cfe8.

lverweijen:
This works for me

    $shellcheck --shell zsh .zshrc

I'm using shellcheck 0.3.4.

lverweijen:
I have implemented the shebang check and other improvements from sh.vim

With regards to chellcheck, maybe it can replace the error message by one that informs the user that the version of shellcheck is too old.

blueyed:
Oh, it had been removed: https://github.com/koalaman/shellcheck/issues/298.
I am using 0.4.4.
We could maybe have it for older versions, but then it's probably better to add it to your config manually (since it is not supported upstream / for current versions)?!

blueyed:
As for the shebang check etc: this should a) not be duplicated, and b) is not necessary in my opinion: "sh" can be anything, but "zsh" is specific already.
See https://github.com/neomake/neomake/pull/513.
I've  also fixed the &efm: do you have a space after the filename?
Then we should support both formats.

Issue #424, opened 2016-05-17
Variable type mismatch error with custom maker and vim-jsx


wordofchristian:
I'm getting this error whenever I try to create a custom maker. I've tried following the advanced example from the README.

```
let g:neomake_javascript_jshint_maker = {
    \ 'args': ['--verbose'],
    \ 'errorformat': '%A%f: line %l\, col %v\, %m \(%t%*\d\)',
    \ }
let g:neomake_javascript_enabled_makers = ['jshint']
```

when I run `Neomake jshint` in a javascript file I get the following error:

```
Error detected while processing function neomake#Make..<SNR>154_Make..neomake#GetMaker:
line   14:
E706: Variable type mismatch for: maker
```

## The problem

I think the problem is related to the https://github.com/mxw/vim-jsx plugin. It sets the filetype to `javascript.jsx`. If I change the filetype back to `javascript` the maker works fine. 

I would assume that I need to make a maker for the filetype `javascript.jsx` but I don't know how to set an option with a period in the name.

Is it possible to create a custom maker that supports both filetypes?

blueyed:
Thanks for the report.
Please test https://github.com/neomake/neomake/pull/425.

PR #425, opened 2016-05-17
Fix E706 with multiple filetypes

blueyed:
Fixes https://github.com/neomake/neomake/issues/424.


sbdchd:
Just a heads up, variables default to the local scope inside functions, so you do not need to add `l:` to variable names.

See the docs [here](https://neovim.io/doc/user/eval.html#local-variable)

blueyed:
@sbdchd 
Yeah, I know.. I just gave in to the linting errors (vint), but I agree that we should keep the existing style.

blueyed:
Amended.  Please review again.

PR #429, opened 2016-05-21
Add support for fish

terlar:
This adds the command fish with error format accordingly.

I find that this is a bit buggy sometimes, e.g. sometimes when I run it, it will only parse part of the error message as the message, but with several consecutive runs it will eventually get the full message. It is kind of like it returns too early.

Is there something weird with my pattern or is this a known issue?

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,9 @@ D:
 Erlang:
 - erlc
 
+fish:
+- fish
+
 Go:
 - go
 - golint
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks!
Do you have an example file where it fails sometimes?

Apart from that: please rebase (and add the filetype sorted by name to the README)

terlar:
A file like this:
```fish
function hello_world
    if true
    end
```

Results in:
```
|| Missing end 
|| to balan
|| ce this
/home/terje/badfile.fish|1 col 26 error| function definition
```

But when saving again, works with:
```
/home/terje/badfile.fish|1 col 26 error| Missing end to balance this function definition
```

It randomly fails reading the whole output.

blueyed:
Thanks!

Merged and fixed.  You need `buffer_output = 1`.

PR #430, opened 2016-05-21
Fix shebang parsing in neomake#makers#ft#sh#sh

blueyed:
`#! /usr/bin/env bash` is apparently a valid/used shebang (yes, with a
space even).

Followup to #362.


blueyed:
@paretje 
What do you think?

blueyed:
Without this you'll get something like:

> '/usr/bin/env: invalid option -- ''n'''

PR #435, opened 2016-05-22
vim: add vimlint maker

blueyed:
This assumes that you have some wrapper script like this:

```
% cat ~/bin/vimlint
#!/bin/sh
~/Vcs/vim-vimlint/bin/vimlint.sh -l ~/Vcs/vim-vimlint -p ~/Vcs/vim-vimlparser "$@"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,19 @@ TypeScript:
 - tsc
 
 Vimscript:
+
 - vint
+- [vimlint](https://github.com/syngan/vim-vimlint)
+
+  It can be installed using npm:
+  [node-vimlint](https://www.npmjs.com/package/vimlint).
+
+  Or you could create a wrapper script ``vimlint`` and add it to your PATH:
+
+  ```sh
+  #!/bin/sh
+  ~/Vcs/vim-vimlint/bin/vimlint.sh -l ~/Vcs/vim-vimlint -p ~/Vcs/vim-vimlparser "$@"
+  ```
 
 YAML:
 - [yamllint](http://yamllint.readthedocs.org/)
--- a/autoload/neomake/makers/ft/vim.vim
+++ b/autoload/neomake/makers/ft/vim.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#vim#EnabledMakers() abort
-    return ['vint']
+    return ['vint', 'vimlint']
 endfunction
 
 function! neomake#makers#ft#vim#vint() abort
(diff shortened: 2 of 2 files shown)


blueyed:
I am not sure about if only vimlint should be used if it's available though: it's a lot slower, but seems to be better / useful additionally..

sbdchd:
_Suggestion:_ You could also link to [node-vimlint](https://www.npmjs.com/package/vimlint), which provides a cli for vimlint that can be installed with npm.

blueyed:
@sbdchd 
Makes sense - should probably be the primary source/link even.

blueyed:
@sbdchd 
Updated the README.  Does it sound good like that?

sbdchd:
@blueyed Looks good

blueyed:
@sbdchd 
Thanks!

wsdjeg:
what is the differences of node-vimlint and vim-vimlint? there are so many tools, which one should I install?

sbdchd:
@wsdjeg node-vimlint provides a wrapper script for vim-vimlint, making it easier to install and use vim-vimlit. By itself, vim-vimlint can be a little difficult to use from command line, hence, the bash script.

One either installs node-vimlint, or creates the bash script themselves for vim-vimlint and adds it to their path.

wsdjeg:
I have install node-vimlint, and also add it to PATH, but I can not use it in neomake. is there any other package need I install?

wsdjeg:
here is some config in my vimrc
```vim
    if g:settings.enable_neomake
        call s:add('neomake/neomake',{'on_cmd' : 'Neomake'})
        if s:tap('neomake')
            call s:defind_hooks('neomake')
            augroup Neomake_wsd
                au!
            augroup END
        endif
    else
        call s:add('wsdjeg/syntastic', {'on_event': 'WinEnter'})
        if s:tap('syntastic')
            call s:defind_hooks('syntastic')
        endif
    endif
    call s:add('syngan/vim-vimlint',{'on_ft' : 'vim'})
    let g:syntastic_vimlint_options = {
                \'EVL102': 1 ,
                \'EVL103': 1 ,
                \'EVL205': 1 ,
                \'EVL105': 1 ,
                \}
    call s:add('ynkdir/vim-vimlparser',{'on_ft' : 'vim'})
    call s:add('todesking/vint-syntastic',{'on_ft' : 'vim'})
    "let g:syntastic_vim_checkers = ['vint']

```

sbdchd:
npm should be able to link the `/bin` of node-vimlint to your path by itself, but I don't think adding it to $PATH would change anything.

Can you call node-vimlint, `vimlint`, from the command line?

wsdjeg:
yes, I can , but how does `vimlint` use `vim-vimlparser` and `vim-lint`, how could it be same as `-l ~/Vcs/vim-vimlint -p ~/Vcs/vim-vimlparser`

sbdchd:
I think the way node-vimlint works is by [installing vim-vimlint and vim-vimlparser](https://github.com/pine/node-vimlint/blob/7a685caf6cea8c4041f92ca6a9f1d1f7fc48e4f1/setup.js), and using a [similar setup to the bash script](https://github.com/pine/node-vimlint/blob/7a685caf6cea8c4041f92ca6a9f1d1f7fc48e4f1/lib/vimlint.js#L11-L12).

wsdjeg:
very nice！thanks a lot.

PR #439, opened 2016-05-24
add mypy to python makers

gregwebs:
* If you use `mypy` I don't think you have a need to run the `python` checker.
* don't know what is up with `frosted` being treated special: I treated it like all the others instead.

I tested this locally, it works for me (mypy-lang==0.4.0)

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -168,8 +168,11 @@ function! neomake#makers#ft#python#vulture()
         \ }
 endfunction
 
+" Because this uses --silent-imports it requires mypy >= 0.4
+" It is annoying for new users to use MyPy without --silent-imports
 function! neomake#makers#ft#python#mypy()
     return {
+        \ 'args': ['--silent-imports'],
         \ 'errorformat': '%f:%l: error: %m',
         \ }
 endfunction
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -5,7 +5,14 @@ function! neomake#makers#ft#python#EnabledMakers()
         return s:python_makers
     endif
 
-    let makers = ['python', 'frosted']
+    if executable('mypy')
+        let makers = ['mypy']
+    else
+        let makers = ['python']
+    endif
+    if executable('frosted')
+        call add(makers, 'frosted')
+    endif
 
     if executable('pylama')
         call add(makers, 'pylama')
@@ -158,5 +165,11 @@ endfunction
 function! neomake#makers#ft#python#vulture()
     return {
         \ 'errorformat': '%f:%l: %m',
+        \ }
+endfunction
+
+function! neomake#makers#ft#python#mypy()
+    return {
+        \ 'errorformat': '%f:%l: error: %m',
         \ }
 endfunction


gregwebs:
Looks like the build failure is spurious

blueyed:
> Looks like the build failure is spurious

Should be fixed by now.

Regarding mypy, I do not find it too useful currently - or rather too noisy with regard to errors, like 

> No library stub file for standard library module 'statistics'
> Cannot find module named 'concurrency.api'
> No library stub file for module 'django.conf'

Therefore it should not be enabled by default.

gregwebs:
We can add the option `--silence-imports` as a default argument so that you don't have to put ignores in your files.

But I am curious, why do you have MyPy installed if you aren't using itl?
This is a similar problem if you have multiple linters installed with different opinions on linting.

It seems problematic that this picks up global installations or one is supposed to configure these by overriding global makers settings in a .vimrc. But these settings should probably be project-specific.

gregwebs:
@blueyed I added the `--silent-imports` option by default. Can you give it a try now?

If MyPy is still not useful for you can you uninstall it?

blueyed:
I've tried `--silent-imports` initially already, which did not really help IIRC.

I have it installed globally, and while I could of course configure the Python makers etc, I just don't think that this is good enough to be used by default (if you have it installed).
Even if not installed globally you might just want to experiment with it (in the virtualenv), and Neomake picking it up would be annoying.

gregwebs:
I am still curious why you have MyPy installed if you don't like it? Why not uninstall it?

I am also curious what issue you are running into with MyPy: we have gone over the specific of imports now, but you haven't given any more specifics.

Isn't the argument about Neomake picking up mypy automatically annoying true of linters also (in particular picking up multiple linters)?

blueyed:
> I am still curious why you have MyPy installed if you don't like it? Why not uninstall it?

For scientific reasons.. ;)
I can imagine enabling this on some projects etc.

To give some numbers: with `--silent-imports` I am still getting 89 errors for a Django project's `models.py`. Among them:

> Need type annotation for variable
> Cannot determine type of 'Meta' in base class 'FooModelMixin'
> Name 'abc.ABC' is not defined  (although `abc` is imported?!)
> Callable[[Any], Any] has no attribute "name"  (when assigning a property to a function)

To be fair it also found some issue:
> List literal expected as the second argument to namedtuple()  (a tuple was used, and worked anyway)

Most of them are `Cannot determine type of 'Meta' in base class 'FooModelMixin'`, with several Mixins that use Django's `Meta` class.

> Isn't the argument about Neomake picking up mypy automatically annoying true of linters also (in particular picking up multiple linters)?

Yes, and I'm trying to not make it more annoying, especially for filetypes I care about.. :)

After all, it's great to have support for `mypy` in Neomake of course.  I just don't like it to be enabled by default currently.

gregwebs:
Thanks for the specifics! I have never used django, so I will trust your judgement there. closing this in favor of #450

gregwebs:
If you want to use MyPy, `  # type: ignore` is your friend, but 80 false errors is a lot. I am usually dealing with less than 10 (outside of imports) in a really large file.

blueyed:
Instead of #450, can you please just remove the enabling by default from this PR?!

Just squash it into a single commit and then force-push it to this PR/branch.

If you need help, please let us know.

gregwebs:
The PR is from a different branch. I would prefer to keep my master where it is right now. Please re-open #450. Sorry I didn't close this out yet.

PR #445, opened 2016-05-27
Use the airline naming format for enable check

vasyharan:
Use the naming format that airline uses for consistency.

PR to have airline initialize neomake
https://github.com/vim-airline/vim-airline/pull/1179


blueyed:
Same question as with https://github.com/vim-airline/vim-airline/pull/1179#issuecomment-222294012: why was this closed again?

vasyharan:
I closed the other PR because the code change was unnecessary after fixing
my runtimepath. I didn't see the need for this change without the other
change.

PR #446, opened 2016-05-28
go: Fix golint maker

sebastianmarkow:
Wrongfully broken in 5ac6c80, golint behaves differently in contrast to the core go toolchain


blueyed:
Thanks!

@mediocregopher 
Is this correct?

mediocregopher:
Yup, looks like it. Sorry, I use gometalinter instead of the normal one, I had assumed they behaved the same but I guess not.

blueyed:
Thanks!

PR #451, opened 2016-06-04
Working Perl maker with perl & perlcritic

cfsalguero:
The original perlcritic was not working because it had neomake#makers#ft#**vim**#EnabledMakers instead of neomake#makers#ft#**perl**#EnabledMakers.

Added **perl** maker for syntax checking.


blueyed:
Thanks!

PR #455, opened 2016-06-11
Make neomake#statusline#LocklistCounts accept a winnr arg

blueyed:
This is useful in custom statusline functions.

PR #457, opened 2016-06-12
only reset counts for buf in file_mode

aykamko:
Currently, Neomake _attempts_ to track per-buffer loclist counts but these counts get reset anytime `s:Make` is called, even if the intention is to only process one particular buffer. In other words, if I have N different buffers open but I only want to check the syntax for one particular buffer, Neomake will (mistakenly) clear the loclist counts for all buffers regardless.

This PR fixes the issue by only resetting the counts for a particular buffer if `neomake#Make` is running in `file_mode`.

### DIFF
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -332,8 +332,12 @@ or warnings and the empty string otherwise.
 These functions get the counts of errors by error type for the |location-list|
 and the |quickfix| respectively. The return value is something like this: >
     {'E': 2, 'W': 1, 'x': 5}
-<Where 'E', 'W' and 'x' are error types. Empty error types are ignored for
-now.
+<Where 'E', 'W' and 'x' are error types. Empty error types are ignored for now.
+
+By default, *LoclistCounts* returns the counts for the current buffer (i.e.
+|bufnr(%)|). *LoclistCounts* also takes an optional |buf| argument: passing a
+buffer number will retrieve the counts for a particular buffer, while
+passing the string `'all'` will return a Dictionary of counts for all buffers.
 
 *neomake#ProcessCurrentWindow*
 This is the function that takes the job output and puts it into the loclist or
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -281,10 +281,17 @@ endfunction
 
 function! s:Make(options) abort
     call neomake#signs#DefineSigns()
-    call neomake#statusline#ResetCounts()
 
+    let buf = bufnr('%')
+    let win = winnr()
     let ft = get(a:options, 'ft', '')
     let file_mode = get(a:options, 'file_mode')
+
+    if file_mode
+        call neomake#statusline#ResetCountsForBuf(buf)
+    else
+        call neomake#statusline#ResetCounts()
+    endif
 
     let enabled_makers = get(a:options, 'enabled_makers', [])
     if !len(enabled_makers)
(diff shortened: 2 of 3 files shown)


blueyed:
Thank you for this!

While this likely makes it better, it's still bad: the errors should not be assigned by window numbers, but ids:
if you create a new window to the left, the previous window 1 becomes 2, for example.
This could be done by only handling buffers, or by using window-local variables.

There is some functionality for this in newer Vim versions (`win_getid`, `win_id2tabwin` etc).
Not taking the tab into account is another major issue currently.

blueyed:
I've seen some module providing a layer / wrapper around this, mentioned in some (Neovim?) PR, but cannot find it currently.

As of now, I think this PR is still good, since it makes it at least not worse.. ;)

blueyed:
Found it: https://github.com/junegunn/fzf/pull/559#issuecomment-217404795

aykamko:
> This could be done by only handling buffers, or by using window-local variables.
How about we make `loclist_counts` and `qf_counts` based on buffers only then? Tbh I'm not sure why they were indexed by window originally either.

blueyed:
Don't know either.
Maybe the idea is to have different makers for the same buffer in multiple windows?!

aykamko:
@blueyed Alright, refactored to be based to buffers only. I also added an optional `bufnr` argument to `neomake#statusline#LoclistCounts()` and updated the docs accordingly.

blueyed:
Seem the way to go, thanks!

Can you also add support for "all" like I've done in https://github.com/neomake/neomake/pull/455/files?

aykamko:
@blueyed Done! squashed and ready to go

doc/neomake.txt
```diff
@@ -332,8 +332,12 @@ or warnings and the empty string otherwise.
 These functions get the counts of errors by error type for the |location-list|
 and the |quickfix| respectively. The return value is something like this: >
     {'E': 2, 'W': 1, 'x': 5}
-<Where 'E', 'W' and 'x' are error types. Empty error types are ignored for
-now.
+<Where 'E', 'W' and 'x' are error types. Empty error types are ignored for now.
+
+By default, *LoclistCounts* returns the counts for the current buffer (i.e.
+|bufnr(%)|). *LoclistCounts* also takes an optional |buf| argument: passing a
```
blueyed: `|bufnr(%)|` is not a valid link, and not valid Vimscript.
Let's use `bufnr('%')` there (in backticks, without the link `|`).  I can fix this when merging.

blueyed:
Thanks!
I have not tried it myself yet, and feel currently uncomfortable merging such things without tests.
So this needs to wait for #458, and then it would be nice to have some test for it, of course.

aykamko:
Ok, ping me with a mention when you merge that PR, I can try to write a test. Or you can choose to take it from here, your call!

blueyed:
@aykamko 
#458 should be ready.

We should have some documentation about the tests etc, but that can wait.
I'd appreciate a review of that PR (both the fix for Vim and the test setup).

blueyed:
Merged in fd80cd8, with slight doc changes:

```diff
diff --git a/doc/neomake.txt b/doc/neomake.txt
index a822f24..ded9a6b 100644
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -335,9 +335,9 @@ and the |quickfix| respectively. The return value is something like this: >
 <Where 'E', 'W' and 'x' are error types. Empty error types are ignored for now.
 
 By default, *LoclistCounts* returns the counts for the current buffer (i.e.
-|bufnr(%)|). *LoclistCounts* also takes an optional |buf| argument: passing a
-buffer number will retrieve the counts for a particular buffer, while
-passing the string `'all'` will return a Dictionary of counts for all buffers.
+`bufnr("%")`), but you can pass an optional argument: passing a buffer number
+will retrieve the counts for a particular buffer, while passing the string
+`'all'` will return a dictionary of counts for all buffers.
 
 *neomake#ProcessCurrentWindow*
 This is the function that takes the job output and puts it into the loclist or
```

blueyed:
Thanks again!

PR #459, opened 2016-06-14
Fix sh maker: recognize output format without 'line' from dash

blueyed:
This is apparently used on Travis, but makes sense in general - e.g.
probably still for Ubuntu systems?!


blueyed:
Took me quite a while to debug this in #458 on Travis, wondering why the loclist was empty.

PR #465, opened 2016-06-16
Fix javac classpath option

### LINKED ISSUE #462: Java classpaths are ignored when specified as a javac arg
When specifying a classpath for a java project like so:

```let g:neomake_java_javac_args = ['-cp', 'some:classpath:here']```

It appears that Neomake is making the arguments silently vanish [here](https://github.com/neomake/neomake/blob/master/autoload/neomake.vim#L107) because the classpath is failing to expand and this results in an empty string

I've tried escaping the `:` as `\:` but that still seems to fail

Another way I've gotten around this is patching neomake locally by replacing the expansion with something that checks for an empty result and returns the plain string like so 

```
function! s:expandIfValid(arg) abort
        if expand(a:arg) != ""
            return expand(a:arg)
        else
            return a:arg
        endif
endfunction
call map(args, 's:expandIfValid(v:val)')
```

but this is a suboptimal solution. Any help would be appreciated!

wsdjeg:
enable `g:neomake_java_javac_classpath`.
Fix #462

### DIFF
--- a/autoload/neomake/makers/ft/java.vim
+++ b/autoload/neomake/makers/ft/java.vim
@@ -169,23 +169,23 @@ function! neomake#makers#ft#java#javac()
         let javac_opts = extend(javac_opts, ['-d', s:shescape(output_dir)])
     endif
 
-    let javac_classpath = ''
+    let javac_classpath = get(g:, 'neomake_java_javac_classpath', '')
 
-    if s:has_maven && g:neomake_java_javac_autoload_maven_classpath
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

blueyed:
@wsdjeg 
What do you think about https://github.com/neomake/neomake/pull/458?
You could have tests for the Java maker.. :)

wsdjeg:
@blueyed I am not  familiar with vader and makefile. how could I test for it?

PR #470, opened 2016-06-17
Add `mix compile` support to the Elixir maker

cjhowe:
The default Elixir maker doesn't support finding dependencies between files in your own project, so it tends to give you incorrect errors. I created a maker using `mix` that seems to work much better. This should be the default, since that's what most Elixir users would expect.

### DIFF
--- a/autoload/neomake/makers/ft/elixir.vim
+++ b/autoload/neomake/makers/ft/elixir.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#elixir#EnabledMakers()
-    return ['elixir']
+    return ['mix']
 endfunction
 
 function! neomake#makers#ft#elixir#elixir()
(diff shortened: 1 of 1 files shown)


geowa4:
Does that warning format work for you? My output is the following and does not match.
```
warning: unused alias Endpoint
  lib/hello_phoenix.ex:5
```

geowa4:
It looks like the format change is in Elixir 1.3.

blueyed:
@cjhowe 
ping

cjhowe:
Ah, yes, I haven't used this with Elixir 1.3 yet. Should be an easy fix.

blueyed:
@cjhowe 
Looking forward to it then.. :)

mhinz:
@cjhowe Master isn't working for you? https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/elixir.vim#L23-L32

blueyed:
Yes, looks like it is fixed in master already, closing.
Please re-open in case we're missing something here.

geowa4:
Should I make a new issue for the formatting in master being broken?

blueyed:
@geowa4 
sure.  But please consider fixing it yourself and create a PR instead.
See the one from Syntastic: https://github.com/vim-syntastic/syntastic/blob/675f67eea29308537438eb49b3b28a444c1e4689/syntax_checkers/elixir/elixir.vim#L41-L43

Issue #240, opened 2015-12-03
Neomake seems to cause scroll lag


superquest:
When I hold "j" with neomake enabled, the scrolling is a bit choppy and laggy. When I do the same without Neomake, scrolling is much, much better. I'm mostly wondering if this is unexpected behavior ... and how to further debug given that I'm still a vim novice. 

To confirm my suspicion that Neovim was causing the slowness, I followed these directions and held "j" in a 300 line python file -- http://stackoverflow.com/questions/12213597/how-to-see-which-plugins-are-making-vim-slow -- which generated this report: 

```
FUNCTIONS SORTED ON TOTAL TIME
count  total (s)   self (s)  function
  530   0.053772   0.004361  neomake#CursorMoved()
  530   0.029793             <SNR>63_Highlight_Matching_Pair()
  530   0.025060             neomake#EchoCurrentError()
    1   0.025025   0.001555  plug#end()
  530   0.024352             neomake#signs#PlaceVisibleSigns()
  530   0.010298   0.002646  <SNR>82_OnCursorMovedNormalMode()
  530   0.007652             <SNR>82_AllowedToCompleteInCurrentFile()
    1   0.007570   0.000216  <SNR>2_reload()
   24   0.007411   0.003825  <SNR>2_source()
   25   0.002936   0.001140  <SNR>2_add()
    1   0.002113   0.000538  <SNR>2_reorg_rtp()
   25   0.001457   0.001117  <SNR>2_infer_properties()
    1   0.001267   0.000762  <SNR>2_remove_rtp()
    2   0.000970   0.000587  <SNR>11_SynSet()
   68   0.000882   0.000330  <SNR>2_rtp()
   94   0.000800             <SNR>2_dirpath()
    1   0.000353   0.000145  plug#begin()
   73   0.000310             pymode#default()
    2   0.000307   0.000136  gitgutter#highlight#define_highlights()
   48   0.000282             <SNR>2_lines()

FUNCTIONS SORTED ON SELF TIME
count  total (s)   self (s)  function
  530              0.029793  <SNR>63_Highlight_Matching_Pair()
  530              0.025060  neomake#EchoCurrentError()
  530              0.024352  neomake#signs#PlaceVisibleSigns()
  530              0.007652  <SNR>82_AllowedToCompleteInCurrentFile()
  530   0.053772   0.004361  neomake#CursorMoved()
   24   0.007411   0.003825  <SNR>2_source()
[...]

PR #471, opened 2016-06-18
Add `neomake_place_signs_at_once` option to show all signs at once

ivalkeen:
This new option allows to render all of the signs at once, which eliminates scrolling lag described here: https://github.com/neomake/neomake/issues/240

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -133,8 +133,15 @@ function! neomake#signs#PlaceVisibleSigns() abort
         if !has_key(s:sign_queue[type], buf)
             continue
         endif
-        let topline = line('w0')
-        let botline = line('w$')
+
+        if !get(g:, 'neomake_place_signs_at_once', 0)
+            let topline = line('w0')
+            let botline = line('w$')
+        else
+            let topline = line('0')
+            let botline = line('$')
+        endif
+
         for ln in range(topline, botline)
             if has_key(s:sign_queue[type][buf], ln)
                 call neomake#signs#PlaceSign(s:sign_queue[type][buf][ln], type)
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -299,6 +299,11 @@ You can use `neomake#signs#DefineHighlights` to get e.g. the "bg" from
 Shows the output returned by |neomake#statusline#LoclistStatus| in the warning
 section of the vim-airline |statusline|. Defaults to 1.
 
+*g:neomake_place_signs_at_once*
+This setting will tell neomake to place all of the signs in the buffer at
+once, instead of only placing visible signs and lazily placing more signs
+when the cursor is moved. Defaults to 0.
+
 ==============================================================================
 4. Functions                                                 *neomake-functions*
 
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -840,7 +840,9 @@ function! neomake#CursorMoved() abort
     let l:line = line('.')
     if s:last_cursormoved[0] != l:line || s:last_cursormoved[1] != bufnr('%')
         let s:last_cursormoved = [l:line, bufnr('%')]
-        call neomake#signs#PlaceVisibleSigns()
+        if !get(g:, 'neomake_place_signs_at_once', 0)
+            call neomake#signs#PlaceVisibleSigns()
+        endif
         call neomake#EchoCurrentError()
     endif
 endfunction


doc/neomake.txt
```diff
@@ -291,6 +291,11 @@ You can use `neomake#signs#DefineHighlights` to get e.g. the "bg" from
 Shows the output returned by |neomake#statusline#LoclistStatus| in the warning
 section of the vim-airline |statusline|. Defaults to 1.
 
+*g:neomake_place_signs_at_once*
+This setting will tell neomake to place all of the signs in the buffer at
```
blueyed: s/neomake/Neomake/

doc/neomake.txt
```diff
@@ -291,6 +291,11 @@ You can use `neomake#signs#DefineHighlights` to get e.g. the "bg" from
...
+once, instead of only placing visible signs and lazily placing more signs
+when the cursor is moved. Defaults to 0.
+
```
blueyed: It is also triggered manually / during the callback, isn't it?!

autoload/neomake/signs.vim
```diff
@@ -125,8 +125,15 @@ function! neomake#signs#PlaceVisibleSigns() abort
...
         endif
-        let topline = line('w0')
-        let botline = line('w$')
+
+        if !get(g:, 'neomake_place_signs_at_once', 0)
+            let topline = line('w0')
+            let botline = line('w$')
+        else
+            let topline = line('0')
+            let botline = line('$')
```
blueyed: What about switching the blocks, and having the true-case first (removing the `!` there).
Seems to be more readable?!

aykamko: I feel like this should be the default behavior in Neovim.

blueyed: I agree.
@ivalkeen what do you think?

Does an option for this then still makes sense (to get the current behavior)?
I could imagine that it's better performance-wise with a lot of signs, but on the other hand, the lazy placing gets called on every cursor move..

ivalkeen: I agree that this should be the default behaviour. I didn't feel confident enough to change it, so I've just added an option which solves the problem for me. 

In general, I think that the best solution would be doing something like [vim-gitgutter](https://github.com/airblade/vim-gitgutter) is doing, which is always placing all the signs at once, but have a configurable limit of the number of signs to be placed, to prevent vim from being unresponsive if there are too many signs.

aykamko:
@ivalkeen Update on this PR?

autoload/neomake/signs.vim
blueyed: That is also what quickfixsigns is doing.

blueyed:
@aykamko 
What's the plan (in your opinion)?

aykamko:
I think we should `let g:neomake_place_signs_at_once = 1` by default and just merge this PR as is. If we get issues later about neomake making vim slow like [ivalkeen mentioned in his comment above](https://github.com/neomake/neomake/pull/471#discussion_r68504916), then we can consider taking the gitgutter approach then.

bruschill:
Is this waiting on anything in particular before it can be merged?

ivalkeen:
There are some comments to be addressed (I hope to be able to do that soon) and there is still an open question whether this should be default behaviour or not (and how the option should be named).

SereznoKot:
Is this PR alive? Maybe help needed?

blueyed:
@SereznoKot 
Needs some decision.

I'd say to go with placing all signs at once, but only if there are less than 100.  This limit could be made configurable, but I would not add it to the documentation for now.
Same with `g:neomake_place_signs_at_once`: not documenting it makes it easier to change.

It would be nice to have some test for this, but IIRC there are no tests for signs at all yet, so I do not want to enforce it here - but that would be something to help out with.

blueyed:
I guess I'll pick this up after #751, using this method:
 - place signs on the current screen always
 - expand to max 50 signs above and below

If there are then still signs left, those are queued, but otherwise there will also be no autocommand to check the queue all the time.

I'll try to hook it into Neomake's own User autocommand system (`NeomakeFinished`, or re-add `NeomakeMakerFinished`).

Maybe starting with something simpler.. ;)

blueyed:
@ivalkeen 
Please feel free to update this PR to just place all signs at once, and we can get that in already.

blueyed:
@ivalkeen 
Still interested? ;)
I did a lot of refactoring lately, but apparently it is not that bad regarding conflicts.

I'd be "happy" to adopt it also once I come across it again, and the engine itself is behaving properly.

blueyed:
@ivalkeen 
Ping?!

blueyed:
Took a more radical approach in https://github.com/neomake/neomake/pull/1033.

PR #472, opened 2016-06-19
Slim: add slim-lint maker

opodartho:


autoload/neomake/makers/ft/slim.vim
```diff
@@ -0,0 +1,13 @@
...
+
+function! neomake#makers#ft#slim#EnabledMakers()
+    return ['slimlint']
+endfunction
+
+function! neomake#makers#ft#slim#slimlint()
+    return {
+        \ 'exe': 'slim-lint',
+        \ 'args': ['--no-color'],
+        \ 'errorformat': '%f:%l %m'
```
blueyed: There will be no error type with this.

You could use `%t%m` instead of `%m` - depending on the format.

Can you provide some example output?
Are there different types after all?

opodartho: Thank you for the review. I update the errorformat to display correct type now.
now it's '%f:%l [%t] %m'
There are two types of error[E]/warning[W].
example output:
app/views/activities/index.slim:3 [W] TrailingWhitespace: Line contains trailing whitespace
app/views/activities/index.slim:6 [E] Malformed indentation

blueyed:
Thanks!

opodartho:
Thank you for doing a great job on neomake. :smile:

PR #463, opened 2016-06-16
Trigger NeomakeMakerFinished User autocmd on maker exit

blueyed:
This allows to invalidate caches, e.g. for the statusline etc.

PR #480, opened 2016-06-28
Move/refactor NeomakeMakerFinished autocommand

blueyed:
Doing this later provides the hook access to the processed
location/quickfix list.

Followup to #463.
@bounceme
Does this still work for you?

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -502,6 +502,17 @@ function! s:CleanJobinfo(jobinfo) abort
     endif
 endfunction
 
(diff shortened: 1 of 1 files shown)


bounceme:
Seems too, thanks!

blueyed:
The problem now is that this hook is not called anymore if there are no lines registered..

Will have to think about that: we could have an explicit hook for when output is processed (and move the current one back), and then another one for when all jobs (for each `Neomake` call) are finished.

PR #482, opened 2016-06-29
Fix usage of has-patch, which is available only for 7.4.237+

blueyed:
Thanks @mhinz! (https://github.com/mhinz/vim-grepper/commit/955d8932868f0e5d390d33d85cbb267e2637e19d#commitcomment-18064940)


davidosomething:
👍

wsdjeg:
oh. nice

PR #486, opened 2016-07-02
NeomakeFinished autocmd

blueyed:
Followup to #480.

@bounceme still good?

The name/place was changed, and also how context is passed in.


blueyed:
I plan to add some other autocmd for when the quickfix/location list is changed, which would allow for faster feedback in case the makers take a while.

blueyed:
I've added `NeomakeCountsChanged` also.

blueyed:
Fixed `NeomakeFinished` for Vim, now also tested.

blueyed:
I'll merge this soon, last chance to speak up.. :)

PR #494, opened 2016-07-07
Adds support for typescript projects.

TyOverby:
Right now we have a typescript file maker that builds a single file
with tsc.  This commit adds support for "typescript projects" which
run the typescript compiler on every typescript file contained in
a typescript project. (a typescript project is a folder that has
a tsconfig.json file somewhere in the directory hierarchy).

I also made a bugfix in neomake.vim which wasn't handling the
non-existance of bufnumber that could happen when running a maker
without having any files opened in neovim.

How to use: 
`:NeomakeProject typescript_project`

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -512,7 +512,7 @@ function! s:AddExprCallback(maker) abort
             call setqflist(list, 'r')
         endif
     endif
-    if counts_changed
+    if counts_changed && has_key(a:maker, 'bufnr')
         call s:neomake_hook('NeomakeCountsChanged', {
                     \ 'file_mode': a:maker.file_mode,
                     \ 'bufnr': a:maker.bufnr})
--- a/autoload/neomake/makers/typescript_project.vim
+++ b/autoload/neomake/makers/typescript_project.vim
@@ -1,0 +1,13 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#typescript_project#typescript_project()
+    return {
+        \ 'exe': 'tsc',
+        \ 'args': ['--noEmit'],
+        \ 'errorformat':
+            \ '%E%f %#(%l\,%c): error %m,' .
+            \ '%E%f %#(%l\,%c): %m,' .
+            \ '%Eerror %m,' .
+            \ '%C%\s%\+%m'
+        \ }
+endfunction


autoload/neomake.vim
```diff
@@ -512,7 +512,7 @@ function! s:AddExprCallback(maker) abort
             call setqflist(list, 'r')
         endif
     endif
-    if counts_changed
+    if counts_changed && has_key(a:maker, 'bufnr')
         call s:neomake_hook('NeomakeCountsChanged', {
                     \ 'file_mode': a:maker.file_mode,
                     \ 'bufnr': a:maker.bufnr})
```
blueyed: I think the proper fix here would be to not pass in `a:maker.bufnr` to the hook, but still call it (since counts changed).
Would be better of in a separate PR.

TyOverby: Would you mind making the PR?  This is the first vimscript that I've written and I don't know how to implement what you recommended :(

blueyed: Sure!

For writing a test I wonder how to reproduce it.. you wrote:

> I also made a bugfix in neomake.vim which wasn't handling the
non-existance of bufnumber that could happen when running a maker
without having any files opened in neovim.

autoload/neomake/makers/typescript_project.vim
```diff
@@ -0,0 +1,13 @@
...
+    return {
+        \ 'exe': 'tsc',
+        \ 'args': ['--noEmit'],
+        \ 'errorformat':
+            \ '%E%f %#(%l\,%c): error %m,' .
+            \ '%E%f %#(%l\,%c): %m,' .
+            \ '%Eerror %m,' .
+            \ '%C%\s%\+%m'
+        \ }
+endfunction
```
TyOverby: > How to use: 
`:NeomakeProject typescript_project`

The "project" makers are in neomake/makers and the "file" makers are neomake/makers/ft.  I modeled this after the cargo maker.

The error format is the same because tsc spits out the same errors regardless of if it's compiling a single file, or a project.

The reason that `-m commonjs` is missing is because the tsconfig file specifies the module type.

autoload/neomake.vim
TyOverby: Here were my steps: 

1. Open neovim on a folder that has a "project" inside of it.  For example, this could be a rust project, or a typescript project (if you merge this change).
2. Without opening any files, run `NeomakeProject cargo` (for rust) or `Neomake typescript_project` for typescript.
3. Neomake crashes

TyOverby: I rolled back my fix and confirmed that those steps caused a crash.

autoload/neomake/makers/typescript_project.vim
blueyed: Check out https://github.com/neomake/neomake/pull/502 regarding this: it might still not pass through `bufnr`, but then it's a matter of the hook handling it.

blueyed: I see.
Can you update the PR to only include that commit, please?
You basically need to `git rebase -i origin/master`, delete the other commit and then `git push -f` to the branch for this PR.

(I was not aware of project makers - but thought that files below `ft/` would be library functions)

blueyed:
Please see https://github.com/neomake/neomake/pull/731.

Do we still need a project maker then?

blueyed:
Closing due to missing feedback.
Feel free to comment / re-open, of course.

PR #495, opened 2016-07-08
corrected info_sign variable name in documentation

CallumHoward:


davidosomething:
why? it corresponds to the highlight group name `NeomakeInfoSign`

also the form is inconsistent if you want the long version, it should be "InformationSign" since "Error", "Warning" are nouns, not adjectives describing the sign (e.g. `sand storm` being a specific storm type, not `sandy storm` describing a normal storm as sandy)

CallumHoward:
@davidosomething Currently the documentation is inconsistent with the code (I found this out the hard way). I thought it would be quick just to correct the documentation rather than correct the variable names in the code. Correcting the code is probably the better route.

blueyed:
Yes, the code is wrong:

> % autoload/neomake/signs.vim
201:    elseif exists('g:neomake_informational_sign')
202:        call extend(opts, g:neomake_informational_sign)

Please change it there in your PR, please.

PR #496, opened 2016-07-08
use only spaces in doc

davidosomething:
found some hard tabs


blueyed:
You missed the first line.. :)
Please add it to this PR, and I would suggest removing the "Last change: Jan 22 2016" there.

blueyed:
Thanks! :cherries:

PR #497, opened 2016-07-08
check for +signs

davidosomething:
just added guards wherever `sign ` is called in case someone is running (n)vim without `+signs`
I sometimes use neomake on regular vim (which is supported by this plugin according to its docs) on older systems that are distributed with vim 7.4 compiled `-signs`

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -65,6 +65,10 @@ endfunction
 
 " type may be either 'file' or 'project'
 function! neomake#signs#PlaceSign(entry, type) abort
+    if !has('signs')
+        return
+    endif
+
     if a:entry.type ==? 'W'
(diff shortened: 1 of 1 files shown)


autoload/neomake/signs.vim
```diff
@@ -245,6 +263,10 @@ endfunction
 
 let s:signs_defined = 0
```
blueyed: What about (just) setting this to `1` in case `signs` is not available?
(yes, it's hacky, but just an idea)

blueyed:
LGTM in general, but I dislike that there are so many guards (extra function calls) needed.

davidosomething:
decided against that in case someone uses any of these these autoload fns in their own scripts/maker definitions in some way. that way their script or maker still works, but they get sign support if its available. if the functions were script-local s: i would have just done one check

PR #483, opened 2016-06-29
introduce buffer specific b:neomake_FT_enabled_makers

davidosomething:
Example use case:

- Project has blog posts in markdown format that need to be linted using `prose` linter
- Project has a `readme.md` that should not be `prose` linted
- Set `b:neomake_markdown_enabled_makers = [ 'prose', 'markdownlint' ]` on the blog post buffer, set `b:neomake_markdwon_enabled_makers = [ 'markdownlint' ]` on the readme.md buffer

PR #503, opened 2016-07-13
Followup to #483: Add buffer specific b:neomake_FT_enabled_makers

blueyed:
Followup to https://github.com/neomake/neomake/pull/483.

/cc @davidosomething


davidosomething:
It works! 👍

PR #518, opened 2016-07-21
fixed multi-line bug on hdevtools

yiblet:
hdevtools had the same bug that we fixed in ghcmod. Currently, hdevtools can output things that will cause the error prompt to take up multiple lines. This small 2 line change fixes that.

### DIFF
--- a/autoload/neomake/makers/ft/haskell.vim
+++ b/autoload/neomake/makers/ft/haskell.vim
@@ -4,9 +4,11 @@ function! neomake#makers#ft#haskell#EnabledMakers()
 endfunction
 
 function! neomake#makers#ft#haskell#hdevtools()
+    let mapexpr = 'substitute(v:val, "\n", "", "g")'
     return {
         \ 'exe': 'hdevtools',
         \ 'args': ['check', '-g-Wall'],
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks for this.
Unfortunately the merge has conflicts by now.  Can you rebase it, please?

yiblet:
alright, will do

yiblet:
So I've looked into my solution, and realized that it didn't work. The problem I was solving is that ghc-mod, hdevtools, and possibly some other haskell makers could sometimes put out messages that ask for the user to "press enter". The WideMessage function in autoload/neomake/util.vim is supposed to catch this problem and truncate the message so that it would fit the window, but for some reason it doesn't. 

image

saaguero:
@yiblet do you have a sample haskell in which this issue can be reproduced?

yiblet:
The picture in my example used to be replicable for the commit of neomake that I was running on. Recently, it's come to my attention that, that example isn't replicable when I'm up to date with the master branch.  I think that someone else might have gotten to this problem, in the time after a pointed out this issue.

blueyed:
Related to the buffer_output changes then likely.
Thanks for reporting back.

PR #521, opened 2016-07-26
Add maker file for clippy

michael-younkin:
`clippy` is a tool for Rust which lints for possible style issues in addition to the usual type and syntax errors detected by the rust compiler. This pull request adds support for `cargo clippy`, which has the same output format as `cargo build`. Note: `clippy` has to be installed with `cargo install clippy` for this to work.

Questions:
 - Is it necessary to specify `exe`?
 - Is there a way to give an error message when `clippy` isn't run from the working directory of a project?

### DIFF
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -1,0 +1,12 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#clippy#clippy()
+    return {
+        \ 'exe': 'cargo',
+        \ 'args': ['clippy'],
+        \ 'errorformat':
+            \   '%-Z%f:%l,' .
+            \   '%+C %s,' .
+            \   '%A%f:%l:%c: %*\d:%*\d\ %t%*[^:]: %m,',
+        \ }
+endfunction


blueyed:
Thanks!

> Is it necessary to specify `exe`?

Yes, since it is different from the function name.

> Is there a way to give an error message when `clippy` isn't run from the working directory of a project?

What do you mean?
Is it necessary to run this from the project root dir?
There is a way to set `cwd` for a maker, but currently this cannot be a callback (to get the root dir for example) AFAIK.

antoyo:
Thanks for your work.
I don't think we should merge this as is, because [`clippy` only runs on Rust nightly](https://github.com/Manishearth/rust-clippy#usage) (it is even mentioned that it only runs on the latest nigthly) and there is a new [error format for Rust nightly](https://github.com/neomake/neomake/issues/549), so that this pull request does not work with Rust nigthly.

michael-younkin:
@blueyed Yes, it is necessary to run this command from the project root; if you don't, it gives you a strange error message. I imagine there are other build tools with the same problem though, so is there any standard way of dealing with it?

@antoyo
Yes, the changing error message format is a problem. :/

But to clarify your concern, is it A) that clippy only runs when you have nightly rust installed, is it B) that clippy isn't at 1.0 yet, or is it C) that they decided to change the compiler output format? Because I don't think A) and B) are good reasons not to merge, but C) might be (at least until the output format is stabilized).

antoyo:
@michael-younkin My concern is C).
Since `clippy` only supports the latest nightly and the latest nightly have a new error format, I fear that this pull request wont work in a few days/weeks.

michael-younkin:
Agreed. I need to investigate more because they were talking about only turning on the new error format when they are connected to a TTY since it's really only helpful for people using the compiler outside of an editor. I also need to find out what clippy is planning to do about the new format (ie. will they update theirs, use the old format, or use some other format).

But in the meantime it's probably best not to merge this because otherwise we'll have to keep updating it until they decide what format to use, which could take some time. :/

antoyo:
I did not know about showing the old format outside of `TTY`, but, in any case, it does not work anymore for me with `neomake` and this pull request with the latest nightly.

michael-younkin:
Yep, I believe it's that's because they just turned it on. Anyway, I closed the pull request and will open a new one once they make up their minds since that could take months and they could keep changing the format in the meantime.

antoyo:
I came up with this new `errorformat`:

```vim
let g:neomake_rust_clippy_maker = {
    \ 'exe': 'cargo',
    \ 'args': ['clippy'],
    \ 'errorformat':
        \   '%-Gerror: Could not compile%.%#,'
        \.  '%-Gerror: aborting due to%.%#,'
        \.  '%Eerror: %m,%C\\s%#--> %f:%l:%c,'
        \.  '%Eerror[%.%#]: %m,%C\\s%#--> %f:%l:%c,'
        \.  '%I\\s%#= help: %m,'
        \.  '%I\\s%#= note: %m,'
        \,
    \ 'append_file': 0,
    \ }

" Use cargo clippy by default
let g:neomake_rust_enabled_makers = ['clippy']
```
I believe it is not complete, but it works for some errors.

If anyone finds an error not covered by this `errorformat`, please tell me.

PR #522, opened 2016-07-27
improve Haskell makers

bonds:
1. only enable makers that are available
1. use Stack to run makers, when Stack is available
1. add a maker for Liquid Haskell

### DIFF
--- a/autoload/neomake/makers/ft/haskell.vim
+++ b/autoload/neomake/makers/ft/haskell.vim
@@ -1,12 +1,51 @@
 
+function! neomake#makers#ft#haskell#MakerAvailable(command)
+    " stack may be able to find a maker binary that's not on the normal path
+    " so check for that first
+    if executable('stack')
+        " run the maker command using stack to see whether stack can find it
+        " use the help flag to run the maker command without doing anything
+        if system('stack exec -- ' . a:command . ' --help > /dev/null 2>&1; echo $?') == 0
+            return 1
+        else " if stack cannot find the maker command, its not available anywhere
+            return 0
+        endif
+    elseif executable(a:command) " stack isn't available, so check for the binary directly
+        return 1
+    else
+        return 0
+    endif
+endfunction
+
 function! neomake#makers#ft#haskell#EnabledMakers()
-    return ['ghcmod', 'hdevtools', 'hlint']
+    let commands = ['ghc-mod', 'hdevtools', 'hlint', 'liquid']
+    let makers = []
+    for command in commands
+        if neomake#makers#ft#haskell#MakerAvailable(command)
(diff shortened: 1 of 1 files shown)


bonds:
@benekastah @nathyong @expipiplus1 @etcinit 

y'all have contributed to the Haskell makers before, if you are interested and inclined, your feedback on this change would be much appreciated!

expipiplus1:
There seems to be a bit of duplication in that file around finding the stack executable and using it. I don't know if vimscript makes this easy to abstract over.

I don't use Stack, but as long as it doesn't interfere with running these things under Nix then I'm happy.

Liquid haskell maker looks good!

bonds:
Ah, you noticed that. ;) Ok, DRYed it up, please take another look.

autoload/neomake/makers/ft/haskell.vim
```diff
@@ -1,10 +1,23 @@
 
 function! neomake#makers#ft#haskell#EnabledMakers()
-    return ['ghcmod', 'hdevtools', 'hlint']
+    let makers = []
+    if executable('ghc-mod')   | call add(makers, 'ghcmod')    | endif
+    if executable('hdevtools') | call add(makers, 'hdevtools') | endif
+    if executable('hlint')     | call add(makers, 'hlint')     | endif
+    if executable('liquid')    | call add(makers, 'liquid')    | endif
+    return makers
```
blueyed: Just return a list here, the `executable` check gets done by Neomake then.
It makes sense though to check it yourself, is some meta-checker should be preferred over others and/or two makers are basically the same etc.

blueyed: It seems like only `stack` should be enabled if it's there, otherwise the list?

bonds: Hmm, well, both ````stack```` and the executable, e.g. ````hlint```` are needed for running the ````hlint```` command via ````stack````. So, my idea was to check for both. But, in fact, one could have an executable that *only* works when run via ````stack````, because ````stack```` adds search paths to look for the executable in, i.e. the current Haskell project's binary directory. So...a better solution would be to run a test like ````stack exec hlint -V```` to see if stack is installed AND whether it can find ````hlint````. I'll have to think about this a bit.

bonds:
This is a bit better. I added comments to explain what its doing too.

blueyed:
@bonds 
Is this good to go?
Can I squash it into one commit?  You could also do so and force-push to this branch, in case you want to massage the commit message.

bonds:
@blueyed good to go and squashed with updated commit message.

blueyed:
Thanks!

PR #527, opened 2016-07-30
[RFC] Write first version of the Racket maker.

HiPhish:
The maker in its current form has a number of issues:
- ~The error message is prefixed with the name of the maker (`raco: bar: unbound idetifier ...`)~
- The location list seems to do some sort of pattern matching as well
- ~Pressing <CR> in the location list on the error message does not open the error, instead an error is thrown: `E924: Current window was closed`~
- If there are no errors the location list will be populated with the output of the expanded file. What I want is for the location list to either contain nothing if there is no error or contain everything if there is an error.
- ~The location list does not open automatically on error like it does for example for Python (I don't have any special settings for Python files).~

I would appreciate help in getting the details right before merging. For comparison, here is what `raco expand` prints on the terminal:

```
test.rkt:3:1: bar: unbound identifier in module
  in: bar
  context...:
   /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt:34:15: loop
   /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt:10:2: show-program
   /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt: [running body]
   /usr/local/Cellar/racket/6.6/share/racket/collects/raco/raco.rkt: [running body]
   /usr/local/Cellar/racket/6.6/share/racket/collects/raco/main.rkt: [running body]
```

And the defective file:

```
#lang racket
(bar "asdf")
```

Nothing special, just using `bar` instead of `car`, and the identifier `bar` is never defined anywhere.

### DIFF
--- a/autoload/neomake/makers/ft/racket.vim
+++ b/autoload/neomake/makers/ft/racket.vim
@@ -1,0 +1,32 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#racket#EnabledMakers()
+    return ['raco']
+endfunction
+
+" This is the same form of syntax-checking used by DrRacket as well. The
+" downside is that it will only catch the first error, but none of the
+" subsequent ones. This is due to how evaluation in Racket works.
+"
+" About the error format: raco will print the first line as
+"     <file>:<line>:<column> <message>
+" Every successive line will be indented by two spaces:
+"       in: <keyword>
+"       context...:
+"       <file>:<line>:<column>: <keyword>
+" The last pattern will be repeated as often as necessary. Example:
+"     foo.rkt:4:1: dfine: unbound identifier in modulemessage
+"       in: dfine
+"       context...:
+"        /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt:34:15: loop
+"        /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt:10:2: show-program
+"        /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt: [running body]
+"        /usr/local/Cellar/minimal-racket/6.6/share/racket/collects/raco/raco.rkt: [running body]
+"        /usr/local/Cellar/minimal-racket/6.6/share/racket/collects/raco/main.rkt: [running body]
+function! neomake#makers#ft#racket#raco()
+    return {
+        \ 'exe': 'raco',
+        \ 'args': ['expand'],
+        \ 'errorformat': '%-G %.%#,%E%f:%l:%c: %m'
+    \ }
+endfunction


blueyed:
@HiPhish 
What is the state of this?

HiPhish:
@blueyed I had completely forgotten about this one. Give me a week or so to come back to it, then I'll report what the status is.

HiPhish:
OK, I have refreshed my memory. Some of the points seem to be solved, but some remain.

> The location list seems to do some sort of pattern matching as well
The error message contains paths, those are matched by Vim and highlighted accordingly. Here is an example. The file in question:
```racket
#lang racket

(foo "bar")
```
The location list contains:
```qf
test.rkt|3 col 1 error| foo: unbound identifier in module
||   in: foo
||   context...:
/usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt|34 col 15 error| loop
/usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt|10 col 2 error| show-program
||    /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt: [running body]
||    /usr/local/Cellar/minimal-racket/6.6/share/racket/collects/raco/raco.rkt: [running body]
||    /usr/local/Cellar/minimal-racket/6.6/share/racket/collects/raco/main.rkt: [running body]
```
Those are paths to files specific to the Racket implementation. I think it's best to just leave the message as is and trust the user not to mess with these files.

> If there are no errors the location list will be populated with the output of the expanded file.

This is because of what `raco expand` does, i.e. expand all macros and print the expanded version. I could have added `'remove_invalid_entries': 1` to the maker, but then the message of actual errors would get mangled because raco will display useful advice on new lines.

HiPhish:
@blueyed OK, I have been using this one for a while and I think it is as good as using raco is going to get. As far as I am concerned this is ready to be merged.

autoload/neomake/makers/ft/racket.vim
```diff
@@ -0,0 +1,17 @@
...
+    return ['raco']
+endfunction
+
+" This is the same form of syntax-checking used by DrRacket as well. The
+" downside is that it will only catch the first error, but none of the
+" subsequent ones. This is due to how evaluation in Racket works.
+function! neomake#makers#ft#racket#raco()
+    return {
+        \ 'exe': 'raco',
+        \ 'args': ['expand', '%:p'],
```
blueyed: `append_file` (default) will append the current file already, so the `%:p` should be removed?!

autoload/neomake/makers/ft/racket.vim
```diff
@@ -0,0 +1,17 @@
...
+        \ 'remove_invalid_entries': 0,
+        \ 'errorformat': '%E%f:%l:%c: %m'
```
blueyed: The errorformat looks rather simple from what you get in your examples.
You might want to check out `:h erroformat` regarding handling of multiline errors.

HiPhish:
@blueyed All done. The errorformat will now only match the first line of the error message, is that OK? Like if I have
```
derp.rkt:4:1: dfine: unbound identifier in module
  in: dfine
  context...:
   /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt:34:15: loop
   /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt:10:2: show-program
   /usr/local/Cellar/racket/6.5/share/racket/pkgs/compiler-lib/compiler/commands/expand.rkt: [running body]
   /usr/local/Cellar/minimal-racket/6.6/share/racket/collects/raco/raco.rkt: [running body]
   /usr/local/Cellar/minimal-racket/6.6/share/racket/collects/raco/main.rkt: [running body]
```
the output will be
```
derp.rkt|4 col 1 error| dfine: unbound identifier in module
```

blueyed:
Makes sense.
Does `'%-G  %.%#,%E%f:%l:%c: %m'` also work?

HiPhish:
@blueyed Yes, it works as well. Should I use that instead? I'm completely new to Vim's errorformat, it's all magic to me.

blueyed:
@HiPhish 
Seems to be better in my opinion.

HiPhish:
@blueyed OK, got it. Anything else?

blueyed:
Do you need the `'remove_invalid_entries': 0,`?
What for?

Otherwise, fine with me.

HiPhish:
@blueyed I have no idea what I did that for. I think I had copied another maker and then changed it to suit Racket, so that might be why it was in there. We don't need it.

blueyed:
Thanks!

PR #536, opened 2016-08-04
Add dogma to elixir markers

fazibear:


autoload/neomake/makers/ft/elixir.vim
```diff
@@ -19,3 +19,11 @@ function! neomake#makers#ft#elixir#credo()
       \ 'errorformat': '[%t] %. %f:%l:%c %m'
       \ }
 endfunction
+
+function! neomake#markers#ft#elixir#dogma()
+    return {
+      \ 'exe': 'mix',
+      \ 'args': ['dogma', '%:p', '--format=flycheck'],
+      \ 'errorformat': '%f:%l:%c: %.: %m'
```
blueyed: If all are errors, please add `%E` to the beginning of the errorformat.

Also, please add it to the README.

blueyed: `%E` means to mark this line as an error.
So you would use `'errorformat': '%E%f:%l:%c: %.: %m'`.

There are several issues regarding to makers not defining the type correctly, and we might use "error" by default because of this.  But anyway, I think it's better to be more explicit.

fazibear:
Done!

README.md
```diff
@@ -99,6 +99,10 @@ CSS:
 D:
 - dmd
 
+Elixir:
+- credo
+- dogma
```
blueyed: You should add " (not enabled by default)" here, if it's really intentional.
Otherwise add it to `neomake#makers#ft#elixir#EnabledMakers`.

fazibear: Good point. Done.

blueyed:
Thanks!

The other makers are missing the "%E", too.

PR #550, opened 2016-08-10
Fix checking if checkstyle xml has already been set

Asdas91:
Right now setting `g:neomake_java_checkstyle_xml` to use a custom checkstyle configuration xml does not work. It is because someone added _g:_ also to the string that is searched for in the defined global variables by mistake.

I noticed that when I was trying to use neomake with checkstyle so I propose this simple fix.


blueyed:
Thanks!

I've verified that there are no other things like that using `ag 'get.*g:.*g:'`.

PR #555, opened 2016-08-13
[elixir] Fix typo in function namespace

tlarrieu:
This pull request fixes the following error : 

```
Error detected while processing [...]/neomake/autoload/neomake/makers/ft/elixir.vim:
line   29:
E746: Function name does not match script file name: neomake#markers#ft#elixir#dogma
Error detected while processing function neomake#Make:
line    3:
E171: Missing :endif
```


blueyed:
Thanks!
We should really have some test that actually sources and runs all makers somehow.

But for that it would be best to also have input that is targeted at them.  But we could have tests without that already.

Issue #507, opened 2016-07-15
Disable a checker... 


kierun:
I have some pieces of Python code that uses Twisted and others that use glib/gobject. Using pylint with those is a recipe for [yak shaving](http://www.urbandictionary.com/define.php?term=yak%20shaving) due to way too many false positives.

For example, `from glib import MainLoop` will cause a "module glib has no member MainLoop"... Well, actually, it does have one.

Is there a way to ignore pylint even if it is installed?

davidosomething:
You can disable pylint all around by manually setting what makers you want:

```
let g:neomake_python_enabled_makers = ['pep8', 'frosted']
```
that will only use pep8 and frosted

When this PR is merged (scheduled for this weekend) you will be able to do it per-buffer, so you could check for files in `/glib/` or something if you needed: https://github.com/neomake/neomake/pull/503

kierun:
Thank you! This works superbly.

Might I humbly suggest this was documented somewhere in the readme or FAQ?

blueyed:
@kierun
Please consider creating a PR for it. The project lives from contributions.

PR #559, opened 2016-08-22
Add: how to select specific linters.

kierun:
Just some documentation so that new users learn of the `neomake_python_enabled_makers` command. See [this issue comment](https://github.com/neomake/neomake/issues/507#issuecomment-241227794).

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,3 +245,17 @@ for all supported makers.
 If you find this plugin useful, please contribute your maker recipes to the
 repository! Check out `autoload/neomake/makers/**/*.vim` to see how that is
 currently done.
+
+## How to select some linters
+
+Sometimes, it might not be appropriate to use some linter or other. In such
+cases, one can enable only the linters that one wants using:
+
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks.
It is in the help already though (`:h g:neomake_<ft>_enabled_makers`): https://github.com/neomake/neomake/blob/master/doc/neomake.txt#L199-L207
It also mentions that it can be a buffer-local var, which is missing here.

I can see the benefit of having it in the README, but maybe we should rather link to the Vim help file, to not cause fragmentation of the documentation?!

kierun:
Up to you really, I do not mind. To be honest, I forgot to look in the vim help (oops) so thought that it merited to be in the readme.

blueyed:
Improved the doc/README in #567, adding a FAQ section (although only as a reference for this issue).
Do you like it?

kierun:
Works for me! Many thanks indeed.

PR #564, opened 2016-08-30
Implement  #361: highlight columns

mhartington:
This is a rebased version of #361 that adds highlights to not only the side column, but the line column as well.

TODO:
 - [ ] check if supported (for older Vim)
 - [ ] tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ Where `{ language }` is replaced with the name of the language, and `{ makername
 does not follow this convention, neomake will not be able to see it, and you
 will get an error message like `{ makername } not found`.
 
+Explanation for the strings making up the errorformat can be found by typing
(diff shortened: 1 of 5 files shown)


bounceme:
It might be good to have a fallback for matchaddpos which was added vim 7.4.330

blueyed:
Can you create the PR from a branch, please? - so only relevant changes show up here?  (I guess you have to create a new PR for that then).  Or reset your master to only include this PR until it is merged.

blueyed:
> It might be good to have a fallback for matchaddpos, which was added vim 7.4.330

I'd say only if it's easy: otherwise you just cannot use it.. but this should be checked then (`has('*matchaddpos')`).

mhartington:
Closing as #361 is updated again. Sorry for the noise!

Issue #499, opened 2016-07-08
Better default for TSC maker 


mhartington:
By default, the arguments for TSC are pretty limited. 
https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/typescript.vim#L7

Only passing in a few options and not using a project tsconfig.json

What would be great is if it could be configured to use a projects tsconfig.json, and pass that with the `--project` flag.

From the command line you would do

```
tsc -p ./tsconfig.json
```

But this introduces another problem. Currently, all makers use the path of the current file when running the command.  But TSC needs an absolute path to the tsconfig.json.

So while `~/projects/main/app.ts` may work for the maker, it won't detect the tsconfig in `~/projects/`

Is it possible to set neomake up like this and configure the make correctly?

blueyed:
You could use something like https://github.com/dbakker/vim-projectroot, which provides functions to get the project root and then configure the maker to use this in your vimrc (via overriding `args`).

OTOH `tsc` could get a feature to look upwards from the file to be linted for a config file, which is typically the case with linters.

mhartington:
heh, it's been brought up before to give `tsc` a feature like this, but no luck yet. 
I'll see if I cant get something together without vim-projectroot and try to make a PR.

mhartington:
Is there an example of using a function for `args`? According to the docs it should be able to accept one but I keep getting a warning that it's expecting a list. 

```
The 'exe' and 'args' entries can be a function (callback) or a dictionary with
an expected entry 'fn' (which has to be a function).
The function has to return a string, and receives no arguments.
```

```
  function! TSCPWD()
    return '-p ' . getcwd()
  endfunction
  function! neomake#makers#ft#typescript#tsc()
      return {
          \ 'args': (TSCPWD()),
          \ 'errorformat':
              \ '%E%f %#(%l\,%c): error %m,' .
              \ '%E%f %#(%l\,%c): %m,' .
              \ '%Eerror %m,' .
              \ '%C%\s%\+%m'
          \ }
  endfunction
```

davidosomething:
`'args': function('TSCPWD()')`

change TSCPWD to return a VimL list (e.g. `return [ 'p' ]` )

blueyed:
We should probably add an example to the docs..

mhartington:
Hmm, looks like it's not currently possible. Since the commands end up being `tsc <%p> <args>`, it can't accept `--project` and a file at  the same time 😞

mhartington:
[...]

PR #565, opened 2016-08-31
Make typescript maker respect tsconfig.json settings

chrisnicola:
Applies the suggested fix from #499

### DIFF
--- a/autoload/neomake/makers/ft/typescript.vim
+++ b/autoload/neomake/makers/ft/typescript.vim
@@ -7,8 +7,9 @@ endfunction
 function! neomake#makers#ft#typescript#tsc()
     return {
         \ 'args': [
-            \ '-m', 'commonjs', '--noEmit'
+            \ '--project', getcwd(), '--noEmit'
         \ ],
+        \ 'append_file': 0,
         \ 'errorformat':
             \ '%E%f %#(%l\,%c): error %m,' .
             \ '%E%f %#(%l\,%c): %m,' .


autoload/neomake/makers/ft/typescript.vim
```diff
@@ -7,8 +7,9 @@ endfunction
 function! neomake#makers#ft#typescript#tsc()
     return {
         \ 'args': [
-            \ '-m', 'commonjs', '--noEmit'
+            \ '--project', getcwd(), '--noEmit'
```
blueyed: The `getcwd()` looks weird..  it should use the buffer's dir or something probably?
`getcwd()` might return e.g. a local cwd.

Do we need something to get a project's root for this?  (see e.g. https://github.com/dbakker/vim-projectroot)

blueyed: If the dir of the file would be enough I think something like `%p:h` could be used.

chrisnicola: No TSC will traverse up the directory tree until it finds the first tsconfig.json. I'll try what you suggested

blueyed: See the go makers IIRC.

blueyed: Also: https://github.com/neomake/neomake/pull/574/files#diff-5dd22f4fc6d3cb6db8065eaddd1aa749R10

blueyed:
Shouldn't this become a project/directory checker then, when it compiles/checks the whole project?!

blueyed:
https://github.com/Microsoft/TypeScript/issues/5858#issuecomment-255185701 recommends:

> your best bet is to specify a new tsconfig.json extends the exiting one, and add your file to it.

Would it be possible to create such a config on the fly?  (assuming that we have a function to find / configure the main `tsconfig.json`)

blueyed:
See https://github.com/neomake/neomake/pull/731.

https://github.com/neomake/neomake/pull/494 also is about a project maker, but would be obsolete / the same.

gsanches:
I work with a general workspace. My solution was the code below:

```viml
function! neomake#makers#ft#typescript#tsc()
      return {
                  \ 'args': ['--project', getcwd() . '/' . substitute(system('echo ' . expand('%:.') . ' | cut -d "/" -f1'), '\n\+$', '', ''), '--noEmit'],
                  \ 'append_file': 0,
                  \ 'errorformat':
                  \   '%E%f %#(%l\,%c): error %m,' .
                  \   '%E%f %#(%l\,%c): %m,' .
                  \   '%Eerror %m,' .
                  \   '%C%\s%\+%m'
                  \ }
endfunction
```

I had to trim the system output.

PR #568, opened 2016-09-01
Add :NeomakeInfo and GitHub issue template

blueyed:
This is just a first shot, and should be mentioned in the docs / issue
template.

TODO:
 - [ ] more info?  what?
 - [x] test

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,31 @@
+### Intro
+
+Please adhere to this template for bug reports.
+For feature requests and if you know what you are doing you can do it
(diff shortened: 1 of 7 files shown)


davidosomething:
Would you consider putting the Info into neovim's `:CheckHealth` ?
e.g. https://github.com/Shougo/deoplete.nvim/blob/46af24d503660d761b229b2b7743911383dc7261/autoload/health/deoplete.vim

blueyed:
> Would you consider putting the Info into neovim's `:CheckHealth` ?

It is not really diagnostic as of now, but why not?

Let's do it in a separate PR, even it it needs refactoring probably, to improve the issues we're receiving.

PR #575, opened 2016-09-02
Vim asynchronous support

saaguero:
This pull request addresses https://github.com/neomake/neomake/issues/319

It was tested successfully in:

- Linux (Fedora 23) with Vim 7.4.2260
- Windows 8 x64 both with gVim as well as plain-old Vim (inside ConEmu) 7.4.2305 (precompiled from https://tuxproject.de/projects/vim/)

With Python & Vim makers.

Can someone test this on Mac OSX?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,18 +2,17 @@
 
 # Neomake
 
-Neomake is a plugin for asynchronous `:make` using
-[Neovim's](http://neovim.org/) job-control functionality. It is inspired by
-the excellent Vim plugins [Syntastic](https://github.com/scrooloose/syntastic)
-and [dispatch.vim](https://github.com/tpope/vim-dispatch).
+A plugin for asynchronous `:make` using [Neovim's](http://neovim.org/) or [Vim's](http://vim.org/)
+job-control functionality. It is inspired by the excellent vim plugins
+[Syntastic](https://github.com/scrooloose/syntastic) and
+[Dispatch](https://github.com/tpope/vim-dispatch).
 
-**This plugin also works in ordinary Vim, but without the asynchronous
-benefits.**
-
-The minimum Neovim version supported by Neomake is `NVIM
-0.0.0-alpha+201503292107` (commit `960b9108c`).
+The minimum Neovim version supported by Neomake is `NVIM 0.0.0-alpha+201503292107` (commit `960b9108c`).
 The minimum Vim version supported by Neomake is 7.4.503 (although if you don't
 use `g:neomake_logfile` older versions will probably work fine as well).
+
+Starting from Vim version 7.4.2260 Neomake takes advantage of Vim's async support.
+
 
 ## How to use (basic)
 
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -16,8 +16,8 @@ let s:need_errors_cleaning = {
     \ }
 
 function! neomake#has_async_support() abort
-    " TODO: add support for Vim's async support (job_start).
-    return has('nvim')
+    return has('nvim') ||
+                \ has('channel') && has('job') && has('patch-7-4-2260')
 endfunction
 
 function! neomake#GetJobs() abort
@@ -33,7 +33,12 @@ endfunction
 
 function! neomake#CancelJob(job_id) abort
     if has_key(s:jobs, a:job_id)
-        call jobstop(a:job_id)
+        call neomake#utils#DebugMessage('Stopping job: ' . a:job_id)
+        if has('nvim')
+            call jobstop(a:job_id)
+        else
+            call job_stop(s:jobs[a:job_id].job)
+        endif
         return 1
     endif
     return 0
@@ -102,12 +107,9 @@ function! s:MakeJob(make_id, maker) abort
         call add(args, '%:p')
     endif
 
-    if neomake#utils#IsRunningWindows()
-        " Don't expand &shellcmdflag argument of cmd.exe
-        call map(args, 'v:val !=? &shellcmdflag ? expand(v:val) : v:val')
-    else
-        call map(args, 'expand(v:val)')
-    endif
+    call neomake#utils#DebugMessage('Arguments before expansion: '. string(args))
+    call neomake#utils#ExpandArgs(args)
+    call neomake#utils#DebugMessage('Arguments after expansion: '. string(args))
 
     if has_key(a:maker, 'cwd')
         let old_wd = getcwd()
@@ -117,39 +119,71 @@ function! s:MakeJob(make_id, maker) abort
 
     try
         let has_args = type(args) == type([])
-        if has('nvim')
-            let argv = [exe]
-            if has_args
-                let argv = argv + args
-            endif
-            call neomake#utils#LoudMessage('Starting: '.join(argv, ' '))
-            let opts = {
-                \ 'on_stdout': function('neomake#MakeHandler'),
-                \ 'on_stderr': function('neomake#MakeHandler'),
-                \ 'on_exit': function('neomake#MakeHandler')
-                \ }
-            let job = jobstart(argv, opts)
-            let jobinfo.start = localtime()
-            let jobinfo.last_register = 0
+        if neomake#has_async_support()
+            if has('nvim')
+                let argv = [exe]
+                if has_args
+                    let argv = argv + args
+                endif
+                call neomake#utils#LoudMessage('Starting: '.join(argv, ' '))
+                let opts = {
+                            \ 'on_stdout': function('neomake#MakeHandler'),
+                            \ 'on_stderr': function('neomake#MakeHandler'),
+                            \ 'on_exit': function('neomake#MakeHandler')
+                            \ }
+                let job = jobstart(argv, opts)
+                let jobinfo.start = localtime()
+                let jobinfo.last_register = 0
 
-            if job == 0
-                throw 'Job table is full or invalid arguments given'
-            elseif job == -1
-                throw 'Non executable given'
-            endif
+                if job == 0
+                    throw 'Job table is full or invalid arguments given'
+                elseif job == -1
+                    throw 'Non executable given'
+                endif
 
-            let jobinfo.id = job
-            let s:jobs[job] = jobinfo
-            let maker_key = s:GetMakerKey(a:maker)
-            let s:jobs_by_maker[maker_key] = jobinfo
-            call s:AddJobinfoForCurrentWin(jobinfo.id)
-            let r = jobinfo.id
-            if !exists('s:jobids_by_makeid[a:make_id]')
-                let s:jobids_by_makeid[a:make_id] = []
+                let jobinfo.id = job
+                let s:jobs[job] = jobinfo
+                let maker_key = s:GetMakerKey(a:maker)
+                let s:jobs_by_maker[maker_key] = jobinfo
+                call s:AddJobinfoForCurrentWin(jobinfo.id)
+                let r = jobinfo.id
+                if !exists('s:jobids_by_makeid[a:make_id]')
+                    let s:jobids_by_makeid[a:make_id] = []
+                endif
+                call add(s:jobids_by_makeid[a:make_id], jobinfo.id)
+            else
+                if has_args
+                    let program = [exe] + args
+                else
+                    let program = [exe]
+                endif
+
+                call neomake#utils#LoudMessage('Starting: ' . string(program))
+                let job = job_start(program, {
+                            \ 'callback': 'neomake#MakeHandlerVim',
+                            \ 'close_cb': 'neomake#MakeHandlerVimClose'
+                            \ })
+                let jobinfo.job = job
+                let jobinfo.start = localtime()
+                let jobinfo.last_register = 0
+                let jobinfo.data = []
+
+                let job_channel = job_getchannel(job)
+                let jobinfo.id = s:GetChannelId(job_channel)
+
+                let s:jobs[jobinfo.id] = jobinfo
+
+                let maker_key = s:GetMakerKey(a:maker)
+                let s:jobs_by_maker[maker_key] = jobinfo
+                call s:AddJobinfoForCurrentWin(jobinfo.id)
+                let r = jobinfo.id
+                if !exists('s:jobids_by_makeid[a:make_id]')
+                    let s:jobids_by_makeid[a:make_id] = []
+                endif
+                call add(s:jobids_by_makeid[a:make_id], jobinfo.id)
             endif
-            call add(s:jobids_by_makeid[a:make_id], jobinfo.id)
         else
-            " Vim, synchronously.
+            call neomake#utils#DebugMessage('Running synchronously')
             if has_args
                 if neomake#utils#IsRunningWindows()
                     let program = exe.' '.join(map(args, 'v:val'))
@@ -159,6 +193,9 @@ function! s:MakeJob(make_id, maker) abort
             else
                 let program = exe
             endif
+
+            call neomake#utils#LoudMessage('Starting: ' . program)
+
             let jobinfo.id = job_id
             let s:jobs[job_id] = jobinfo
             call s:AddJobinfoForCurrentWin(jobinfo.id)
@@ -426,7 +463,7 @@ function! s:Make(options, ...) abort
             let jobinfo = s:jobs_by_maker[maker_key]
             let jobinfo.maker.next = copy(a:options)
             try
-                call jobstop(jobinfo.id)
+                call neomake#CancelJob(jobinfo.id)
             catch /^Vim\%((\a\+)\)\=:E900/
                 " Ignore invalid job id errors. Happens when the job is done,
                 " but on_exit hasn't been called yet.
@@ -668,6 +705,22 @@ function! s:RegisterJobOutput(jobinfo, lines) abort
     endif
 endfunction
 
+function! s:GetChannelId(channel) abort
+    return ch_info(a:channel)['id']
+endfunction
+
+function! neomake#MakeHandlerVim(channel, line) abort
+    call add(s:jobs[s:GetChannelId(a:channel)].data, a:line)
+endfunction
+
+function! neomake#MakeHandlerVimClose(channel) abort
+    call neomake#utils#DebugMessage('Channel has been closed: ' . a:channel)
+    let channel_id = s:GetChannelId(a:channel)
+    let jobinfo = s:jobs[channel_id]
+    call neomake#MakeHandler(channel_id, jobinfo.data, 'stdout')
+    call neomake#MakeHandler(channel_id, job_info(jobinfo.job)['exitval'], 'exit')
+endfunction
+
 function! neomake#MakeHandler(job_id, data, event_type) abort
     if !has_key(s:jobs, a:job_id)
         return
@@ -708,7 +761,7 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
     elseif a:event_type ==# 'exit'
         " Handle any unfinished lines from stdout/stderr callbacks.
         if has_key(jobinfo, 'lines')
-            if jobinfo.lines[-1] ==# ''
+            if len(jobinfo.lines) && jobinfo.lines[-1] ==# ''
                 call remove(jobinfo.lines, -1)
             endif
             if len(jobinfo.lines)
@@ -732,8 +785,8 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
             endtry
         endif
         call s:CleanJobinfo(jobinfo)
-        if has('nvim')
-            " Only report completion for neovim, since it is asynchronous
+
+        if neomake#has_async_support()
             call neomake#utils#QuietMessage(get(maker, 'name', 'make').
                                           \ ' completed with exit code '.status)
         endif
@@ -760,7 +813,7 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
         endif
 
         " Trigger autocmd if all jobs for a s:Make instance have finished.
-        if has('nvim')
+        if neomake#has_async_support()
             let make_id = -1
             for [k, v] in items(s:jobids_by_makeid)
                 if index(v, a:job_id) != -1
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -202,18 +202,6 @@ function! neomake#utils#GetHighlight(group, what) abort
 endfunction
 
 function! neomake#utils#ExpandArgs(args) abort
-    if neomake#utils#IsRunningWindows()
-        call neomake#utils#ExpandArgsInWindows(a:args)
-    else
-        call neomake#utils#ExpandArgsInLinux(a:args)
-    endif
-endfunction
-
-function! neomake#utils#ExpandArgsInWindows(args) abort
-    " Don't expand &shellcmdflag argument of cmd.exe nor arguments like '"something"'
-    call map(a:args, "(v:val ==? &shellcmdflag || v:val =~? '.*\".*\".*') ? v:val : expand(v:val)")
-endfunction
-
-function! neomake#utils#ExpandArgsInLinux(args) abort
-    call map(a:args, 'expand(v:val)')
+    " Only expand those args that start with \ and a single %
+    call map(a:args, "v:val =~# '\\(^\\\\\\|^%$\\|^%[^%]\\)' ? expand(v:val) : v:val")
 endfunction
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -200,3 +200,20 @@ function! neomake#utils#GetHighlight(group, what) abort
   endif
   return val
 endfunction
+
+function! neomake#utils#ExpandArgs(args) abort
+    if neomake#utils#IsRunningWindows()
+        call neomake#utils#ExpandArgsInWindows(a:args)
+    else
+        call neomake#utils#ExpandArgsInLinux(a:args)
+    endif
+endfunction
+
+function! neomake#utils#ExpandArgsInWindows(args) abort
+    " Don't expand &shellcmdflag argument of cmd.exe nor arguments like '"something"'
+    call map(a:args, "(v:val ==? &shellcmdflag || v:val =~? '.*\".*\".*') ? v:val : expand(v:val)")
+endfunction
+
+function! neomake#utils#ExpandArgsInLinux(args) abort
+    call map(a:args, 'expand(v:val)')
+endfunction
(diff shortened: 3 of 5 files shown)


blueyed:
The build fails because there's a vint issue in one of the changed files, not necessarily from you:

> autoload/neomake.vim:112:58: Use robust operators `=~#` or `=~?` instead of `=~` (see Google VimScript Style Guide (Matching))

autoload/neomake.vim
```diff
@@ -33,7 +33,11 @@ endfunction
 
 function! neomake#CancelJob(job_id) abort
     if has_key(s:jobs, a:job_id)
-        call jobstop(a:job_id)
+        if has('nvim')
+            call jobstop(a:job_id)
+        else
+            call job_stop(s:jobs[a:job_id].job)
+        endif
```
blueyed: Needs some doc adjustment IIRC (because it refers to `jobstop` explicitly).  Please remove that implementation detail then.

autoload/neomake.vim
```diff
@@ -117,39 +123,75 @@ function! s:MakeJob(make_id, maker) abort
...
+                if !exists('s:jobids_by_makeid[a:make_id]')
+                    let s:jobids_by_makeid[a:make_id] = []
+                endif
+                call add(s:jobids_by_makeid[a:make_id], jobinfo.id)
+            else
+                if has_args
+                    if neomake#utils#IsRunningWindows()
+                        let program = [exe.' '.join(map(args, 'v:val'))]
+                    else
+                        let program = ['/bin/sh', '-c', exe.' '.join(map(args, 'shellescape(v:val)'))]
```
blueyed: Do we need the wrapping in a shell here?  (`sh -c`)
If so, please document why.

autoload/neomake.vim
```diff
@@ -426,7 +471,12 @@ function! s:Make(options, ...) abort
             let jobinfo = s:jobs_by_maker[maker_key]
             let jobinfo.maker.next = copy(a:options)
             try
-                call jobstop(jobinfo.id)
+                call neomake#utils#DebugMessage('Stopping job: ' . jobinfo.id)
+                if has('nvim')
+                    call jobstop(jobinfo.id)
+                else
+                    call job_stop(jobinfo.job)
+                endif
```
blueyed: Use `neomake#CancelJob` here?!

autoload/neomake.vim
```diff
@@ -668,6 +718,22 @@ function! s:RegisterJobOutput(jobinfo, lines) abort
     endif
 endfunction
 
+function! s:accumulate_job_output(id, line) abort
+  call add(s:jobs[a:id].data, a:line)
+endfunction
+
+function! neomake#MakeHandlerVim(channel, line) abort
+    call s:accumulate_job_output(ch_info(a:channel)['id'], a:line)
```
blueyed: `s:accumulate_job_output` can be inlined here?!

doc/neomake.txt
```diff
@@ -439,7 +439,7 @@ You can make quickfixsigns respect Neomake's signs using this option: >
 
 6.2 How to configure the makers to be used?~
 
-See |g:neomake_<ft>_enabled_makers| (press `K` on the link to go there).
+See |g:neomake_<ft>_enabled_makers| (press `<C-]>` on the link to go there).
```
blueyed: Good catch! :)

autoload/neomake.vim
```diff
@@ -102,12 +106,14 @@ function! s:MakeJob(make_id, maker) abort
         call add(args, '%:p')
     endif
 
+    call neomake#utils#DebugMessage('Arguments before expansion: '. string(args))
     if neomake#utils#IsRunningWindows()
-        " Don't expand &shellcmdflag argument of cmd.exe
-        call map(args, 'v:val !=? &shellcmdflag ? expand(v:val) : v:val')
+        " Don't expand &shellcmdflag argument of cmd.exe nor arguments like '"something"'
+        call map(args, "(v:val ==? &shellcmdflag || v:val =~ '.*\".*\".*') ? v:val : expand(v:val)")
     else
```
blueyed: Would be nice to have tests for this, even when they only run on Windows then.
Feel free to factor it out into its own function for that.

saaguero: Is it possible to unit test with Vader mocking IsRunningWindows to return true?

blueyed: I think so.. you would have to overwrite the function in the test.
Not sure about restoring it then though - maybe store the function ref before and then assign it back?!
But if you factor out the part that is inside of `isRunningWindows` then there's no need for it.  And the surrounding function is huge IIRC anyway, so factoring it out is not too bad.

autoload/neomake.vim
saaguero: I've been testing it without this wrapping shell, but without this "trick" it seems that it doesn't work... I did some research but I couldn't find any hint about the reasoning behind this. Projects like vim-gutter and agrep are taking the same approach. Could it be a Vim's bug? In the meantime, do you think a simple comment like this description would be good enough?

saaguero:
Thanks for your suggestions @blueyed! I've just pushed code changes, please take a look.

autoload/neomake/utils.vim
```diff
@@ -68,7 +68,9 @@ function! neomake#utils#WideMessage(msg) abort " {{{2
     "convert tabs to spaces so that the tabs count towards the window
     "width as the proper amount of characters
     let chunks = split(msg, "\t", 1)
+    " vint: -ProhibitUnnecessaryDoubleQuote
     let msg = join(map(chunks[:-2], 'v:val . repeat(" ", &tabstop - strwidth(v:val) % &tabstop)'), '') . chunks[-1]
+    " vint: +ProhibitUnnecessaryDoubleQuote
```
blueyed: There's a vint issue, where it does enable the warning also when called with `--error` in this case: https://github.com/Kuniwak/vint/issues/180.
I think you should just change it to:

    let msg = join(map(chunks[:-2], "v:val . repeat(' ', &tabstop - strwidth(v:val) % &tabstop)"), '') . chunks[-1] 

And not fiddle with controlling Vint through comments.

.travis.yml
```diff
@@ -59,7 +59,7 @@ script:
     elif [ "$ENV" = "vint" ]; then
       vint .
     elif [ "$ENV" = "vint-diff" ]; then
-      vint "$CHANGED_VIM_FILES"
+      vint $CHANGED_VIM_FILES
```
blueyed: Oops, thanks!
Will cherry-pick it to master.

blueyed: https://github.com/neomake/neomake/pull/579

tests/utils.vader
```diff
@@ -12,3 +12,34 @@ Execute (NeomakeTestsEnsureExe creates exe):
...
+  AssertEqual expected_args, args
+
+Execute (Arguments in Linux are expanded correctly):
+  let param1 = '--param1="key1:val1 key2:val2"'
+  let param2 = '--param2=true'
+  let args = ['%', param1 , param2]
+  let expected_args = [expand('%'), param1, param2]
+
+  call neomake#utils#ExpandArgsInLinux(args)
+  AssertEqual expected_args, args
```
blueyed: Thanks for these tests!
:trophy:

doc/neomake.txt
```diff
@@ -56,7 +56,7 @@ fantastic vim plugins such as syntastic and dispatch.
 <
                                                              *:NeomakeCancelJob*
 :NeomakeCancelJob *{job_id}*
```
blueyed: This should not have `*` around it, resulting in a warning:

> Vim(helptags):E154: Duplicate tag "job_id" in file /home/user/.vim/neobundles/.neobundle/doc/neomake.txt

Same below.. just use `{jobid}` here (or  `{job_id}`, but it's not highlighted then correctly, which is a Vim issue though).

`*` is used incorrectly also below, but let's keep that for another PR (e.g. with `*filemode*`).
`*tag*` is meant to create a unique tag in the help files.

doc/neomake.txt
```diff
@@ -22,7 +22,7 @@ CONTENTS                                                      *neomake-contents*
 ==============================================================================
 1. Introduction                                           *neomake-introduction*
 
-Neomake leverages neovim's |job-control| feature where available to run
+Neomake leverages neovim's or vim's |job-control| feature where available to run
```
blueyed: "vim" (and "neovim") should be capitalized.

autoload/neomake/utils.vim
saaguero: That was my problem, disabling/enabling vint checks through comments, then leaves something wrong when running vint --error. I thought that ' ' were necessary everywhere, but your suggestion seems to be working fine, thanks!

doc/neomake.txt
saaguero: Cool finding, I can reproduce this with `:helptags [doc_dir]`... isn't there a vim-doclint? :P

blueyed:
Looks really nice, thanks!
I will give it a try later these days.

doc/neomake.txt
blueyed: [You bet!](https://github.com/machakann/vim-vimhelplint) :)

Will add the following to the Makefile shortly:
```make
build:
	mkdir $@
build/vim-vimhelplint: build
	git clone https://github.com/machakann/vim-vimhelplint $@
vimhelplint: build/vim-vimhelplint
	vim -esN --cmd "set rtp+=build/vim-vimhelplint" -c 'filetype plugin on' \
		-c 'e doc/neomake.txt' -c 'verb VimhelpLintEcho' -c q
```

autoload/neomake.vim
blueyed: Have you tried `let program = [exe] + args`?  (I assume that the escaping is not necessary)

blueyed: I have just read that in the help:

> 						*job-term*
"term": "open"		Start a terminal and connect the job
			stdin/stdout/stderr to it.
			NOTE: Not implemented yet!

saaguero: Excellent, I will add a comment referring to Vim's documentation.

blueyed: So I assume you were not getting anything on stdout/stderr then?

saaguero: You were right, [exe] + args works perfectly both for Linux as well as Windows. Thanks!

blueyed:
Just did a quick test run in Vim, and it worked well.

Please squash the commits (probably just into a single one?!).
I could also do this from the GitHub merge button, but then the description might not be what you expect.

blueyed:
Too bad that this in not tested on Travis with a Vim version that has the required async support..

saaguero:
Let me give a quick check about test in Vim with async support, as I think the last time I tested it they were failing.

We can download and build Vim as part of travis build, it would take more time, but might be good. What do you think?

blueyed:
> Let me give a quick check about test in Vim with async support

Sure, but that's what I've just tested.  And non-async _should_ be covered by tests already.. :innocent: 

We should switch over to https://github.com/tweekmonster/vim-testbed for improved tests with different Vim versions probably.  Requires `sudo=true` though then for all builds, so I am not too convinced.

autoload/neomake/utils.vim
```diff
@@ -200,3 +200,20 @@ function! neomake#utils#GetHighlight(group, what) abort
...
+    endif
+endfunction
+
+function! neomake#utils#ExpandArgsInWindows(args) abort
+    " Don't expand &shellcmdflag argument of cmd.exe nor arguments like '"something"'
+    call map(a:args, "(v:val ==? &shellcmdflag || v:val =~? '.*\".*\".*') ? v:val : expand(v:val)")
+endfunction
+
+function! neomake#utils#ExpandArgsInLinux(args) abort
+    call map(a:args, 'expand(v:val)')
```
blueyed: I have just found a weird issue in this regard, when trying to reproduce something else with the Vim version used on Travis.  On Arch Linux, `expand()` behaves weird:

        " XXX: with locally built Vim 7.3.429 (like on Travis), this expands
        " weird, and results in a command-not-found error:
        " :echo expand("sh -c '1 2 3'")
        " sh
        " -c
        " '1
        " 2
        " 3'
        call map(args, 'expand(v:val)')

I don't know why that does not happen on Travis though?!

I think we should not just use `expand` like that, but only expand words in list items that start with a`%`, or something similar restricted.
This way the hack for Windows would not be required at all?!

saaguero: I've tried setting up a docker instance of Arch but couldn't make it work (pacman seems not working), will look into this later.

OTOH, if I did the math right, we have the following makers relying on this feature:

```
ft/cf3.vim:10:        \ 'args': ['-cf', '%:p'],
ft/nix.vim:10:        \ 'args': ['%:p', '--parse-only'],
ft/elixir.vim:18:      \ 'args': ['credo', 'list', '%:p', '--format=oneline',
ft/elixir.vim:27:      \ 'args': ['dogma', '%:p', '--format=flycheck'],
ft/moon.vim:7:        \ 'args': ['-l', '%:p'],
ft/php.vim:39:        \ 'args': ['%:p', 'text', 'codesize,design,unusedcode,naming'],
ft/c.vim:26:        \ 'args': ['%:p', '--', '-Wall', '-Wextra'],
ft/typescript.vim:23-            \ '%:p', '--format verbose'
```

As you can see, they take advantage of this in order to handle the position of the file path `%:p`

Taking your proposal into account, we might take one of the following approaches:

- expand only when the arg starts with a Vim expansion char: `%` `#` `<`
- do not expand and let the makers' authors handle the expansions, ie: `[expand('%:p'), 'another-arg']`

And yes, whichever approach we pick, it might save us the expansion issues we have in Windows.

blueyed: >expand only when the arg starts with a Vim expansion char: % # <

"%" should be enough.

blueyed: FWIW, `args` can be a callback also - but I am not sure if it gets invoked in the right context.
But that is the way we should do it.

The current misuse of `expand` also prevents you from using something like `sh -c 'date +%N`..

README.md
```diff
@@ -2,19 +2,18 @@
...
 # Neomake
 
-Neomake is a plugin for asynchronous `:make` using
-[Neovim's](http://neovim.org/) job-control functionality. It is inspired by
-the excellent Vim plugins [Syntastic](https://github.com/scrooloose/syntastic)
-and [dispatch.vim](https://github.com/tpope/vim-dispatch).
+A plugin for asynchronous `:make` using [Neovim's](http://neovim.org/) or [Vim's](http://vim.org/)
+job-control functionality. It is inspired by the excellent vim plugins
+[Syntastic](https://github.com/scrooloose/syntastic) and
+[Dispatch](https://github.com/tpope/vim-dispatch).
```
davidosomething: why was this modified? the plugin is called dispatch.vim

README.md
```diff
@@ -2,19 +2,18 @@
 
 # Neomake
 
-Neomake is a plugin for asynchronous `:make` using
-[Neovim's](http://neovim.org/) job-control functionality. It is inspired by
-the excellent Vim plugins [Syntastic](https://github.com/scrooloose/syntastic)
-and [dispatch.vim](https://github.com/tpope/vim-dispatch).
+A plugin for asynchronous `:make` using [Neovim's](http://neovim.org/) or [Vim's](http://vim.org/)
+job-control functionality. It is inspired by the excellent vim plugins
```
davidosomething: `Vim` should be capital

README.md
saaguero: Thanks for the catch, merge problems.

blueyed:
The tests are failing with Vim 7.4.2347 (vim-master).

What is the state of this PR currently?
Do you have an idea how to fix this, have you reproduced it locally?

I've started on a bunch of fixed that are likely conflicting here, e.g. handling the bufnr/winnr correctly for serialized builds.

saaguero:
This is still working fine in both Linux and Windows, and yes, some tests seem to be failing (don't know really why, do you have some suggestions with Vader? I'm trying to debug so to find out some hints...

I would suggest once these tests are fix, we should merge the pull request otherwise it would be harder to keep it up in sync with upstream.

saaguero:
I've found out that the test function [RunNeomake](https://github.com/neomake/neomake/blob/223f8bebaa0a39706829db8e593b7c05a94203b1/tests/_setup.vader#L23-L24) seems to be having problems in Vim (with async support) when running it like `RunNeomake sh`... but if you use it with another maker, like `RunNeomake shellcheck` it looks to be running OK.

Furthermore, the function "Neomake" *does* work as expected with the maker `sh`...

Do you know what might be happening?

saaguero:
I keep testing a little bit more and I've found that the combination of running an async Vim command and immediately a blocking command (ie: sleep n) seems to be interfering with callbacks, since only the close handler is being invoked and no data is available.

Example:

- Open tests/fixtures/error.sh
- Run `Neomake sh | sleep 10m` # fails

The weird thing is that for other makers (ie: shellcheck) it works as expected, furthermore, sleeping 1m also seems to work (but not all the times):

- Run `Neomake sh | sleep 1m` # sometimes work
- Run `Neomake shellcheck | sleep 10m` # works

saaguero:
I created a thread in vim_dev in case someone knows about this behavior:

https://groups.google.com/forum/#!topic/vim_dev/us740TrOxNQ

q0LoCo:
I think the problem is error.sh. I test error.sh without async, it doesn't work too. 
Run error.sh with /bin/sh directly, got 
`error.sh: 6: error.sh: Syntax error: newline unexpected (expecting ")")`
error.sh only has 5 lines, but the syntax error point to line number 6.
Append a empty line to error.sh, it works.

saaguero:
@q0LoCo I'm not having that problem, `/bin/sh -n errors.sh` works fine (ie: it prints syntax errors in line 5)

Just to clarify a little bit more, the underlying problem is that errors are not being retrieved when using the maker sh like:

- Neomake sh | sleep 10m

If you run it without the sleep command, it works fine, and errors signs are being showed as expected.

But the sleep is necessary since it is used for testing as a way to check the job_status before making further assertions.

Take a look at the Vim thread discussion I've created, it includes a simple reproduction example.

saaguero:
@blueyed I've been trying to look for an alternative in the meantime, and I was able to refactor around 4 failing tests (in async Vim) to use shellcheck instead of sh.

The only thing to take into account with this is that we should refactor the Docker images to  include the tool shellcheck. We can either base it on https://hub.docker.com/r/nlknguyen/alpine-shellcheck/ or, like I did it. provision a precompiled shellcheck binary.

Let me know your thoughts.

blueyed:
@saaguero 
1. in https://groups.google.com/d/msg/vim_dev/us740TrOxNQ/0p3Jbwx3AQAJ you are using just `sh errors.sh`, and it should be `sh -n errors.sh` normally.  Does that make a difference.

2. Have you tried splitting it into several commands:

```vim
  function! s:wait_for_finished_jobs()
    while len(neomake#GetJobs())
      echon '.'
      sleep 10m
    endwhile
  endfunction
  command! NeomakeTestsWaitForFinishedJobs call s:wait_for_finished_jobs()
```

3. Regarding shellcheck in the Docker image: I think we should just install it via our `Dockerfile.tests`.

saaguero:
> it should be sh -n errors.sh normally. Does that make a difference?

I've missed that parameter in the Vim thread, but it has the same problem, so there is no difference.

> Have you tried splitting it into several commands

I've tried and got the same problem.

> Regarding shellcheck in the Docker image: I think we should just install it via our Dockerfile.tests.

I've tried following a recipe to compile shellcheck (it needs haskell toolchain) following https://hub.docker.com/r/mitchty/alpine-ghc/~/dockerfile/ but somehow I could not install it through apk.

An alternative would be to upload an already compiled shellcheck for Alpine into this repository and provision it as part of the Dockerfile building phase, that way we can have the tool inside the container.

saaguero:
@blueyed 

I've tried a little bit about what has been discussed in the Vim's thread. I haven't found any clear reason about why is this happening.

I would suggest either of the following alternatives in order to merge this PR:

- Refactor a couple of tests to use shellcheck instead of sh, this will be applicable only for Vim in async mode. Again, for this we will have to adapt the Docker machines, the easiest solution would be to include the precompiled shellcheck binary and install it as part of the Docker image building phase.

- Ignore (with a clear  comment) tests that are failing right now for Vim in async mode. We can even create a bug in this project as a way to track the Vim thread discussion and fix these tests once we have a clear solution.

blueyed:
@saaguero 
Looking into this now.

Have you followed up on https://groups.google.com/d/msg/vim_dev/us740TrOxNQ/mxAclBv7AQAJ in any way?

blueyed:
`make testvim` also hangs randomly at:
```
"[Vader-workbench]" [New File] 
Neomake: Invalid maker name: non-existent
Neomake: Maker not found: nonexistent
Neomake: Nothing to make: no enabled makers.
Neomake: Calling User autocmd: NeomakeFinished
```
for me.

blueyed:
Here is a patch that fixes one test:
```diff
diff --git i/tests/neomake.vader w/tests/neomake.vader
index 9596102..7e2943e 100644
--- i/tests/neomake.vader
+++ w/tests/neomake.vader
@@ -25,8 +25,7 @@ Execute (neomake#Make in project mode with no filetype and no makers):
   let &makeprg = 'sh -c "sleep 0.1"'
   echom '====='
   let job_ids = neomake#Make(0, [])
-  if has('nvim')
-    " Vim runs synchronously.
+  if neomake#has_async_support()
     AssertEqual len(job_ids), 1
     let jobs = neomake#GetJobs()
     AssertEqual len(jobs), 1
```

I've looked briefly into "NeomakeSh: handle unfinished output on exit": it might be that Vim never gets a change to finish the run during that test setup.
That might be the case also for the remaining ~5 failures.
1. I think "Postprocessing should update list" looks like a valid failure: the list is empty, but should not be.

saaguero:
> Have you followed up on https://groups.google.com/d/msg/vim_dev/us740TrOxNQ/mxAclBv7AQAJ in any way?

Kind of, some channel logs here and there, some GDB debugging... the only thing that I've found is that the issue is reproducible, it seems like the channel is closed just before it gets the chance to read (for the case of `RunJob | sleep 50m`)... will update that thread once I get more data, but I don't have much more than what  I've already posted there.

> Here is a patch that fixes one test

Thanks, I've already included this in my local changes, didn't push as I have other tests refactored with shellcheck.

Tests that involve the use of "sh" are the ones that have the most problems with Vim async, those ones I'm referring to when I say to use "shellcheck". That's why I'm proposing to either refactor them or ignore them for VIm in async mode (with an issue to track them.)

blueyed:
> That's why I'm proposing to either refactor them or ignore them for VIm in async mode (with an issue to track them.)

Since those are tests that would trigger issues with Vim in general, I'd rather keep them as-is, and see what the real issue there it.
But I can see that tests might be not representing real usage, especially when wrapped in a command/function.  But exchanging it with some other tools looks like a bad workaround to me.

saaguero:
My suggestion is branching (either creating a new test or with an if) for the failing tests to use shellcheck *only* for Vim in async mode. That way we preserve original tests as usually, for both Neovim as well as Vim without async mode.

There aren't that much failing tests, around 4.

As you said, those tests are not representing a real use case (for now)... let's say that the test discovered a bug in Vim, but not in the general usage of the Neomake plugin. Besides, other makers do work as expected with this combination of `Neomake maker | sleep 50m` so we can even narrowed it down to just `Neomake sh | sleep 50m` maker.

What would you suggest? I'm asking because I think this feature would be nice to be used by the Neomake users and let them start reporting bugs if there happen to be.

blueyed:
>> Have you followed up on https://groups.google.com/d/msg/vim_dev/us740TrOxNQ/mxAclBv7AQAJ in any way?
> Kind of, some channel logs here and there, some GDB debugging...

But you have not posted any update / finding there apparently?

> As you said, those tests are not representing a real use case (for now)

That was just a guess..  I am reluctant to merging anything that existing tests "catch", even if it might be wrong.

> let's say that the test discovered a bug in Vim, but not in the general usage of the Neomake plugin

.. this then means that users are also hit by the same bug, assuming that the behavior is the same as within the tests.

> What would you suggest? I'm asking because I think this feature would be nice to be used by the Neomake users and let them start reporting bugs if there happen to be.

I am currently in the process of fixing issues that affect Neomake in general (from an older stash), and that's why there a conflict already.  Hopefully this will help this PR once it gets rebased.
I am +0 on merging this otherwise - it mainly depends on how much effort you want to put into this after it's been merged after all!
I have a few more things, but hopefully finish most of them during the weekend/today - so then you can rebase it.

blueyed:
I am looking into this now.
One issue I've found is that Vim's job ids start at 0, and that was previously used internally to return "no job was created".

blueyed:
The main issue seems to be that `job_start(['sh', '-c', 'sh -n /path/to/file'])` needs to be used.
Check out `ch_logfile()` which is useful for debugging.

saaguero:
> But you have not posted any update / finding there apparently?

Look at my second post in that thread... I've been gdb debugging just a bit Vim's code but didn't get anything new. I will update the thread discussion once I get more information.

> Check out ch_logfile() which is useful for debugging.

I'm using it, and also I've made some changes in Vim's code based on Bram's suggestions (using ch_logs to log additional information into that ch_logfile.)

> I am currently in the process of fixing issues that affect Neomake in general (from an older stash), and that's why there a conflict already. Hopefully this will help this PR once it gets rebased.

Cool! Go ahead with all your changes so that I can rebase then.

> I am reluctant to merging anything that existing tests "catch", even if it might be wrong.
> I am +0 on merging this otherwise - it mainly depends on how much effort you want to put into this after it's been merged after all!

I completely understand, if a test fails, make it pass. On the other hand, we have identified the root cause, proposing an alternative (ie: using shellcheck instead of sh for those tests.) Also, this seems to be an external bug (Vim's bug) under certain conditions.

So that's why I think it would be a smart move to keep this finding as a "tracking bug" and merge this PR back into upstream.

I'm using this fork every day and seems to be working fine, so up to you (and the community) if they want this to be merged.

blueyed:
Here we go: https://github.com/neomake/neomake/issues/636.

I agree that Vim is buggy in this regard (see my post at vim_dev: https://groups.google.com/d/msg/vim_dev/us740TrOxNQ/IcBgP7YQBQAJ), but for it means to not merge it then: people will report strange issues, just like it's the case with buffer_output=0 by default with multiline-linters.
There is a major hack in the updated PR, but only activated for test currently - although I could imagine that it would need to be activated for Vim in general - until the issue is fixed there.
Therefore I tend to only merge this once it's fixed in Vim, and then we require the patch that fixed it.

I'd be happy btw to push this to the neomake repo instead and give you access here, so we can work on the branch together.

A lot of things however could just go into master, and then this PR could rebased on top of it again.

Thanks for staying with me btw.. :)

saaguero:
That's great! We both found some strange things going on in Vim's async.

Your workaround looks great!  (giving a 0.1 sleep in the shell itself)

Should we start working only in PR #637 and closed this one?

blueyed:
> Your workaround looks great! (giving a 0.1 sleep in the shell itself)

Except for that it needs to handle the exit status / return value.  Will be true then always at the moment.

> Should we start working only in PR #637 and close this one?

OK.

PR #576, opened 2016-09-02
sh: pass in filetype to shellcheck

blueyed:
This avoids the warning from shellcheck in case if there is no shebang.
This does not parse / look closely on any existing shebang line on purpose.
Ref: https://github.com/neomake/neomake/pull/409#issuecomment-219785830

/cc @lverweijen


blueyed:
What do you think about using "sh" for an empty filetype?

PR #581, opened 2016-09-04
Travis: add vimhelplint

blueyed:
Currently it does not work with Travis' Vim though: https://github.com/machakann/vim-vimhelplint/issues/5.

Let's wait on that issue.

Alternatively we should/could use a Docker based approach for the tests anyway.


blueyed:
Works: https://travis-ci.org/neomake/neomake/jobs/158331845#L443.

It gets only run if the doc was changed, so should be green again (for this PR).

PR #590, opened 2016-09-07
Support multi-line errorformat for Flow

fortes:
Newline stripping didn't seem to work when using Flow for JS. I've updated the `errorformat` to support the multi-line format that flow is outputting.

### DIFF
--- a/autoload/neomake/makers/ft/javascript.vim
+++ b/autoload/neomake/makers/ft/javascript.vim
@@ -41,12 +41,9 @@ function! neomake#makers#ft#javascript#standard()
 endfunction
 
 function! neomake#makers#ft#javascript#flow()
-    " Replace "\n" by space.
-    let mapexpr = 'substitute(v:val, "\\\\n", " ", "g")'
     return {
         \ 'args': ['--old-output-format'],
-        \ 'errorformat': '%E%f:%l:%c\,%n: %m',
-        \ 'mapexpr': mapexpr,
+        \ 'errorformat': '%E%f:%l:%c\,%n: %m,%Z%m',
         \ }
 endfunction
 


blueyed:
Check out the "buffer_output" setting also: https://github.com/neomake/neomake/blob/ec4d0b0a35791a486b836850d17ed9021b4fef0f/doc/neomake.txt#L159.

I think this should be enabled by default: there's not much gain by processing output while it comes normally I would say.

fortes:
@blueyed Thanks for the pointer. I tried that locally just now, but it doesn't not fix the issue on my setup.

fortes:
Here is the relevant verbose output:

```
Neomake: [#7] stdout: flow: ['', '/home/fortes/test.js:21:1,5: function call', 'Error:', '/home/fortes/test.js:21:4,4: number', 'This type is incompatible with', '/home/fortes/test.js:17:19,24: string', '', 'Found 1 error', '']
Neomake: [#7] flow: processing 8 lines of output.
Neomake: [#7] Removing invalid entry: {'lnum': 0, 'bufnr': 0, 'col': 0, 'valid': 0, 'vcol': 0, 'nr': -1, 'type': '', 'maker_name': 'flow', 'pattern': '', 'text': ''}
```

fortes:
```
Neomake: [#10] stdout: flow: ['', '/home/fortes/test.js:8:17,48: ./resizable_view.less', 'Required module not found', '', '/home/fortes/test.js:9:20,40: classnames', 'Required module not found', '', 'Found 2 errors', '']
```

With verbose output, you can see that the results are coming in on two lines.

blueyed:
Seems to work for me, using the approach from https://github.com/neomake/neomake/pull/677 with the following linter output and the errorformat `'%E%f:%l:%c\,%n: %m,%Z%m'`:
```

/home/fortes/test.js:8:17,48: ./resizable_view.less
Required module not found

/home/fortes/test.js:9:20,40: classnames
Required module not found

Found 2 errors```

You might want to stick a `,%-G` at the end to remove the empty lines.

PR #593, opened 2016-09-08
Allow buflocal maker settings

sanmiguel:
The way `neomake` expands the filename doesn't allow me to programmatically create a string that contains some expansion of the buffer name (e.g. `expand('%:t:r') . '_SUITE'`)

By allowing me to set e.g. `b:neomake_erlang_eunit_maker` I can construct this string myself when opening the buffer, and set a maker specific to this buffer.

```
    let b:neomake_erlang_eunit_maker = {
        \ 'exe': 'rebar',
        \ 'args': ['-q', 'eunit', 'skip_deps=true', 'suites=' . expand('%:t:r')],
        \ }
```

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -192,7 +192,7 @@ function! neomake#GetMaker(name_or_maker, ...) abort
         if len(fts)
             for ft in fts
                 for scope in [b:, g:]
-                    let m = get(g:, 'neomake_'.ft.'_'.a:name_or_maker.'_maker')
+                    let m = get(scope, 'neomake_'.ft.'_'.a:name_or_maker.'_maker')
                     if type(m) == type({})
                         let maker = m
                         break
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -191,12 +191,14 @@ function! neomake#GetMaker(name_or_maker, ...) abort
     else
         if len(fts)
             for ft in fts
-                let m = get(g:, 'neomake_'.ft.'_'.a:name_or_maker.'_maker')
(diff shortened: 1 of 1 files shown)


autoload/neomake.vim
```diff
@@ -191,12 +191,14 @@ function! neomake#GetMaker(name_or_maker, ...) abort
...
         if len(fts)
             for ft in fts
-                let m = get(g:, 'neomake_'.ft.'_'.a:name_or_maker.'_maker')
-                if type(m) == type({})
-                    let maker = m
-                    break
-                endif
-                unlet m
+                for scope in [b:, g:]
+                    let m = get(g:, 'neomake_'.ft.'_'.a:name_or_maker.'_maker')
```
davidosomething: still using `g:` ?

sanmiguel: Good spot - artifact from a rebase. Let me fix that.

davidosomething:
so this basically allows one to configure `b:neomake_FT_MAKER_maker` when `g:neomake_FT_MAKER_maker` does not exist ? I am not certain what it does since setting up `b:neomake_FT_MAKER_maker` is already a feature:
https://github.com/neomake/neomake/blob/master/autoload/neomake.vim#L236-L254

sanmiguel:
Oh, hah! I actually wrote this patch about 6-8 months ago, but forgot about it. I didn't see anything obvious when I looked (but to be fair I clearly didn't look very hard), so rebased and submitted.

That section you linked explains why this worked when I tested my change, even though I'd made that mistake.

Thanks a lot! I'll close this PR :)

PR #595, opened 2016-09-08
python: Add py3kwarn maker

Mattias-:
[py3kwarn](https://github.com/liamcurry/py3kwarn) detects code that is not Python 3 compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,7 @@ Python:
 - python
 - [vulture](https://bitbucket.org/jendrikseipp/vulture) [not enabled by default]
 - [mypy](http://mypy-lang.org/) [not enabled by default]
+- [py3kwarn](https://github.com/liamcurry/py3kwarn) [not enabled by default]
 
 Ruby:
 - mri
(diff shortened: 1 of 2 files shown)


blueyed:
Please ignore the Travis build issue (although the "abort" is missing also with your function definition): I'm working on improving the Python makers anyway, and will address those then.

Mattias-:
@blueyed 👍 Thank you! I should have checked how the functions where defined for other ft, will be more careful next time!

Perhaps a 'howto' or best practices section could be added to [CONTRIBUTING.md](https://github.com/neomake/neomake/blob/master/CONTRIBUTING.md) on the preferred way to define makers.

blueyed:
> I should have checked how the functions where defined for other ft, will be more careful next time!

Don't worry, most of them are not e.g. using the `abort` attribute.  Honestly it's also not needed in this simple case (AFAIK), but we have some linter for that now.. ;)

Improving the CONTRIBUTING guide would be nice, yes.

PR #601, opened 2016-09-09
:NeomakeInfo - different message when no ft

davidosomething:
didn't even test it, hopefully i didn't make some glaring mistake

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -886,8 +886,12 @@ function! neomake#DisplayInfo() abort
     echom '== Enabled makers =='
     echom 'For the current filetype (with :Neomake): '
                 \ .string(neomake#GetEnabledMakers(ft))
-    echom 'You can define g:neomake_'.ft.'_enabled_makers'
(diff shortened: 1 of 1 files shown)


autoload/neomake.vim
```diff
@@ -886,8 +886,12 @@ function! neomake#DisplayInfo() abort
     echom '== Enabled makers =='
     echom 'For the current filetype (with :Neomake): '
                 \ .string(neomake#GetEnabledMakers(ft))
-    echom 'You can define g:neomake_'.ft.'_enabled_makers'
+    if !empty(ft)
+        echom 'You can define g:neomake_'.ft.'_enabled_makers'
                 \ .' to configure it (or b:neomake_'.ft.'_enabled_makers).'
+    else
+        echom 'The current buffer does not have a filetype.'
+    endif
```
blueyed: Please switch the `if` statement to have the positive case first.
And then it should also be covered by a test - see the existing one for it, which should be simple to extend.

PR #610, opened 2016-09-13
Add flake8 error code to displayed error message

shvenkat:
It is useful to know the flake8 error code when you need to: (a) look it up in documentation, or (b) suppress it by updating the flake8 config or directive in the code comments (e.g. `# flake8: disable=F401`).

Currently, the status line message does not show the error code, just the message:
```
flake8: 'foo' imported but unused
```

This PR adds the error code to the message, as follows:
```
flake8: F401 'foo' imported but unused
```

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -83,6 +83,7 @@ function! neomake#makers#ft#python#Flake8EntryProcess(entry)
     else
         let type = ''
     endif
+    let a:entry.text = a:entry.type . a:entry.nr . ' ' . a:entry.text
     let a:entry.type = type
 endfunction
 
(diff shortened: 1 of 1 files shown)


shvenkat:
The Travis build error appears to be unrelated to this PR.

blueyed:
What about removing it from `entry.nr` then?

With this PR it is a bit duplicated:

> t.py|1 col 1 error 401| F401 'foo' imported but unused

blueyed:
@shvenkat 
What do you think regarding my comment?

shvenkat:
@blueyed 
Makes sense. Pushed another commit -- is this what you had in mind?

blueyed:
Exactly, thanks!

Issue #598, opened 2016-09-08
Change default for buffer_output


blueyed:
Doc: https://github.com/neomake/neomake/blob/ec4d0b0a35791a486b836850d17ed9021b4fef0f/doc/neomake.txt#L159

The setting updates the errors as soon as input is coming from a maker, which is often only a few chars/bytes.

It causes problems often, and does not provide much benefit: usually linters output all there findings in the end.  In this case it causes also extra processing.

Especially with certain things in the `&errorformat` (e.g. `%Z` or any multiline indication in general), it should be disabled by default.  But it's easier to just disable it by default always.

What do you think?

blueyed:
For reference, I've initially started on detecting this automatically based on the errorformat:
```diff
diff --git i/autoload/neomake/utils.vim w/autoload/neomake/utils.vim
index f60e5bb..57c704f 100644
--- i/autoload/neomake/utils.vim
+++ w/autoload/neomake/utils.vim
@@ -59,6 +59,10 @@ function! neomake#utils#DebugObject(msg, obj) abort
     call neomake#utils#DebugMessage(a:msg.' '.neomake#utils#Stringify(a:obj))
 endfunction
 
+function! neomake#utils#GetBufferOutputDefault(efm)
+    return a:efm =~# '\v(^|,)\%C|\%Z'
+endfunction
+
 " This comes straight out of syntastic.
 "print as much of a:msg as possible without "Press Enter" prompt appearing
 function! neomake#utils#WideMessage(msg) abort " {{{2
diff --git i/tests/utils.vader w/tests/utils.vader
index 0625a53..3f509bd 100644
--- i/tests/utils.vader
+++ w/tests/utils.vader
@@ -70,3 +70,10 @@ Execute(neomake#utils#redir):
   endif
   AssertEqual neomake#utils#redir(['NeomakeTestCommand', 'echon 3']), "\n1\n23"
   AssertThrows neomake#utils#redir(['NeomakeTestCommand', 'echoerr 3'])
+
+Execute (GetBufferOutputDefault):
+  AssertEqual neomake#utils#GetBufferOutputDefault('%E%f:%l: %m'), 0
+  AssertEqual neomake#utils#GetBufferOutputDefault('%E%f:%l: %m,%Cfoo'), 1
+  AssertEqual neomake#utils#GetBufferOutputDefault('%E%f:%l: %m,%Zfoo'), 1
[...]

Issue #549, opened 2016-08-10
New Rust error message format


antoyo:
Hello.
Rust nightly now shows [errors in a new format](https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html).
It would be nice to support it in neomake (if the new format is fixed).
Thanks.

antoyo:
[The new format is now in Rust stable.](https://blog.rust-lang.org/2016/09/29/Rust-1.12.html)
So, I believe it we should switch to it now.
I currently use this `errorformat`, but it does not always work:
```vim
let g:neomake_rust_clippy_maker = {
    \ 'exe': 'cargo',
    \ 'args': ['clippy'],
    \ 'errorformat':
        \   '%-Gerror: Could not compile%.%#,'
        \.  '%-Gerror: aborting due to%.%#,'
        \.  '%Eerror[%.%#]: %m,%C\\s%#--> %f:%l:%c,'
        \.  '%Eerror: %m,%C\\s%#--> %f:%l:%c,'
        \.  '%Wwarning: %m,%C\\s%#--> %f:%l:%c,'
        \.  '%Ihelp: %m,'
        \.  '%I\\s%#= help: %m,'
        \.  '%I\\s%#= note: %m,'
        \,
    \ 'append_file': 0,
    \ }
```

blueyed:
Check https://github.com/neomake/neomake/pull/613 I guess?!
Closing this in favor of #613.

blueyed:
@antoyo 
Please provide feedback there, and then we can get it finalized/merged.

Issue #428, opened 2016-05-19
Provide hook to populate quickfix/loclist through Vim function


yyadavalli:
The `ensime-vim` plugin provides `EnTypeCheck` command to check scala files. 
Is there a way I can create a maker for neomake that uses the vim command?

P.S. I did not find any such mechanism on searching, want to make sure I did not miss something. I'd be willing send to a pull request if there is interest

wsdjeg:
As I know neomake's maker is the tool in your PATH. you can run it in your terminal. I am not sure if there is a way to use the command in vim RTP

blueyed:
@wsdjeg 
If I understand it correctly, the question is about using a Vim command instead of an executable.

@yyadavalli 
You'll have to investigate this yourself by looking at the code/doc - I don't know it myself without looking into it.
Should be possible in general, but might need some code changes.

blueyed:
I think it would be useful in general (and also for the tests) to have a `func` setting/attribute, that would be preferred to `exe`.  This would call a function instead of an external program.
But that would not be async, since `jobstart` (in Neovim) and `job_start` (in Vim) is only meant for calling external programs..

andreypopp:
The same is required to integrate with https://github.com/the-lambda-church/merlin

blueyed:
@andreypopp 
How would the integration with merlin look like, i.e. which command/function would you like to use?

andreypopp:
See how it is done for syntactic: https://github.com/the-lambda-church/merlin/blob/master/vim/merlin/syntax_checkers/ocaml/merlin.vim it calls a function.

blueyed:
@andreypopp 
It seems to hook into getting the errors for the loclist though, not running a maker through a function.
That seems to be a more universal approach altogether.

And ensime-vim appears to use the same approach: https://github.com/ensime/ensime-vim/blob/98d52714333bf07c96f494d7b0879e6f56d2815b/plugin_integrations/syntastic/syntax_checkers/scala/ensime.vim#L8-L14.

So we should provide a way to skip running a maker, but have a hook/function being run to populate the quickfix / location list (for ensime-vim: https://github.com/ensime/ensime-vim/blob/c258770c2f0115e47a498b07a54bd01fd2a2d064/ensime_shared/typecheck.py#L49-L61).

andreypopp:
Alright, any pointers on how to implement that?

blueyed:
Not really, apart from what I wrote above.
[...]

PR #613, opened 2016-09-17
New errorfmt for Rust

### LINKED ISSUE #592: Rustc/cargo error format broken
The new nightly format is different from the previous format, hence the cargo maker does not work anymore

fuine:
This PR closes #592. Stable 1.12 version of Rust, which introduces new errorfmt as default, will land in a few days and i think that it's reasonable to switch to the new errorfmt.

Introduced makers try to remove unnecessary informations, while maintaining error/warning messages, error description taken from the pointer line and note/help lines. The only thing that is missing is proper parsing of error/warning width, as pointer line contains both markers (^^^^) and error description.

More information about new format can be found [here](https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html). I decided against parsing lifetime messages, because in my opinion they can't be shown in a sane way.

### DIFF
--- a/autoload/neomake/makers/cargo.vim
+++ b/autoload/neomake/makers/cargo.vim
@@ -1,11 +1,9 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#cargo#cargo()
+function! neomake#makers#cargo#cargo() abort
     return {
         \ 'args': ['build'],
         \ 'errorformat':
-            \   '%-Z%f:%l,' .
-            \   '%+C %s,' .
-            \   '%A%f:%l:%c: %*\d:%*\d\ %t%*[^:]: %m,',
+            \ neomake#makers#ft#rust#rustc()['errorformat'],
         \ }
 endfunction
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -1,19 +1,24 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#ft#rust#EnabledMakers()
+function! neomake#makers#ft#rust#EnabledMakers() abort
     return ['rustc']
 endfunction
 
-function! neomake#makers#ft#rust#rustc()
+function! neomake#makers#ft#rust#rustc() abort
     return {
-        \ 'args': ['-Z', 'parse-only'],
         \ 'errorformat':
-            \ '%-G%f:%s:,' .
-            \ '%f:%l:%c: %trror: %m,' .
-            \ '%f:%l:%c: %tarning: %m,' .
-            \ '%f:%l:%c: %m,'.
-            \ '%f:%l: %trror: %m,'.
-            \ '%f:%l: %tarning: %m,'.
-            \ '%f:%l: %m',
+            \ '%-Gerror: aborting due to previous error,'.
+            \ '%-Gerror: aborting due to %\\d%\\+ previous errors,'.
+            \ '%-Gerror: Could not compile `%s`.,'.
+            \ '%Eerror[E%n]: %m,'.
+            \ '%Eerror: %m,'.
+            \ '%Wwarning: %m,'.
+            \ '%Inote: %m,'.
+            \ '%-Z\ %#-->\ %f:%l:%c,'.
+            \ '%G\ %#\= %*[^:]: %m,'.
+            \ '%G\ %#|\ %#%\\^%\\+ %m,'.
+            \ '%I%>help:\ %#%m,'.
+            \ '%Z\ %#%m,'.
+            \ '%-G%s',
         \ }
 endfunction
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -4,14 +4,7 @@
 let s:rustup_has_nightly = -1
 
 function! neomake#makers#clippy#clippy() abort
-    let errorfmt = '%Eerror[E%n]: %m,'.
-                 \ '%Eerror: %m,'.
-                 \ '%Wwarning: %m,'.
-                 \ '%Inote: %m,'.
-                 \ '%-Z\ %#-->\ %f:%l:%c,'.
-                 \ '%I\ %#\= %t%*[^:]: %m,'.
-                 \ '%I\ %#|\ %#%\\^%\\+ %m,'.
-                 \ '%-G%s,'
+    let errorfmt = neomake#makers#ft#rust#rustc()['errorformat']
 
     " When rustup and a nightly toolchain is installed, that is used.
     " Otherwise, the default cargo exectuable is used. If this is not part


blueyed:
Thanks!

Do you feel like fixing the vint issues in the touched files?  (just add abort to all functions, i.e. `function! neomake#makers#cargo#cargo() abort`)

(Ref https://github.com/Kuniwak/vint/issues/181)

I think we should not merge this before the new version is released, should we?

fuine:
Sure, I will fix vints issues for rust. I'm not sure if we shouldn't leave old errorfmt for the time being and just add new to makers. People might still use older versions of compiler for some time and maybe it'd not be the worst idea. That said i have almost no experience with errorfmts and neomake and i'm not sure if long list of errorfmts doesn't induce some kind of performance hit.

blueyed:
Please check https://github.com/neomake/neomake/issues/549#issuecomment-250550544.

autoload/neomake/makers/cargo.vim
```diff
@@ -1,11 +1,16 @@
...
         \ 'args': ['build'],
         \ 'errorformat':
-            \   '%-Z%f:%l,' .
-            \   '%+C %s,' .
-            \   '%A%f:%l:%c: %*\d:%*\d\ %t%*[^:]: %m,',
+            \ '%Eerror[E%n]: %m,'.
+            \ '%Eerror: %m,'.
+            \ '%Wwarning: %m,'.
+            \ '%Inote: %m,'.
+            \ '%-Z\ %#-->\ %f:%l:%c,'.
```
YorickPeterse: @fuine I believe `-Z` is deprecated, and it will trigger warnings. For example, using this Rust file:

```rust
fn main() {
    let $syntax_error = "hello";
}
```

The output in the quickfix window is:

```
|| the option `Z` is unstable and should only be used on the nightly compiler, but it is currently accepted for backwards compatibility; this will soon change, see issue #31847 for more details
/tmp/test.rs|2 col 9 error| expected pattern, found `$`
|| aborting due to 2 previous errors
```

YorickPeterse: See https://github.com/rust-lang/rust/issues/31847 for more information.

fuine: Are you referring to the -Z no-trans in rustc maker? I should probably remove that flag as it is currently unstable. I think that cargo maker is fine.

YorickPeterse: @fuine I think the entire `-Z` flag is unstable at the moment, but I'm not certain.

fuine: @YorickPeterse i am aware of that, but the only usage of that flag in this PR is in the `rust.vim` file. I would like to know why did you comment on `cargo.vim` file. Is there something i am missing here?

YorickPeterse: @fuine I commented on this particular line because it was the first instance of `-Z` being used, that's all.

fuine: Ah ok, this particular -Z is actually the end of multiline message in vim's errorfmt (refer to [docs](http://vimdoc.sourceforge.net/htmldoc/quickfix.html#errorformat) to see more. Anyway thanks for pointing that out, i removed it from rust.vim.

fuine:
@antoyo i decided against putting `= note` and `= help` in the errorfmt because the only place i have seen them were clippy error messages and it looks like they don't come in one chunk (neomake interprets error/warning message as one chunk and then slurps additional `= note/help` lines as another chunk and we lose information about line/file at this point. If you've seen other examples where this is not the case feel free to describe it and we will try to put this in the errorfmt.

@YorickPeterse i removed `-Zno-trans` from the rustc call. Now neomake will just call rustc without any flags which will build whole file and emit binary. I'm not sure if it's desired behaviour, but i wasn't able to find any other stable flags that would prevent emission of binary and only report errors/warns. One solution would be to keep on using `-Zno-trans` and ignore the warning in the errorfmt, but i don't think that is a good solution.

@blueyed I think the PR is ready to be merged. One thing that is interesting is the json error format which currently is unstable but will probably stabilize soon. Do you have any plans to support such format? As you can see [here](https://i.imgur.com/XfO6vXN.png) it contains a lot more precise information to work with (locaction list contains current output of the new errorfmt and on the right you can see the 'normal' error message and json output). Currently there's no sane way to parse the lifetime annotations and show them in neomake.

blueyed:
> neomake interprets error/warning message as one chunk and then slurps additional = note/help lines as another chunk and we lose information about line/file at this point

Sounds like it's related to buffer_output (https://github.com/neomake/neomake/issues/598)?!  Or do you mean something else?

As for json: there is `json_decode` and we could map it to location list entries I guess?!
See https://github.com/neomake/neomake/issues/428.

fuine:
> > neomake interprets error/warning message as one chunk and then slurps additional = note/help lines  as another chunk and we lose information about line/file at this point

> Sounds like it's related to buffer_output (#598)?! Or do you mean something else?

Hmm, i had the impression that it chopped the output of the clippy (one consistent warning message was parsed as 2 lines of output), but i have just tried it again and this time it works. I'm not sure if it was related to stderr flushing by clippy or maybe it was me who made some stupid mistake. Anyways it works now and i have added additional `help` annotations that sometimes appear in the compiler output. 

I noticed that we use the same errorfmt in three different files (`clippy.vim`, `cargo.vim`, `rust.vim`). Wouldn't it make more sense if we define it in one place and then use the definition in these three makers?

As for json i will look into it after that error format stabilizes. I think that we should merge this PR first and if it will be more beneficial to parse json then new PR should be opened.

blueyed:
> Hmm, i had the impression that it chopped the output of the clippy (one consistent warning message was parsed as 2 lines of output), but i have just tried it again and this time it works. 

It depends on when it's flushed and e.g. if it is received during the "1st second" (but with only second-based resolution) - very fragile code there.. working on it in https://github.com/neomake/neomake/pull/607.

> I noticed that we use the same errorfmt in three different files (clippy.vim, cargo.vim, rust.vim). Wouldn't it make more sense if we define it in one place and then use the definition in these three makers?

Totally..! You should use the existing/leading/main function then to just return it in the other maker definitions.
Do you want to do this in this PR?  Should be a separate commit then I'd say.

blueyed:
JFI: #598 has been merged - maybe it helps?! (i.e. please rebase)

fuine:
I have rebased and introduced a single definition of rust's errorfmt (in `rust.vim`). Please note that i have almost no experience in writing vimscript so any feedback on that small refactor would be helpful.
I have also decided to add ignore patterns against 3 error messages that i feel are redundant.

Do you want me to squash my commits before you merge this PR?

autoload/neomake/makers/ft/rust.vim
```diff
@@ -1,19 +1,24 @@
...
+            \ '%Eerror[E%n]: %m,'.
+            \ '%Eerror: %m,'.
+            \ '%Wwarning: %m,'.
+            \ '%Inote: %m,'.
+            \ '%-Z\ %#-->\ %f:%l:%c,'.
+            \ '%G\ %#\= %*[^:]: %m,'.
+            \ '%G\ %#|\ %#%\\^%\\+ %m,'.
+            \ '%I%>help:\ %#%m,'.
+            \ '%Z\ %#%m,'.
+            \ '%-G%s,',
```
blueyed: Not sure about the trailing comma here (in the errorformat), better remove it if it's not there on purpose.

blueyed:
LGTM, please squash and address the comment I've left.

fuine:
Removed the trailing comma and squashed, feel free to merge.

blueyed:
Thanks a lot!

jonhoo:
@fuine I'm not too familiar with the internals of neomake, so I don't know how hard this would be to do, but currently most type errors are simply rendered as `cargo: mismatched types`. This is not a particularly helpful error message. I suspect the reason is that Rust's new error messages include the type information on *later* lines of the output, and not in the `error:` line. Is there some way to still match the type information and emit it in the neomake output?

fuine:
@jonhoo you are right, but the maker matches these lines. If you open quickfix/loclist you will see the rest of the error. There's no easy way to make it appear by deafult without opening lists.

jonhoo:
That's true, but running `:cope` and `:cclo` (even with hotkeys) is sort of tedious. If we could come up with a good regex to bring the types in question into the statusline, that would be awesome! I don't even know where to begin implementing something like that though :/

fuine:
I agree, can you open an issue about that problem? I will try to do something about it in the nearest future, because that is one of the few inconveniences that i hit with the current state of things.

jonhoo:
@fuine see #793.

PR #616, opened 2016-09-20
Rust clippy maker

Eisfreak7:
Clippy is a rust linter that has more lints than standard rustc (`Neomake! clippy` will show the rustc/cargo lints as well as the clippy ones).
Since clippy integrates with the rust compiler, it has the same errorformat (I copied the error format from #613). Clippy needs a nightly compiler to work. Since it is the most convenient way, `rustup` (manages multiple rust installations) is used if available to run clippy with nightly rust, while other compilations can use stable rust.

I also added cargo and clippy to the Readme, with the annotation that `Neomake!` needs to be used (wasn't obvious for me).

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,8 @@ Ruby:
 Rust:
 
 - rustc
+- cargo [Call with `Neomake! cargo`]
+- [clippy](https://github.com/Manishearth/rust-clippy) [Call with `Neomake! clippy`, needs [rustup](https://github.com/rust-lang-nursery/rustup.rs) with a nightly toolchain or a nightly rust installation]
 
 Scala:
 
--- a/README.md
+++ b/README.md
@@ -230,8 +230,8 @@ Ruby:
 Rust:
 
 - rustc
-- cargo [Call with `Neomake! cargo`]
-- [clippy](https://github.com/Manishearth/rust-clippy) [Call with `Neomake! clippy`, needs [rustup](https://github.com/rust-lang-nursery/rustup.rs) with a nightly toolchain or a nightly rust installation]
+- cargo (call with `Neomake! cargo`)
+- [clippy](https://github.com/Manishearth/rust-clippy) (call with `Neomake! clippy`, needs nightly rust, supports [rustup](https://github.com/rust-lang-nursery/rustup.rs))
 
 Scala:
 
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -1,5 +1,8 @@
 " vim: ts=4 sw=4 et
 
+" Yet to be determined
+let s:rustup_has_nightly = -1
+
 function! neomake#makers#clippy#clippy() abort
     let errorfmt = '%Eerror[E%n]: %m,'.
                  \ '%Eerror: %m,'.
@@ -13,7 +16,7 @@ function! neomake#makers#clippy#clippy() abort
     " When rustup and a nightly toolchain is installed, that is used.
     " Otherwise, the default cargo exectuable is used. If this is not part
     " of a nightly rust, this will fail.
-    if !exists('s:rustup_has_nightly')
+    if !exists('s:rustup_has_nightly') || s:rustup_has_nightly == -1
         if !executable('rustup')
             let s:rustup_has_nightly = 0
         else
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -21,7 +21,7 @@ function! neomake#makers#clippy#clippy() abort
             let s:rustup_has_nightly = 0
             call system('rustc --version | grep -q "\-nightly"')
             if v:shell_error
-                call neomake#utils#LoudMessage('Clippy requires a nightly rust installation.')
+                call neomake#utils#ErrorMessage('Clippy requires a nightly rust installation.')
             endif
         else
             call system('rustup show | grep -q "^nightly-"')
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -19,6 +19,10 @@ function! neomake#makers#clippy#clippy() abort
     if !exists('s:rustup_has_nightly') || s:rustup_has_nightly == -1
         if !executable('rustup')
             let s:rustup_has_nightly = 0
+            call system('rustc --version | grep -q "\-nightly"')
+            if v:shell_error
+                call neomake#utils#LoudMessage('Clippy requires a nightly rust installation.')
+            endif
         else
             call system('rustup show | grep -q "^nightly-"')
             let s:rustup_has_nightly = !v:shell_error
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -16,7 +16,7 @@ function! neomake#makers#clippy#clippy() abort
     " When rustup and a nightly toolchain is installed, that is used.
     " Otherwise, the default cargo exectuable is used. If this is not part
     " of a nightly rust, this will fail.
-    if !exists('s:rustup_has_nightly') || s:rustup_has_nightly == -1
+    if s:rustup_has_nightly == -1
         if !executable('rustup')
             let s:rustup_has_nightly = 0
             call system('rustc --version | grep -q "\-nightly"')
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -15,9 +15,9 @@ function! neomake#makers#clippy#clippy() abort
     " of a nightly rust, this will fail.
     if !exists('s:rustup_has_nightly')
         if !executable('rustup')
-            let s:rustup_has_nightly = false
+            let s:rustup_has_nightly = 0
         else
-            call system('rustup show | grep -q nightly')
+            call system('rustup show | grep -q "^nightly-"')
             let s:rustup_has_nightly = !v:shell_error
         endif
     endif
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -25,4 +25,5 @@ function! neomake#makers#clippy#clippy() abort
             \ 'args': ['clippy'],
             \ 'errorformat': errorfmt,
             \ }
+    endif
 endfunction
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -1,0 +1,28 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#clippy#clippy()
+    let errorfmt = '%Eerror[E%n]: %m,'.
+                 \ '%Eerror: %m,'.
+                 \ '%Wwarning: %m,'.
+                 \ '%Inote: %m,'.
+                 \ '%-Z\ %#-->\ %f:%l:%c,'.
+                 \ '%I\ %#\= %t%*[^:]: %m,'.
+                 \ '%I\ %#|\ %#%\\^%\\+ %m,'.
+                 \ '%-G%s,'
+
+    " When rustup and a nightly toolchain is installed, that is used
+    " Otherwise, the default cargo exectuable is used. If this is not part
+    " of a nightly rust, this will fail.
+    if executable('rustup') && system('rustup show | grep nightly | wc -l') >= 2
+        return {
+            \ 'exe': 'rustup',
+            \ 'args': ['run', 'nightly', 'cargo', 'clippy'],
+            \ 'errorformat': errorfmt,
+            \ }
+    else
+        return {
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -230,6 +230,8 @@ Ruby:
 Rust:
 
 - rustc
+- cargo (call with `Neomake! cargo`)
+- [clippy](https://github.com/Manishearth/rust-clippy) (call with `Neomake! clippy`, needs nightly rust, supports [rustup](https://github.com/rust-lang-nursery/rustup.rs))
```
blueyed: > needs nightly rust

That will certainly change, so better have a more relative version/naming here..

autoload/neomake/makers/clippy.vim
```diff
@@ -0,0 +1,28 @@
...
+    let errorfmt = '%Eerror[E%n]: %m,'.
+                 \ '%Eerror: %m,'.
+                 \ '%Wwarning: %m,'.
+                 \ '%Inote: %m,'.
+                 \ '%-Z\ %#-->\ %f:%l:%c,'.
+                 \ '%I\ %#\= %t%*[^:]: %m,'.
+                 \ '%I\ %#|\ %#%\\^%\\+ %m,'.
+                 \ '%-G%s,'
+
+    " When rustup and a nightly toolchain is installed, that is used
```
blueyed: s/used/used./

Also refer to the version number of the next release, rather then "nightly" here.

autoload/neomake/makers/clippy.vim
```diff
@@ -0,0 +1,28 @@
...
+    " Otherwise, the default cargo exectuable is used. If this is not part
+    " of a nightly rust, this will fail.
+    if executable('rustup') && system('rustup show | grep nightly | wc -l') >= 1
```
blueyed: I would cache the `system` call here maybe: IIRC this function gets called not only once.

```vim
if !exists('s:is_rustup_nightly')
    call system('rustup show | grep -q nightly')
    let s:is_rustup_nightly = !v:shell_error
endif
```

README.md
Eisfreak7: Probably won't, at least not soon.
Here's the developer saying it: https://www.reddit.com/r/rust/comments/4le45w/clippy_on_stable/d3mnrxe
Basically, clippy needs access to compiler internals that can only be accessed with a nightly compiler, since they are inherently not stable.

autoload/neomake/makers/clippy.vim
Eisfreak7: I added the punctuation.
As I said above, clippy actually needs nightly rust, not a specific version.
It probably officialy always needs the latest nightly, though in practice it'll work with an older one as long as there weren't any substantial changes.

autoload/neomake/makers/clippy.vim
Eisfreak7: I thought about that, but how'd I do that? Introduce a global variable?
I haven't really used viml much outside of my own vimrc.
Besides that, it's probably a negligable performance hit. But if you have a suggestion how to cache it I'll gladly do it.

autoload/neomake/makers/clippy.vim
```diff
@@ -0,0 +1,28 @@
...
+            \ 'exe': 'rustup',
+            \ 'args': ['run', 'nightly', 'cargo', 'clippy'],
+            \ 'errorformat': errorfmt,
+            \ }
+    else
+        return {
+            \ 'exe': 'cargo',
+            \ 'args': ['clippy'],
+            \ 'errorformat': errorfmt,
+            \ }
```
blueyed: Missing `endif` here, also reported by CI.

Eisfreak7: Fixed

autoload/neomake/makers/clippy.vim
```diff
@@ -0,0 +1,29 @@
...
+                 \ '%Inote: %m,'.
+                 \ '%-Z\ %#-->\ %f:%l:%c,'.
+                 \ '%I\ %#\= %t%*[^:]: %m,'.
+                 \ '%I\ %#|\ %#%\\^%\\+ %m,'.
+                 \ '%-G%s,'
+
+    " When rustup and a nightly toolchain is installed, that is used.
+    " Otherwise, the default cargo exectuable is used. If this is not part
+    " of a nightly rust, this will fail.
+    if executable('rustup') && system('rustup show | grep nightly | wc -l') >= 1
```
blueyed: You might want to cache the `system` call here.  See https://github.com/neomake/neomake/issues/620.

blueyed: Without knowing `rustup` this seems to be too broad/expensive: isn't there a `--version` call?
Can you provide an example output of `rustup show`, please?

Eisfreak7: I'll look into the caching.
rustup manages multiple installed versions of the rust toolchain, `--version` would only tell the version of rustup, which is not of interest hier. `rustup show` shows the installed toolchain versions, for example:
```
Default host: unknown-linux-gnu

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu (default)
nightly-x86_64-unknown-linux-gnu

active toolchain
----------------

stable-x86_64-unknown-linux-gnu (default)
rustc 1.11.0 (9b21dcd6a 2016-08-15)
```

It might well be that there is a more efficient way to do this, but I don't know it.

blueyed:
> How could I cache it? Create a new global variable? I'm not experienced with vimscript.

Use a script local var.  See https://github.com/neomake/neomake/pull/616#discussion_r79768387.

autoload/neomake/makers/clippy.vim
blueyed: What does `time rustup show` gives us? (i.e. how slow is it?)

blueyed: As for matching this, we should use at least `grep -q '^nightly-'` then, and no `wc -l`.
You get the error code of `system` in v:shell_error` btw.

Eisfreak7:
```
> time rustup show | grep -q '^nightly'
rustup show  0.04s user 0.01s system 93% cpu 0.050 total
grep --color=auto -q '^nightly'  0.00s user 0.00s system 0% cpu 0.019 total
```

Eisfreak7:
Currently I just use the installed version of rust when rustup is not installed, whichever it is.
Like I said above, nightly is needed and this will likely stay that way.
So maybe it would be a good idea to also check `rustc --version` in that case and give a useful error message if its not nightly? It would introduce another check, but it would be a lot more clear to the user why it is failing.

autoload/neomake/makers/clippy.vim
```diff
@@ -0,0 +1,38 @@
...
+    " Otherwise, the default cargo exectuable is used. If this is not part
+    " of a nightly rust, this will fail.
+    if !exists('s:rustup_has_nightly')
+        if !executable('rustup')
+            let s:rustup_has_nightly = 0
+        else
+            call system('rustup show | grep -q "^nightly-"')
+            let s:rustup_has_nightly = !v:shell_error
+        endif
+    endif
```
blueyed: Very good.
An improvement would be to set it to -1 in the outer scope, which allows to refresh the state when the script is resourced (then it gets set to -1 again).  Currently it will be set always.  Sorry for not suggesting this right away.
(I often re-source files, e.g. specific makers while changing them using `:Runtime` from https://github.com/tpope/vim-scriptease)

blueyed:
> So maybe it would be a good idea to also check `rustc --version` in that case and give a useful error message if its not nightly? It would introduce another check, but it would be a lot more clear to the user why it is failing.

Sounds good to me.
But it should not be too annoying - maybe just use `neomake#utils#LoudMessage`?

I could also imagine integrating this in `NeomakeInfo` and/or a general health check, see https://github.com/neomake/neomake/pull/568#issuecomment-245647966.

autoload/neomake/makers/clippy.vim
Eisfreak7: I'm not sure what you mean by that. Maybe like this:
```
let s:rustup_has_nightly = -1

function! neomake....
```
But that will never get executed, will it?

Eisfreak7:
I'm not sure why but somehow I can't get a message to show up. Even if I set the verbosity to 3 and `call neomake#utils#ErrorMessage("Test")` is the first line in the function.

autoload/neomake/makers/clippy.vim
blueyed: The init (to -1) will be done when the script is (re-)sourced.

blueyed:
That's strange.. it should show up then..
Try calling the function where you've defined it with `debug` as in `:debug call ...` and then step (`s` into it).  Use `n` for next etc.  Check `:help debug` if you do not know about it already.

Eisfreak7:
Didn't know that vim had an internal debugger, that's neat.

The error ended up being that I called neomake with
```
autocmd! BufWritePost *.rs silent Neomake! clippy
```
in my vimrc. Weired that the silenced command didn't make a noise.

Eisfreak7:
Accidently closed this

blueyed:
> Weired that the silenced command didn't make a noise.

That's the point of it I guess..?!
There is `unsilent` which you can use to work around this, but when you're using `silent` that it should be for a reason, and I dislike plugins overusing it.

autoload/neomake/makers/clippy.vim
```diff
@@ -0,0 +1,45 @@
...
+                 \ '%Inote: %m,'.
+                 \ '%-Z\ %#-->\ %f:%l:%c,'.
+                 \ '%I\ %#\= %t%*[^:]: %m,'.
+                 \ '%I\ %#|\ %#%\\^%\\+ %m,'.
+                 \ '%-G%s,'
+
+    " When rustup and a nightly toolchain is installed, that is used.
+    " Otherwise, the default cargo exectuable is used. If this is not part
+    " of a nightly rust, this will fail.
+    if !exists('s:rustup_has_nightly') || s:rustup_has_nightly == -1
```
blueyed: You can remove the `exists`-check here.

autoload/neomake/makers/clippy.vim
```diff
@@ -0,0 +1,45 @@
...
+        if !executable('rustup')
+            let s:rustup_has_nightly = 0
+            call system('rustc --version | grep -q "\-nightly"')
+            if v:shell_error
+                call neomake#utils#LoudMessage('Clippy requires a nightly rust installation.')
```
blueyed: Sorry, I was confusing by the levels/names.
That would show up only with level >= 2.
I think it should be at least 1 (`neomake#utils#QuietMessage`) or even `neomake#utils#ErrorMessage`.
Since it only gets displayed once, a error sounds good to me.
What do you think?

Eisfreak7: I agree, since neomake can't function as its supposed to that should be an error. (Although I think the levels are confusingly namend when a "Quiet" message can be "heard" and a "Loud" one can't.)

Eisfreak7:
> That's the point of it I guess..?!

Yes of course, I meant that question as a joke (which tends not to work over the internet). I had just copied and changed another autocmd from my vimrc without thinking too much about it, in the end the "weired" errors are always cause by some dumb behavior of the user (me).

> Good work still, left some comments.

Thanks, I appreciate the feedback.

blueyed:
> I meant that question as a joke (which tends not to work over the internet).

They invented smileys just for that. :)

> I can't figure out why travis is complaining, the changes seem correct to me (and are working).

Only "testnvim" is failing, checking it out.
Does it work locally for you?

Eisfreak7:
True :)

Works for me.

blueyed:
Here we go.

Thanks a lot - speaking as a potential Rust user in the future.. :)

Eisfreak7:
Thanks for taking the time to give feedback and merging this :)

rembo10:
Hey - if using a nightly toolchain from a specific date, this doesn't work. I have nightly-2016-10-19 installed, which doesn't get found with rustup run nightly. Not sure the best way to solve this. Maybe just use the default toolchain instead of fixing it to nightly? Or even just cargo clippy works

Eisfreak7:
Using the default toolchain doesn't work - the intended behavior is that I can compile and run with the compiler version of my choosing, but still run clippy with the latest nightly one, which clippy [recommends](https://github.com/Manishearth/rust-clippy). Why do you want to run it with a specific nightly version?
`carco clippy` is run when rustup isn't installed.

rembo10:
Ah, ok, thanks for the link. I don't need to use clippy with a specific nightly version - I'll just compile it against nightly.

The latest nightlies are breaking diesels macro imports, so I was using a specific nightly version until it was fixed, so I thought I had to use the same version with clippy

blueyed:
(JFI: I am unsubscribing.. please create a new issue if there's anything in here to be fixed)

PR #624, opened 2016-09-22
Improve readability of hlint output.

dkasak:
Currently, when the cursor is on a line with an error, neomake will echo errors
like these for hlint:

    hlint: Eta reduceFound:  mkListView window props = mkListViewEx window toLabel propsWhy not:  mkListView window = mkListViewEx window toLabel

This pull request changes it to this:

    hlint: Eta reduce | Found: mkListViewEx window present props = mkViewEx singleListBox window present props | Why not: mkListViewEx = mkViewEx singleListBox

In other words, it compresses whitespace while ensuring semantically different
parts of the output are separated, adding delimiters to make it more readable.


blueyed:
Do you feel like writing a test for this?
I think we should have tests for the makers itself, e.g. in `tests/ft/haskell.vader` in this case.

dkasak:
Okay, I'll take a look, but possibly won't be able to get around to it today.

dkasak:
Also, what kind of test did you have in mind? Something like passing a few different entry objects to `neomake#makers#ft#haskell#HlintEntryProcess` and checking whether it processed the output as expected?

blueyed:
Yes, a unit test for this seems to be good.

The testing framework is a bit rough/unstructured, but you should be able to run `make tests/ft/haskell.vader` in a separate terminal etc.

blueyed:
Check https://github.com/neomake/neomake/commit/1ac37c1679f272f0c1f334c96c8e15241dca6e5b.

dkasak:
I added a few unit tests that cover relevant cases that I could think of. By the way, your tests for 1ac37c1 don't work and pass the checks by accident because vimscript is awful. See my comments there.

blueyed:
> running make testnvim outputs a look of garbage and then just seems to hang there for a long time

Should not be the case.
Unfortunately we do not have Neovim in Docker yet, where it would be more reproducible.

tests/haskell.vader
```diff
@@ -0,0 +1,39 @@
+Include: _setup.vader
```
blueyed: I'd prefer it to be in `tests/haskell.vader` - might need Makefile adjustments etc though, so I can do it.

blueyed:
Great work, and thanks for finding a bug in the tests!
Let me know if you want to move the tests, otherwise I can do it.

tests/haskell.vader
dkasak: Sorry, prefer what to be in `tests/haskell.vader`?

blueyed: Oh.. I've meant to say `tests/ft/haskell.vader` - to keep maker related tests aside.

dkasak: Oh, right. I tried to put it under `tests/ft` initially but I couldn't get it to work on the first try. I wouldn't mind if you do it. :grinning:

blueyed: Yeah, seems to be an issue with Vader (https://github.com/junegunn/vader.vim/issues/88).
Can you move it to `ft_haskell.vader` then, please?

dkasak: No problem.

dkasak: Done.

blueyed:
Needs a rebase.
I went ahead and did it in https://github.com/neomake/neomake/pull/652, since it's due to conflicts caused by the then-still picky linter.

PR #627, opened 2016-09-23
Travis: vint-diff: ignore errors about missing abort attribute

blueyed:
There are still too many makers without this, and it is annoying having
to fix them then always.


blueyed:
Merged slightly different in https://github.com/neomake/neomake/commit/e0a7bd88934a0b90be7d731f29ae562041a2e6b5.

PR #636, opened 2016-09-24
Vim async, take 2

blueyed:
Followup of https://github.com/neomake/neomake/pull/575.

The main hack being https://github.com/neomake/neomake/pull/636/commits/650d4562aac94e4198e3f3b397fa786b582cf7e4 probably, but other crucial things were changed, e.g. using "raw" mode to behave the same as Neovim does - for non-buffered output.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -201,7 +201,7 @@ function! s:MakeJob(make_id, maker) abort
             call s:AddJobinfoForCurrentWin(jobinfo.id)
             call neomake#MakeHandler(job_id, split(system(program), '\r\?\n', 1), 'stdout')
             call neomake#MakeHandler(job_id, v:shell_error, 'exit')
-            let r = 0
+            let r = -1
         endif
     finally
         if exists('old_wd')
@@ -473,7 +473,7 @@ function! s:Make(options, ...) abort
             let maker.exit_callback = a:options.exit_callback
         endif
         let job_id = s:MakeJob(make_id, maker)
-        if job_id != 0
+        if job_id != -1
             call add(job_ids, job_id)
         endif
         " If we are serializing makers, stop after the first one. The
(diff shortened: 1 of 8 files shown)


autoload/neomake.vim
```diff
@@ -661,6 +708,31 @@ function! s:RegisterJobOutput(jobinfo, lines) abort
...
+    let jobinfo = s:jobs[channel_id]
+    call neomake#MakeHandler(channel_id, split(a:output, "\n", 1), 'stderr')
+endfunction
+
+function! neomake#MakeHandlerVimClose(channel) abort
+    call neomake#utils#DebugMessage('Channel has been closed: ' . a:channel)
+    let channel_id = s:GetChannelId(a:channel)
+    let jobinfo = s:jobs[channel_id]
+    call neomake#MakeHandler(channel_id, job_info(jobinfo.job)['exitval'], 'exit')
+endfunction
```
blueyed: This part really smells like it should be refactored.. even when it's just having one function and using a partial (see `:h Partial`) for stdout/stderr.
Also I am not sure if all the channel/job mapping is needed.
Messing with this should be easy now since tests are good again.. :)

blueyed:
Moved to main repo: https://github.com/neomake/neomake/pull/637

autoload/neomake.vim
saaguero: What do you have in mind for refactoring? I see this fine for now, as it's clearly handling Vim channel's callbacks, and acting as proxy to call MakeHandler.

I agree that the code (not just this Vim async part) should be refactored, I mean, the complete autoload/neomake.vim should be re designed / refactored... we can tackle this as an important step in another PR, but I'm all ears.

blueyed: There was a lot of duplicated code, also for the if/else with Neovim/Vim.
As for the latest part see https://github.com/neomake/neomake/pull/637/commits/642be494eff4ce15653bdbdd53bd8b2469556102.

PR #637, opened 2016-09-24
Vim async, take 2

blueyed:
Followup of https://github.com/neomake/neomake/pull/575.

The main hack being https://github.com/neomake/neomake/pull/636/commits/650d4562aac94e4198e3f3b397fa786b582cf7e4 probably, but other crucial things were changed, e.g. using "raw" mode to behave the same as Neovim does - for non-buffered output.

TODO
 - [x] refactor: https://github.com/neomake/neomake/pull/636#discussion_r80370397
 - [x] handle return value with sleep hack.  Add a test.

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -297,3 +297,8 @@ function! neomake#utils#redir(cmd) abort
     endtry
     return neomake_redir
 endfunction
+
+function! neomake#utils#ExpandArgs(args) abort
+    " Only expand those args that start with \ and a single %
+    call map(a:args, "v:val =~# '\\(^\\\\\\|^%$\\|^%[^%]\\)' ? expand(v:val) : v:val")
+endfunction
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
     - ENV=docker_test MAKE_ARGS=DOCKER_VIM=vim73
     - ENV=docker_test MAKE_ARGS=DOCKER_VIM=vim74-trusty
     - ENV=docker_test MAKE_ARGS=DOCKER_VIM=vim74-xenial
+    - ENV=docker_test MAKE_ARGS=DOCKER_VIM=vim8000
     - ENV=docker_test MAKE_ARGS=DOCKER_VIM=vim-master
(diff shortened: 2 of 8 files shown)


blueyed:
Looks like we should wait for https://groups.google.com/d/msg/vim_dev/LhXQJusQScM/_wV4u5y5AAAJ to be included.
Also the followup (https://groups.google.com/d/msg/vim_dev/LhXQJusQScM/t2jb38q9AAAJ) seems to be relevant regarding the escaping issue you've found on windows?!

saaguero:
While those two Vim patches might be useful, I think the changes in this PR so far are good to go.

Also, the user might have an older Vim version with job/channel support so we shouldn't be relying on those VIm patches to be included.

blueyed:
> Also, the user might have an older Vim version with job/channel support so we shouldn't be relying on those VIm patches to be included.

I'd like to use the patch level then to only enable async for Vim in that case.

It is like with the `buffer_output` stuff that I've just fixed: weird issues will show up.
I'll look into this PR tomorrow again, and _could_ imagine to ship it with the hack for now, but then it needs to be enabled not only for the tests I assume.

blueyed:
Ok, rebased and cleaned up.. I've squashed everything into your commit.

Please review it thoroughly.

autoload/neomake/utils.vim
```diff
@@ -297,3 +297,8 @@ function! neomake#utils#redir(cmd) abort
     endtry
     return neomake_redir
 endfunction
+
+function! neomake#utils#ExpandArgs(args) abort
+    " Only expand those args that start with \ and a single %
+    call map(a:args, "v:val =~# '\\(^\\\\\\|^%$\\|^%[^%]\\)' ? expand(v:val) : v:val")
```
bounceme: I don't understand how this regex works. maybe this needs to be a double quote string?

bounceme: Oh, now i see. I hate the string expressions in viml

blueyed:
Merging this now finally to celebrate 999 stars on GitHub.

saaguero:
Thanks for this addition and the great improvements you made to the code!

PR #640, opened 2016-09-25
FIX #639

### LINKED ISSUE #639: neomake#utils#GetSetting does not support settings that are lists
### Issue type

- Bug report
- Feature request

### Description

Settings overrides are things like `b:neomake_javascript_eslint_exe = 'myeslintbin'`
Right now the new `neomake#utils#GetSetting` introduced here https://github.com/neomake/neomake/commit/b49189af6c3b098f6b7df5fc8232b7729bdcc17e#commitcomment-19160638
checks that the variable is `!=# ''` for empty tests.

The settings override for args is actually `b:neomake_javascript_eslint_args = [ 'a', 'list' ]`
so the string check does not work and throws an error.

### Expected behavior

The test for the override config_var being empty should use `empty()`, which will handle both String and List types

### Steps to reproduce

- Override buffer args for a maker e.g. `let b:neomake_scss_sasslint_args = [ 'some', 'valid', 'args']`
- `:Neomake`

PR coming :)

davidosomething:
I've never written Vader tests so not sure if the assert works on lists

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -180,16 +180,20 @@ endfunction
 function! neomake#utils#GetSetting(key, maker, default, fts, bufnr) abort
   if len(a:fts)
     for ft in a:fts
+      " Look through the neomake setting override vars for a filetype maker,
+      " like neomake_scss_sasslint_exe (should be a string), and 
+      " neomake_scss_sasslint_args (should be a list)
       let config_var = 'neomake_'.ft.'_'.a:maker.name.'_'.a:key
       if has_key(g:, config_var)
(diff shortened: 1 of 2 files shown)


davidosomething:
Yea, @blueyed wanna help me with the tests? Not sure how to run it, getting 
```
(5/5) [EXECUTE] neomake#utils#GetSetting
    (5/5) [EXECUTE] (X) Vim(let):E706: Variable type mismatch for: g:neomake_myft_mymaker_serialize
      > /tmp/vgGu3lW/138, line 18
```

blueyed:
Thanks!

Issue #620, opened 2016-09-22
Running :Neomake in a Haskell stack project blocks the UI


dkasak:
# Issue

Running the `:Neomake` command blocks the UI for a few seconds for me. Setting `g:neomake_verbose = 3` reveals that this happens before the first line of neomake's output appears.

I've traced the problem to [this if block](https://github.com/neomake/neomake/blame/master/autoload/neomake/makers/ft/haskell.vim#L5). If this block is removed (and the following `elseif` changed to an `if`), the UI doesn't block after `:Neomake` and the command succeds for me (since I also have the makers installed outside the project).

The problem seems to be that this uses a `system()` call, which is synchronous, to check for a maker and the command that is run can be a little laggy. The proper way of doing this would be to check for the makers asynchronously, but I'm not very proficient with vimscript.

## Steps to reproduce

1. `cd` to a stack project.
2. Open a Haskell source file inside the project in vim.
3. Run `:Neomake`.

## Output of the ":NeomakeInfo" command

```
= Neomake debug information =                                                                                                                                             
Async support: 1                                                                                                                                                          
Current filetype: haskell                                                                                                                                                 
== Enabled makers ==                                                                                                                                                      
For the current filetype (with :Neomake): ['ghcmod', 'hlint']                                                                                                             
You can define g:neomake_haskell_enabled_makers to configure it (or b:neomake_haskell_enabled_makers).                                                                    
For the project (with :Neomake!): []                                                                                                                                      
You can define g:neomake_enabled_makers to configure it.                                                                                                                  
== Settings ==                                                                                                                                                            
g:neomaker_cpp_enabled_makers = ['gcc']                                                                                                                                   
g:neomaker_python_enabled_makers = ['flake8', 'python', 'pylint']                                                                                                         
g:neomake_verbose = 3                                                                                                                                                     
g:neomaker_python_pylint_maker = {'args': '--rcfile=/home/dkasak/.pylintrc'}                                                                                              
g:neomaker_haskell_enabled_makers = ['hlint']                                                                                                                   
g:neomaker_c_enabled_makers = ['splint']
```

## Versions

```
:version                                                                                                                                                                  
NVIM 0.1.5                                                                                                                                                                
Build type: RelWithDebInfo                                                                                                                                                
[...]

PR #644, opened 2016-09-25
cache Haskell maker availability checks

bonds:
attempts to address #620


autoload/neomake/makers/ft/haskell.vim
```diff
@@ -1,5 +1,6 @@
 
 function! neomake#makers#ft#haskell#MakerAvailable(command)
+    let s:stack_is_executable = 0
```
blueyed: Please init this outside of the function, and use -1 initially.
This way it gets reset when resourcing the file.

blueyed: But I think doing so for just `s:makers`, and then using `unlet! s:makers` to distinguish "not initialized" from "empty list, because nothing was found" should be enough.

The `executable` call is rather cheap, and you'll use `s:makers` then anyway.  `stack exec` should be avoided.

autoload/neomake/makers/ft/haskell.vim
```diff
@@ -1,5 +1,6 @@
 
 function! neomake#makers#ft#haskell#MakerAvailable(command)
```
blueyed: This function is script-local and should be renamed to `s:MakerAvailable` for clarity.

bonds:
@blueyed please take another look

dkasak:
Looks good to me. I tested your changes and it blocks shortly the first time and then doesn't block afterwards, as expected. :+1:

blueyed:
Awesome, thanks!

PR #646, opened 2016-09-26
Add R linting with the lintr package.

jimhester:


jimhester:
Sure I agree they should be documented, where would you like me to do so? I don't see any maker specific documentation.

Looks like the only other maker that sets global options is (https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/java.vim), but I don't think those options are externally documented either.

blueyed:
As for now, the best might be to add this information to the wiki (where the list of makers moved)?!

I think they should be in a dict `neomake_r_lintr_options` instead, and then `neomake#utils#GetSetting` (which could be simplified for this then) should get used in the maker, which would allow to override it in the usual way per buffer etc?!  What do you think?

jimhester:
using `neomake#utils#GetSetting` sounds fine, it looks like that function may need to be refactored a little, unless you can get a maker object from within a maker definition.

blueyed:
@jimhester 
Please rebase and report back any issues.

blueyed:
Also wrt to your last comment, of course.

blueyed:
@jimhester 
Ping.

blueyed:
Closing due to missing feedback.
Should be easy to pick up again for anyone interrested.

Issue #584, opened 2016-09-05
Add subtype or new type for "Style"


blueyed:
This is inspired by Syntastic, which has a subtype, typically used for "Style".

This could be used by linters like pep8.

I've experimented with just using `S` as the error type for this - not sure if it makes sense to have `E.S`, `W.S` and `I.S` in the end?!

PR #654, opened 2016-09-27
Add golint as Warning instead of Error

AitorATuin:
According to golint the linter is giving style errors which imho should be considered as warnings instead of errors.

```
Golint differs from gofmt. Gofmt reformats Go source code, whereas golint prints out style mistakes.

Golint differs from govet. Govet is concerned with correctness, whereas golint is concerned with coding style. Golint is in use at Google, and it seeks to match the accepted style of the open source Go project.

The suggestions made by golint are exactly that: suggestions.```

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -29,7 +29,7 @@ endfunction
 function! neomake#makers#ft#go#golint()
     return {
         \ 'errorformat':
-            \ '%f:%l:%c: %m,' .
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

What do you think about https://github.com/neomake/neomake/issues/584 in this context?

AitorATuin:
It could actually make sense, in that case it's not any more a matter of opinion since we could have our own category for those kind of errors/warnings. Everyone should be happy then :)

PR #659, opened 2016-09-28
fix the default error sign color

iamcco:
use `fg` and `fg#` for NeomakeErrorSignDefault not `bg` and `bg#`

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -235,8 +235,8 @@ function! neomake#signs#DefineHighlights() abort
 
     for [group, fgs] in items({
                 \ 'NeomakeErrorSign': [
-                \   neomake#utils#GetHighlight('Error', 'bg'),
-                \   neomake#utils#GetHighlight('Error', 'bg#')],
+                \   neomake#utils#GetHighlight('ErrorMsg', 'fg'),
+                \   neomake#utils#GetHighlight('ErrorMsg', 'fg#')],
                 \ 'NeomakeWarningSign': [
                 \   neomake#utils#GetHighlight('Todo', 'fg'),
                 \   neomake#utils#GetHighlight('Todo', 'fg#')],


autoload/neomake/signs.vim
```diff
@@ -235,8 +235,8 @@ function! neomake#signs#DefineHighlights() abort
 
     for [group, fgs] in items({
                 \ 'NeomakeErrorSign': [
-                \   neomake#utils#GetHighlight('Error', 'bg'),
-                \   neomake#utils#GetHighlight('Error', 'bg#')],
+                \   neomake#utils#GetHighlight('Error', 'fg'),
+                \   neomake#utils#GetHighlight('Error', 'fg#')],
```
blueyed: I see that `Error` is not even defined by default (using `nvim -u NONE` and then `:verb hi Error`), but `ErrorMsg` would be.
I based this on what the solarized theme gives me (where it is reversed):
> Error          xxx cterm=reverse ctermfg=1 guifg=White guibg=Red

So I think this PR should be changed to use `ErrorMsg` and `fg*`.

iamcco: I see, `Error` had been changed to `ErrorMsg`

blueyed:
Thanks, please squash it into one commit.
I could do it, but you appear to use different accounts and want to chose the correct one.. :)

iamcco:
OK, I am done.

blueyed:
Sorry, just noticed that `ErrorMsg` is reversed for me, too.. :/
Need to investigate.
Maybe we should just use an own definition (and not link it)..

blueyed:
What is `:verb hi ErrorMsg` for you?
And `:verb hi Error`?

blueyed:
Previously I had `NeomakeErrorSignDefault xxx ctermfg=1 ctermbg=21 guifg=White`, with your patch it is `NeomakeErrorSignDefault xxx ctermbg=21 guifg=Red`.
What colorscheme are you using?
(I have a customized solarized one)

iamcco:
with vim-gotham colorscheme and this patch:
`:verb hi Error` is `Error          xxx ctermfg=1 ctermbg=8 guifg=#c23127 guibg=#11151c`
`:verb hi ErrorMsg` is `ErrorMsg       xxx ctermfg=1 ctermbg=8 guifg=#c23127 guibg=#11151c`
`:verb hi NeomakeErrorSignDefault` is `NeomakeErrorSignDefault xxx ctermfg=1 ctermbg=8 guifg=#c23127 guibg=#11151c`

with vim-gotham colorscheme and no patch:
`:verb hi Error` is `Error          xxx ctermfg=1 ctermbg=8 guifg=#c23127 guibg=#11151c`
`:verb hi ErrorMsg` is 'ErrorMsg       xxx ctermfg=1 ctermbg=8 guifg=#c23127 guibg=#11151c`
`:verb hi NeomakeErrorSignDefault` is `NeomakeErrorSignDefault xxx ctermfg=8 ctermbg=8 guifg=#11151c guibg=#11151c`

so with no patch i can't see the error sign

iamcco:
use an own definition is a good way

blueyed:
TL;DR: please try https://github.com/neomake/neomake/pull/667.

This just left for reference:
Looked into it a bit more now: the typical use for themes is to have (their) "red" as the bg for Error ("Error" appears to only be defined when syntax highlighting is active, that's why I missed it with `nvim -u NONE` and no file).

Using a default of `ctermfg=1 guifg=red` would remove the benefit of picking up customized highlights from color schemes.

Therefore I think that gotham theme should explicitly define `NeomakeErrorSign`, and/or you do it in your config:
```vim
    augroup my_error_signs
        au!
        autocmd ColorScheme * hi NeomakeErrorSign ctermfg=red guifg=red
    augroup END
```

For reference, Syntastic just does `highlight link SyntasticErrorSign error`, but we want to use the bg of the SignColumn here.

iamcco:
I get it, thanks for your work

blueyed:
Have you tried #667?

iamcco:
yes, it works well 👍

Issue #650, opened 2016-09-27
Vim: maker completion is broken with execute()


saaguero:
### Issue type
- Bug report

After a git bisect session, this seems to be introduced by https://github.com/neomake/neomake/commit/2b67775b8bad9e1c893abe3c717085f53c414d86

### Description
-  Available makers are not being shown when using the Neomake command for a supported filetype.

### Expected behavior
- I can see the list of available makers for the current filetype when pressing the key `Tab` after Neomake command.

### Steps to reproduce

- Open a file with available makers (ie: a sh script)
- Use the command `:Neomake` and then press the key `Tab` for getting a list of the available Makers to be used.

### Output of the ":verb NeomakeInfo" command

<details>
```
#### Neomake debug information
Async support: 0
Current filetype: sh

##### Enabled makers
For the current filetype (with :Neomake): ['sh', 'shellcheck']
NOTE: you can define g:neomake_sh_enabled_makers to configure it (or b:neomake_sh_enabled_makers).
For the project (with :Neomake!): []
NOTE: you can define g:neomake_enabled_makers to configure it.

##### Settings
g:neomake_logfile = '/home/saaguero/temp/neomake.log'
g:neomake_verbose = 3

#### :version
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Sep 20 2016 22:43:53)
Included patches: 1-6
Compiled by saaguero@saaguero-mobl2.amr.corp.intel.com
Huge version without GUI.  Features included (+) or not (-):
+acl             +cscope          -footer          +menu            +num64           +startuptime     +virtualedit
+arabic          +cursorbind      +fork()          +mksession       +packages        +statusline      +visual
+autocmd         +cursorshape     +gettext         +modify_fname    +path_extra      -sun_workshop    +visualextra
-balloon_eval    +dialog_con      -hangul_input    +mouse           -perl            +syntax          +viminfo
-browse          +diff            +iconv           -mouseshape      +persistent_undo +tag_binary      +vreplace
++builtin_terms  +digraphs        +insert_expand   +mouse_dec       +postscript      +tag_old_static  +wildignore
+byte_offset     -dnd             +job             -mouse_gpm       +printer         -tag_any_white   +wildmenu
+channel         -ebcdic          +jumplist        -mouse_jsbterm   +profile         -tcl             +windows
+cindent         +emacs_tags      +keymap          +mouse_netterm   +python          +termguicolors   +writebackup
[...]

PR #661, opened 2016-09-29
Vim execute does not work in some cases, stick with redir

saaguero:
Fixes https://github.com/neomake/neomake/issues/650.

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -252,9 +252,6 @@ function! neomake#utils#CompressWhitespace(entry) abort
 endfunction
 
 function! neomake#utils#redir(cmd) abort
-    if exists('*execute')
-        return execute(a:cmd)
-    endif
     if type(a:cmd) == type([])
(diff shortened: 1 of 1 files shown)


blueyed:
Not convinced.. have to look into #650 actually and would rather see a failing test.. :)

saaguero:
Well, that's  the interesting part, if  you use the new Vim's execute function in a test, it doesn't fail. We can dig in but I much rather prefer not breaking the master with these changes (ie: the use of Vim's execute).

Also, there isn't benefit in the way we're using it. It's better to maintain less code.

blueyed:
I can reproduce the issue using Vim manually and would expect it to fail also in an automated test.
Have you really written one to see that it succeeds there?  Then please provide the patch.

Anyway, as per https://github.com/neomake/neomake/issues/650#issuecomment-250612364 I think Neovim is still first-class citizen here, and it works there as expected.

A PR with a failing test would be always more convincing, of course.

Issue #600, opened 2016-09-08
Fix errors reported by vimhelplint


blueyed:
Mostly about line-width, but is also has some hidden treasures.

https://travis-ci.org/neomake/neomake/jobs/158598921#L335

davidosomething:
opened PR #685

PR #662, opened 2016-09-29
clear up some confusing info in the docs

davidosomething:


doc/neomake.txt
```diff
@@ -88,12 +88,16 @@ All 'args' will be |expand()|ed: >
...
+The function must return a string, and the function will be called with no
+arguments.
 
-Running "|:Neomake!| lint" from file.txt would cause "lint /path/to/file.txt" to
-be run.
+In the above example, running "|:Neomake!| lint" from file.txt would cause
+"lint /path/to/file.txt" to be run. Do note that this is atypical usage and is
+provided as an example of args expansion only, since "|:Neomake!|" commands
+typically don't require the filename. The preferred method of running Neomake
+against the current file is "|:Neomake|" without the exclamation mark.
```
blueyed: Well, this somehow manifests the arg usage in the doc.
And it's not really "args expansion" but using "lint" as the maker, isn't it?

doc/neomake.txt
```diff
@@ -103,10 +107,14 @@ usable on an individual file, you should also include the filtype in the name: >
...
         \ 'errorformat': '%f:%l:%c: %m',
         \ }
-    " Use this maker from a c file:
+
+    " Run this maker against the open file:
     :Neomake lint
-    " Or use it on the whole project:
+    " This would run "lint --option x myfile.c"
+
+    " Or use it on the whole project, executing the command from Vim's cwd
```
blueyed: s/`cwd`/current working directory/

doc/neomake.txt
davidosomething: This example in general was a poor choice since in involves misusing `:Neomake!`

What do you think about removing it entirely?

The fact that args are expanded is already mentioned earlier https://github.com/davidosomething/neomake/blob/cf62376ea2c7ab6d7f975971b15371115aba57cc/doc/neomake.txt#L85-L89

blueyed: > What do you think about removing it entirely?

Sounds good!

davidosomething: okay, removed the arg-expansion-abusive example

doc/neomake.txt
davidosomething: updated

doc/neomake.txt
```diff
@@ -31,13 +31,13 @@ fantastic vim plugins such as syntastic and dispatch.
 ==============================================================================
 2. Commands                                                   *neomake-commands*
 
-                                                         *:Neomake* *:NeomakeFile*
+                                                       *:Neomake* *:NeomakeFile*
```
blueyed: Well, this is wrong when being concealed.. the `*` will be removed and then this is less indented than the line above.
Or is this intentional?
Guess it might be good as-is for now (since the width is too wide anyway), see https://github.com/neomake/neomake/issues/600.

davidosomething: I intentionally moved it to line up with the other helptag names -- they're right aligned to the same column with noconceal

blueyed: I get `conceallevel=2` for help files.. not for you?

blueyed: But I can see that it's bad when being viewed raw.
Maybe there should be the same level of conceal chars on each line then, i.e. split `:Neomake! :NeomakeProject` into two lines?

davidosomething: I turned it off in my vimrc, I'll leave it up to you since it's just a matter of personal style

blueyed:
Thanks!
Adjusted and merged.

Are you interested in tackling https://github.com/neomake/neomake/issues/600 btw? :)

davidosomething:
sure

PR #665, opened 2016-09-30
Allow stylelint in .scss

rstacruz:
Stylelint (http://stylelint.io/) supports CSS, SCSS, and Less. In fact, it has more [rules](http://stylelint.io/user-guide/rules/), more [published configs](https://www.npmjs.com/search?q=stylelint-config), and even more GitHub stars compared to sass-lint—I would consider it a better-supported project than the alternatives.

This adds support for Stylelint in SCSS.

### DIFF
--- a/autoload/neomake/makers/ft/scss.vim
+++ b/autoload/neomake/makers/ft/scss.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#scss#EnabledMakers() abort
-    return executable('sass-lint') ? ['sasslint'] : ['scsslint']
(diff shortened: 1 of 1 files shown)


rstacruz:
Also, I think `executable('...')` isn't a very good check (but does the job for now); the better thing to do is to check the current project for a `.stylelintrc`, `stylelint.config.js`, or package.json's `stylelint`.

partounian:
You should extend this to support Less 👍

blueyed:
Thanks!
We can do Less in a new PR.
@partounian 
Feel free to submit one.

Please add stylelint to the wiki for scss.

PR #668, opened 2016-10-02
Call neomake#signs#DefineHighlights on VimEnter

blueyed:
Doing in the plugin file directly seems unclean, but only doing it for
ColorScheme will not be triggered for the `:colorscheme` in a vimrc.

### DIFF
--- a/plugin/neomake.vim
+++ b/plugin/neomake.vim
@@ -17,8 +17,7 @@ augroup neomake
   au!
   au WinEnter,CursorHold * call neomake#ProcessCurrentWindow()
   au CursorMoved * call neomake#CursorMoved()
-  au ColorScheme * call neomake#signs#DefineHighlights()
+  au ColorScheme,VimEnter * call neomake#signs#DefineHighlights()
 augroup END
-call neomake#signs#DefineHighlights()
 
 " vim: sw=2 et


dojoteef:
@blueyed I just noticed that this change broke some setup in my .vimrc file. According to the vim documentation VimEnter only fires after all other setup is done.

```
VimEnter                        After doing all the startup stuff, including
                                loading vimrc files, executing the "-c cmd"
                                arguments, creating all windows and loading
                                the buffers in them.
```

So anyone who has code executing on BufEnter, WinEnter, BufWinEnter, etc can no longer rely on the signs being defined. That said, I simply fixed my issue by directly calling `neomake#signs#DefineSigns()`, but that does not feel like the best approach.

blueyed:
@dojoteef 
I see.
But I think it's bad to trigger the autoload here always.. OTOH that gets done through VimEnter then, too.

I would like to do this just before signs get displayed even.
What is your use case?  Why do they have to be defined?  Can you link to your code?

dojoteef:
@blueyed Thanks for the quick response!

I took a second look at my reasoning for needing that code and realized there is a better way, so I removed my usage of it. The availability of both the `g:neomake_*_sign` options and `neomake#sign#Redefine*Sign` seem like they are sufficient to address any usage concerns.

PR #669, opened 2016-10-02
Adds c clang-tools to cpp makers

ebenoist:
These checkers should be enabled by default for C++ files as well as C files


blueyed:
Thanks!

PR #676, opened 2016-10-03
Added Lint.jl maker for Julia.

joshbode:
Added a Julia maker based on [Lint.jl](https://github.com/tonyhffong/Lint.jl)

### DIFF
--- a/autoload/neomake/makers/ft/julia.vim
+++ b/autoload/neomake/makers/ft/julia.vim
@@ -1,0 +1,22 @@
+function! neomake#makers#ft#julia#EnabledMakers() abort
+    return ['julia']
+endfunction
+
+function! neomake#makers#ft#julia#julia() abort
+    return {
+\       'errorformat': '%f:%l %t%*[^ ] %m',
+\       'args': ['-e', '
+\           try
+\               using Lint
+\           catch
+\               println("$(basename(ARGS[1])):1 E999 Install Lint.jl: Pkg.add(\\"Lint\\")");
+\               exit(1)
+\           end;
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!
Please add it to the list of makers at https://github.com/neomake/neomake/wiki/Makers, and then I'll merge it.

blueyed:
Thanks!

tobanw:
I can't get this to work. I'm using Neovim and Lint.jl works fine when I use it in the julia REPL.

Makers for other filetypes will show the errors in the location list (`:lopen`), but calling `:Neomake` on a julia file doesn't put anything in the location list and just gives this message in the status line:
```
Neomake: line: completed with exit code 1.
```

joshbode:
Hmm - sorry, not sure what happened there!
I'm sure it had been working, but when I checked it wasn't even working for me.
PR #738 should fix it.

joshbode:
By the way, I checked by rolling back to cf14e7a it worked back then ;)
So there must have been some kind of change elsewhere in how command quoting works.

blueyed:
Interesting. Would be good to have this bisected.
Likely https://github.com/neomake/neomake/commit/0d1b248883d5b22f18353177f75ee28f73bb5a0a, skipping the expand().

joshbode:
That's the one - prior to 0d1b248 the original escaping still worked.

blueyed:
@joshbode 
Thanks!
After all I think it's good to not `expand()` everything, but it might have broken other makers.. :/
(that's why I think tests for makers are important, too)

Issue #287, opened 2016-01-24
tern_lint support


yotamN:
Hi, I think implementing [tern_lint](https://github.com/angelozerr/tern-lint) will be great, there isn't implementation of this in neo/vim right now and it will really help for me and I'm sure a lot others. Sadly I don't know vimscript to add this myself.

davidosomething:
I just tried defining this as a maker following syntastic's approach -- tern-lint appears to only run on a project-basis (based on .ternproject's location, and this is how tern works since it checks all modules). It is also incredibly slow.

Will highly recommend integrating this as part of yout continuous integration process (or manually) instead of a neomake maker linter since the job will rarely finish fast enough to be useful.

see also: https://github.com/scrooloose/syntastic/commit/e8a06ea247905024870aa9ba9241be1992845477

davidosomething:
tern lint has problems right now -- the line numbers don't work correctly

I have an open PR at tern-lint https://github.com/angelozerr/tern-lint/pull/74 that formats its JSON output into something easily usable by Neomake.
The maker config is https://github.com/davidosomething/dotfiles/blob/master/vim/plugin/plug-neomake.vim#L98-L105

POC test screenshot, you can see the line number doesn't match up:
http://screencast.com/t/txrtOniIF

davidosomething:
Looking at syntastic's implementation, I'll try it with byte2line later https://github.com/scrooloose/syntastic/blob/master/autoload/syntastic/preprocess.vim#L369-L374

davidosomething:
latest update: my updates for tern-lint support were merged https://github.com/angelozerr/tern-lint/pull/74 -- waiting on an [npm publish](https://github.com/angelozerr/tern-lint/issues/76)

davidosomething:
@yotamN see #686 for more info, but like @bounceme I don't think it's worth moving forward

Issue #543, opened 2016-08-09
Neomake "IDE" (lint as you type)


dojoteef:
Hi! I'm really loving neomake's asynchronous mode. I liked it so much that I spent some time implementing an [IDE like feature](https://github.com/dojoteef/dotfiles/blob/ecafedff904df58222712ae06eb5a232c450781d/link/.vimrc#L463) in my vimrc which will lint code as you type. I'm not sure how extensible or flexible it is since I have to hack around a bit to make it work (though my approach doesn't change neomake in anyway). I'd love to hear some feedback and see if this is something that neomake could try to support natively so the ugly hacks would not be necessary (especially if changes in neomake break the code).

NOTE: I got inspiration for this from the following [blog post](https://www.gregjs.com/vim/2015/lint-as-you-type-with-neovim-and-neomake/). The approach in the blog post had a major caveat, it would write your file to disk every time you left insert mode so the neomake makers could run over the code. I wanted an approach that would work even while you were still insert mode and would not require saving you changes to the file you were editing. That's what my code supports.

blueyed:
Wow, very decent job! :cherries: 

I somehow like that you're using hidden API already (which was added for tests) with `neomake#has_async_support`.. :)

Since I cannot leave comments directly in the code/commit, I'll leave some here:

- s:neomake_windo (https://github.com/dojoteef/dotfiles/blob/ecafedff904df58222712ae06eb5a232c450781d/link/.vimrc#L662): check out https://github.com/vim-scripts/ArgsAndMore/blob/master/autoload/ArgsAndMore.vim#L127 (you can probably make it easier with `noautocmd`?!).
- you should use a prefix with `b:lastchangedtick`

Altogether, I've only skimmed it, but you certainly should move it into its own repo/plugin (especially given that it takes the most part of your vimrc currently).

Given all  the code I assume that it is pretty bad with regard to performance, isn't it?

dojoteef:
> Wow, very decent job! 🍒

Thanks! It's my first attempt at any real vimscript and I'm pretty happy with the end result.

> I somehow like that you're using hidden API already (which was added for tests) with `neomake#has_async_support`.. :)

Yeah I had to go digging through the neomake source code to figure out how best to implement the functionality needed for this feature.

[...]

PR #686, opened 2016-10-06
sort javascript makers, add ternlint maker

davidosomething:
- FIX: #287 -- add tern-lint support as the `ternlint` maker
- CHANGED: sort javascript makers alphabetically (there's no logic currently)
- CHANGED: add `ternlint` maker to default makers

blocked by https://github.com/angelozerr/tern-lint/issues/76

### DIFF
--- a/autoload/neomake/makers/ft/javascript.vim
+++ b/autoload/neomake/makers/ft/javascript.vim
@@ -1,24 +1,14 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#ft#javascript#EnabledMakers()
-    return ['jshint', 'jscs', 'eslint']
+function! neomake#makers#ft#javascript#EnabledMakers() abort
+    return ['eslint', 'jshint']
 endfunction
 
-function! neomake#makers#ft#javascript#jshint()
-    return {
-        \ 'args': ['--verbose'],
-        \ 'errorformat': '%A%f: line %l\, col %v\, %m \(%t%*\d\)',
-        \ }
-endfunction
+" ============================================================================
+" Makers sorted alphabetically
+" ============================================================================
 
-function! neomake#makers#ft#javascript#jscs()
-    return {
-        \ 'args': ['--no-colors', '--reporter', 'inline'],
-        \ 'errorformat': '%E%f: line %l\, col %c\, %m',
-        \ }
-endfunction
-
-function! neomake#makers#ft#javascript#eslint()
+function! neomake#makers#ft#javascript#eslint() abort
     return {
         \ 'args': ['-f', 'compact'],
         \ 'errorformat': '%E%f: line %l\, col %c\, Error - %m,' .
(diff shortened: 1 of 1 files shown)


bounceme:
Honestly, i consider tern-lint as dangerous. To have any kind of accuracy it needs to look at all the dependencies in your project. This is absolutely impractical in most projects and the process will continue for minutes/ until the node stack is blown. Consider how easy people will forget that the process is running? what about running synchronously? Many people don't seem to know about C-c and would just kill vim losing their work. I would also have some concern about the sheer size of output, massive in many cases ( because the dependency parse ). could viml actually handle it all?

davidosomething:
**I'm going to close this PR** because  
- tern-lint is indeed not good enough to be useful
- neomake needs more features to use it well

blocked by #428, #543, https://github.com/angelozerr/tern-lint/pull/72

I have a test case here, where I run .tern-lint on 836k of JS split amongst 124 files
https://gist.github.com/davidosomething/1b6e06bc2d263547cd9a4ce4ac13214e  

- There are a ton of false positives, but that's something to update in tern-lint
- The total time is 1.2seconds to run tern-lint (`node_modules` and vendor stuff excluded) so speed is not a factor when configured correctly
- stack size and files opened IS a factor for large projects

I'll probably update this in the far future to not enable tern by default (and even though it is on, user would need to have both npm installed tern-lint globally and configured a .tern-project to run it). Then users would have to manually use `:Neomake tern` to run it.

Some notes addressing your points above and if anyone wants to take over this:

- tern by default does not run with the modules plugin so will not look at your dependencies
- with the various modules plugins on, you can also specify `dontLoad` to not load modules (in which case only files included in loadEagerly in `.tern-project` will be checked)
    - A good addition for tern-lint is as you've done in https://github.com/angelozerr/tern-lint/pull/72/files to populate loadEagerly based on the CLI file arg, otherwise it runs against all files loaded eagerly in the tern project or already loaded in the tern server.

- the tern server closes itself after 5 minutes be default
- when running tern-lint synchronously (i.e. the case of neomake in Vim7.4) it will indeed block the vim process -- as will all other make programs, and the person using neomake in vim7.4 should know about that already
- if the node stack is blown (i have yet to encounter this), the process will probably exit with error, in which case neovim just reports an error with the maker
- if the files opened limit is reached (OS limit) everything will shit the bed (problem to address in tern-lint, again I like PR 72 in tern-lint)

- re:https://github.com/angelozerr/tern-lint/pull/72 @angelozerr was under the impression that you'd be passing in a stream as you type (on-the-fly linting as mentioned in other neomake PRs) to the tern-lint cli. This is not the case with neomake -- it passes in a filename when run with `:Neomake`, and tern-lint cli will run it against the file in the filesystem. No point in running neomake on `InsertChanged` (until #543)
    - The server approach mentioned is a good way to go about it -- depends on #428

PR #695, opened 2016-10-10
Typo in examplecode for RedefineErrorSign

noctux:
The example code in the documentation for `neomake#signs#RedefineErrorSign` calls said function as `neomake#utils#RedefineErrorSign` which is not defined (probably it was moved/renamed?). This commit fixes this discrepancy.


blueyed:
Thanks!
Seems to have been forgotten in c998504.

PR #704, opened 2016-10-13
improve Haskell makers

bonds:
- add support for non-cabal projects to hdevtools maker
- add async support to avoid blocking on system calls

### DIFF
--- a/autoload/neomake/makers/ft/haskell.vim
+++ b/autoload/neomake/makers/ft/haskell.vim
@@ -1,55 +1,61 @@
 unlet! s:makers
-
-function! s:MakerAvailable(command) abort
-    " stack may be able to find a maker binary that's not on the normal path
-    " so check for that first
-    if executable('stack')
-        " run the maker command using stack to see whether stack can find it
-        " use the help flag to run the maker command without doing anything
-        if system('stack exec -- ' . a:command . ' --help > /dev/null 2>&1; echo $?') == 0
-            return 1
-        else " if stack cannot find the maker command, its not available anywhere
-            return 0
-        endif
-    elseif executable(a:command) " stack isn't available, so check for the maker binary directly
-        return 1
-    else
-        return 0
-    endif
-endfunction
+unlet! s:uses_cabal
 
 function! neomake#makers#ft#haskell#EnabledMakers() abort
-    " cache whether each maker is available, to avoid lots of (UI blocking) system calls...the user must restart vim if a maker's availability changes
     if !exists('s:makers')
+        " cache whether each maker is available, to avoid lots of (UI blocking)
+        " system calls
+        " TODO: figure out how to do all this configuration async instead of
+        " caching it--that would allow the user to change directories from
+        " within vim and recalculate maker availability without restarting vim
         let commands = ['ghc-mod', 'hdevtools', 'hlint', 'liquid']
         let s:makers = []
+        let s:jobs = []
         for command in commands
-            if s:MakerAvailable(command)
+            " stack may be able to find a maker binary that's not on the normal
+            " path so check for that first
+            if executable('stack')
+                " run the maker command using stack to see whether stack can
+                " find it use the help flag to run the maker command without
+                " doing anything
+                let stack_command = [
+                      \   'stack'
+                      \ , 'exec'
+                      \ , '--'
+                      \ , command
+                      \ , '--help'
+                      \ ]
+                if has('nvim')
+                    let job_id = jobstart(
+                        \ stack_command,
+                        \ { 'command': command
+                        \ , 'on_exit': function('s:CheckStackMakerAsync')
+                        \ })
+                    if job_id > 0
+                        call add(s:jobs, job_id)
+                    endif
+                else
+                    call extend(stack_command, ['> /dev/null 2>&1;', 'echo $?'])
+                    if system(join(stack_command, ' ')) == 0
+                        call add(s:makers, substitute(command, '-', '', 'g'))
+                    endif
+                endif
+            elseif executable(command) " no stack bin so check for maker bin
                 call add(s:makers, substitute(command, '-', '', 'g'))
             endif
         endfor
+        if has('nvim')
+            call jobwait(s:jobs)
+        endif
     endif
     return s:makers
 endfunction
 
-function! s:TryStack(maker) abort
-    if executable('stack')
-        if !has_key(a:maker, 'stackexecargs')
-            let a:maker['stackexecargs'] = []
-        endif
-        let a:maker['args'] = ['--verbosity', 'silent', 'exec'] + a:maker['stackexecargs'] + ['--'] + [a:maker['exe']] + a:maker['args']
-        let a:maker['exe'] = 'stack'
-    endif
-    return a:maker
-endfunction
-
 function! neomake#makers#ft#haskell#hdevtools() abort
-    let mapexpr = 'substitute(substitute(v:val, " \\{2,\\}", " ", "g"), "`", "''", "g")'
-    return s:TryStack({
+    let params = {
         \ 'exe': 'hdevtools',
         \ 'args': ['check', '-g-Wall'],
-        \ 'stackexecargs': ['--no-ghc-package-path'],
-        \ 'mapexpr': mapexpr,
+        \ 'mapexpr': s:CleanUpSpaceAndBackticks(),
         \ 'errorformat':
             \ '%-Z %#,'.
             \ '%W%f:%l:%v: Warning: %m,'.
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/haskell.vim
```diff
@@ -1,55 +1,61 @@
...
+                " find it use the help flag to run the maker command without
+                " doing anything
+                let stack_command = [
+                      \   'stack'
+                      \ , 'exec'
+                      \ , '--'
+                      \ , command
+                      \ , '--help'
+                      \ ]
+                if has('nvim')
```
blueyed: I think this should rather use a (to be done/enhanced) abstraction in Neomake, based on `neomake#has_async_support` at least, so Vim would be covered (later), too.
What about using something like / based on `neomake#Sh` here, which would need some flag to be silent for the usual processing then?  This certainly would involve some internal refactoring..

autoload/neomake/makers/ft/haskell.vim
```diff
@@ -59,7 +65,36 @@ function! neomake#makers#ft#haskell#hdevtools() abort
...
+              \   'ls '
+              \ . '$(stack --verbosity silent path --project-root)'
+              \ . '/*.cabal'
+        if has('nvim')
+            let job_id = jobstart(
+                  \   command
+                  \ , {'on_exit': function('s:CheckHdevPPAsync')}
+                  \ )
+            if job_id > 0
+                call jobwait([job_id])
```
blueyed: What is the benefit of using a job here, when waiting for it just below?
Callbacks for other jobs might happen, but otherwise it would still block, no?

blueyed: But `jobwait` blocks, doesn't it?

autoload/neomake/makers/ft/haskell.vim
blueyed: It would only be something in that regard, not exactly this.
For `jobwait` there would be a separate function then, and it looks like for Vim that would need to wait manually, using `job_status()` and `sleep` then.
It was just a suggestion to make it work also in Vim, and an opportunity for some refactoring / simplification.  I can see that this might be asked too much though definitely.. ;)

bonds:
I don't think I want to reimplement async support on Vim, though I'm happy to take advantage of it if the Neomake project does it. I suggest my patch makes the experience better for Neovim users without hurting the experience for Vim users. And I did my best to make it easy to follow, for the next contributor. It should also be easy to add Vim support once I get my hands on Vim 8.0 (I use OpenBSD stable, so after March perhaps?), which I plan to do.

I'm also happy to use the Neovim async abstraction once its finished, feel free to bug me about it then. :)

blueyed:
I see.. Neomake uses Vim's jobs already (but it has still issues), and therefore there is an internal abstraction for this already - that's what I've meant to use.

Regarding this PR then, I'd still think the 2nd `jobstart`/`jobwait` is pointless (as there is no outer loop)?!

bonds:
You are right. I changed the command to 'sleep 100' and it locked the UI for that time. I was assuming the UI wouldn't block while waiting for that code to finish, as long as it was running as a job. Hmm. In that case, what I think I really want is to move the logic into the maker itself, as the UI doesn't block when the maker is running. And the makers list would just be a static list again.

All we get with this patch is the maker detection stuff runs in parallel instead of in series. Its noticeably faster on my 7-year old laptop, but I'm shooting for the 'sleep 100' not blocking the UI. :) Alright, stay tuned for another patch.

bonds:
I haven't come up with a good way to config the haskell makers async, so I'm going back to the strategy of caching the system call results for now. Perhaps it would make sense to run the config async at launch and when changing directories, then jobwait until the results are available. That would at least reduce the amount of jobwait related UI blocking that would happen by the time you get around to running a maker.

Not sure its worth the added complexity at the end of the day. The ideal solution it to shove all that config work into the job that runs the maker, or to have some facility for making maker jobs dependent on another job, that way there's *no* UI blocking during configuring nor while running the makers.

So...the main benefit of the current patch is that it fixes the hdevtools behaviour. Secondary benefit is a bit less UI blocking at startup if you're on Neovim.

blueyed:
Ok, let's merge this.
Can you squash it please, or should I do it using the GitHub UI?
There is a error from vint ("autoload/neomake/makers/ft/haskell.vim:149:39: Undefined variable: self (see :help E738)", but self comes from `jobstart()` in Neovim).

bonds:
@blueyed squashed ... what do you make of the vimlint error?

Alok:
@bonds would this fix the issue of the `hdevtools` maker reporting `module not found` for globally installed packages?

bonds:
yes, at least, it does for me, please try it and report your experience!

Alok:
It fixed the hdevtools issue for me.

blueyed:
> what do you make of the vimlint error?

Needs to be ignored probably?!  Search for `@vimlint` - this gets done in other places already.
Can look into doing it myself when reviewing this again.

bonds:
fixed

autoload/neomake/makers/ft/haskell.vim
```diff
@@ -1,55 +1,63 @@
...
+                        \ { 'command': command
+                        \ , 'on_exit': function('s:CheckStackMakerAsync')
+                        \ })
+                    if job_id > 0
+                        call add(s:jobs, job_id)
+                    endif
+                else
+                    let stack_command_str =
+                        \ join(stack_command, ' ') .
+                        \ '> /dev/null 2>&1; echo $?'
```
blueyed: This looks like it might want to be a list instead?!

autoload/neomake/makers/ft/haskell.vim
```diff
@@ -59,7 +67,28 @@ function! neomake#makers#ft#haskell#hdevtools() abort
...
+    " a cabal file
+    if !exists('s:uses_cabal')
+        let s:uses_cabal = 0
+        if executable('stack')
+            if system(
+                    \   'ls '
+                    \ . '$(stack --verbosity silent path --project-root)'
+                    \ . '/*.cabal'
+                    \ . ' > /dev/null 2>&1; echo $?'
+                    \ ) == 0
```
blueyed: You could get the output from `stack --verbosity silent path --project-root` using `system()` here and the use this with Vim's `glob()` probably.

autoload/neomake/makers/ft/haskell.vim
```diff
@@ -103,15 +132,47 @@ function! neomake#makers#ft#haskell#hlint() abort
...
+    if executable('stack')
+        if !has_key(a:maker, 'stackexecargs')
+            let a:maker['stackexecargs'] = []
+        endif
+        let a:maker['args'] =
+            \   ['--verbosity', 'silent', 'exec']
+            \ + a:maker['stackexecargs']
+            \ + ['--']
+            \ + [a:maker['exe']]
+            \ + a:maker['args']
```
blueyed: This could at least join the `['--']` and `[a:maker['exe']`.

blueyed:
Left some comments, but I am generally :+1:.
Let me know if it should be merged as is, or otherwise push some fixups, preferable into a single commit already, that we could merge then.

autoload/neomake/makers/ft/haskell.vim
bonds: I'm not sure I follow. Are you saying I should extend the stack_command list before converting the whole thing to a string?

autoload/neomake/makers/ft/haskell.vim
bonds: Got it, will do.

autoload/neomake/makers/ft/haskell.vim
bonds: will do

autoload/neomake/makers/ft/haskell.vim
blueyed: You can pass in commands as lists, but might not work with older Vim etc - so feel free to leave it as is.

bonds: I just tried passing the command as a list, but I think that caused some escaping of the piping characters, which broke them, so I'm going to leave this as-is.

autoload/neomake/makers/ft/haskell.vim
```diff
@@ -1,55 +1,61 @@
...
+                    let job_id = jobstart(
+                        \ stack_command,
+                        \ { 'command': command
+                        \ , 'on_exit': function('s:CheckStackMakerAsync')
+                        \ })
+                    if job_id > 0
+                        call add(s:jobs, job_id)
+                    endif
+                else
+                    extend(stack_command, '> /dev/null 2>&1;', 'echo $?')
```
blueyed: s/extend/call extend/

Too bad that apparently only the linter caught this.. :/

blueyed:
Keep on rocking.
I am a bit jealous that Haskell makers are getting so much love, while the Python makers are waiting for more groundwork before they can also become great again.. ;)

bonds:
come to the Haskell side of the force, we have cool makers :)

bonds:
If you don't have anything else to add let me know and I'll squash this.

blueyed:
Yes, please squash.

PR #729, opened 2016-10-22
Add Verilog support with iverilog

robinkjoy:


blueyed:
Thanks.
Slightly amended in 4685d96.
Please add it to the list of makers in the wiki.

PR #746, opened 2016-10-29
add pandoc filetype

Alok:
For the `vim-pandoc` plugin and anyone setting `ft=pandoc`.


blueyed:
Thanks!
But it shows that there needs to be some alias system for this.

PR #747, opened 2016-10-30
Improve the shellcheck maker compatibility

dojoteef:
- The shellcheck maker now respects the '# shellcheck shell=<shell>'
  directive
- The shellcheck maker now correctly determines if zsh support is
  included
- Added utility functions for parsing & comparing semver version strings
  SEE: http://semver.org
- Included tests for all new functionality

NOTE: I noticed some weird issues that I did not address. For example, the filetype is being used to determine if shellcheck supports the specific shell, but the filetype should will always be 'sh' since it is an 'sh' maker. Additionally there is a zsh maker, and it should probably use the version determination code I put in to see if shellcheck supports zsh. Maybe the shellcheck code could better be refactored to use the neomake#utils#GetSupersetOf functionality.

### DIFF
--- a/autoload/neomake/makers/ft/sh.vim
+++ b/autoload/neomake/makers/ft/sh.vim
@@ -4,14 +4,35 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
     return ['sh', 'shellcheck']
 endfunction
 
-" IDEA: could be detected once from "shellcheck --help" (since older versions
-" support zsh, and newer might do so again).
-let s:shellcheck_supported = ['sh', 'bash', 'dash', 'ksh']
+function! s:shellcheck_determine_supported() abort
+    let s:shellcheck_supported = ['sh', 'bash', 'dash', 'ksh']
+
+    " zsh support was available from the first release of shellcheck and
+    " removed in version 0.3.6. Before shellcheck version 0.3.1 there was no
+    " way to get the version information.
+    let l:version_line = matchstr(split(system('shellcheck --version'), '\n'),
+                \ '^version:')
+    if v:shell_error != 0
+        let s:shellcheck_supported += ['zsh']
+    else
+        let l:version = matchstr(l:version_line, '^version:\s*\zs.*$')
+        if neomake#utils#CompareSemanticVersions(l:version, '0.3.6') == -1
+            let s:shellcheck_supported += ['zsh']
+        endif
+    endif
+endfunction
+call s:shellcheck_determine_supported()
+
+function! s:shellcheck_getshell() abort
+    let pattern = '^#\s*shellcheck\s*shell='
+    let line = matchstr(getline(1, line('$')), l:pattern)
+    return matchstr(l:line, l:pattern . '\zs[^ \t]*$')
+endfunction
 
 function! neomake#makers#ft#sh#shellcheck() abort
     let args = ['-fgcc']
     let shebang = matchstr(getline(1), '^#!\s*\zs.*$')
-    if !len(shebang)
+    if !len(shebang) && !len(s:shellcheck_getshell())
         if index(s:shellcheck_supported, &filetype) != -1
             let args += ['-s', &filetype]
         endif
--- a/autoload/neomake/makers/ft/sh.vim
+++ b/autoload/neomake/makers/ft/sh.vim
@@ -12,13 +12,11 @@ function! s:shellcheck_determine_supported() abort
     " way to get the version information.
     let l:version_line = matchstr(split(system('shellcheck --version'), '\n'),
                 \ '^version:')
-    if v:shell_error != 3 || empty(l:version_line)
-        echom 'zsh is supported!'
+    if v:shell_error == 3 || empty(l:version_line)
         let s:shellcheck_supported += ['zsh']
     else
         let l:version = matchstr(l:version_line, '^version:\s*\zs.*$')
         if neomake#utils#CompareSemanticVersions(l:version, '0.3.6') == -1
-            echom 'zsh is supported!'
             let s:shellcheck_supported += ['zsh']
         endif
     endif
--- a/autoload/neomake/makers/ft/sh.vim
+++ b/autoload/neomake/makers/ft/sh.vim
@@ -12,11 +12,13 @@ function! s:shellcheck_determine_supported() abort
     " way to get the version information.
     let l:version_line = matchstr(split(system('shellcheck --version'), '\n'),
                 \ '^version:')
-    if v:shell_error != 0
+    if v:shell_error != 3 || empty(l:version_line)
+        echom 'zsh is supported!'
         let s:shellcheck_supported += ['zsh']
     else
         let l:version = matchstr(l:version_line, '^version:\s*\zs.*$')
         if neomake#utils#CompareSemanticVersions(l:version, '0.3.6') == -1
+            echom 'zsh is supported!'
             let s:shellcheck_supported += ['zsh']
         endif
     endif
@@ -25,7 +27,7 @@ call s:shellcheck_determine_supported()
 
 function! s:shellcheck_getshell() abort
     let pattern = '^#\s*shellcheck\s*shell='
-    let line = matchstr(getline(1, line('$')), l:pattern)
+    let line = matchstr(getline(1, 2), l:pattern)
     return matchstr(l:line, l:pattern . '\zs[^ \t]*$')
 endfunction
 
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -322,3 +322,17 @@ function! neomake#utils#ExpandArgs(args) abort
     " Only expand those args that start with \ and a single %
     call map(a:args, "v:val =~# '\\(^\\\\\\|^%$\\|^%[^%]\\)' ? expand(v:val) : v:val")
 endfunction
+
+function! neomake#utils#hook(event, context) abort
+    if exists('#User#'.a:event)
+        let g:neomake_hook_context = a:context
+        call neomake#utils#DebugMessage('Calling User autocmd '.a:event
+                                      \ .' with context: '.string(a:context))
+        if v:version >= 704 || (v:version == 703 && has('patch442'))
+            exec 'doautocmd <nomodeline> User ' . a:event
+        else
+            exec 'doautocmd User ' . a:event
+        endif
+        unlet g:neomake_hook_context
+    endif
+endfunction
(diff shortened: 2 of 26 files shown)


dojoteef:
As far as I can tell, [this code](https://github.com/neomake/neomake/blob/4e56506c/autoload/neomake/makers/ft/sh.vim#L15) will never hit for any filetypes other than `sh`. I even tried it by setting my filetype to `bash` and indeed neomake does not realize that it has any makers for the `bash` filetype.

I think ideally the [zsh maker](https://github.com/neomake/neomake/blob/4e56506c/autoload/neomake/makers/ft/zsh.vim) could benefit from the [version checking code](https://github.com/neomake/neomake/pull/747/commits/614ec68e8e2c28803b61b4e36caf544738b82f2b#diff-e0a7a2befcb788aeac3ff156ac6c10b8R7) I added to ensure a version of shellcheck with zsh support is installed.

autoload/neomake/makers/ft/sh.vim
```diff
@@ -4,14 +4,35 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
...
+        if neomake#utils#CompareSemanticVersions(l:version, '0.3.6') == -1
+            let s:shellcheck_supported += ['zsh']
+        endif
+    endif
+endfunction
+call s:shellcheck_determine_supported()
+
+function! s:shellcheck_getshell() abort
+    let pattern = '^#\s*shellcheck\s*shell='
+    let line = matchstr(getline(1, line('$')), l:pattern)
```
blueyed: This is bad for large files.. should be limited probably?!

autoload/neomake/makers/ft/sh.vim
```diff
@@ -4,14 +4,35 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
...
+function! s:shellcheck_determine_supported() abort
+    let s:shellcheck_supported = ['sh', 'bash', 'dash', 'ksh']
+
+    " zsh support was available from the first release of shellcheck and
+    " removed in version 0.3.6. Before shellcheck version 0.3.1 there was no
+    " way to get the version information.
+    let l:version_line = matchstr(split(system('shellcheck --version'), '\n'),
+                \ '^version:')
+    if v:shell_error != 0
+        let s:shellcheck_supported += ['zsh']
```
blueyed: Can this check for the specific error code in that case, probably 3 (as with `--not-an-option`)?

autoload/neomake/makers/ft/sh.vim
```diff
@@ -4,14 +4,35 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
...
+    else
+        let l:version = matchstr(l:version_line, '^version:\s*\zs.*$')
+        if neomake#utils#CompareSemanticVersions(l:version, '0.3.6') == -1
+            let s:shellcheck_supported += ['zsh']
+        endif
+    endif
+endfunction
+call s:shellcheck_determine_supported()
```
blueyed: This should be done lazily, only when GetEnabledMakers is really called - and just once (if not cached already).

autoload/neomake/makers/ft/sh.vim
dojoteef: Unfortunately that will not work for all versions. I just verified the exit codes for versions 0.1.0 & 0.2.0. On version 0.1.0 there is no failure for the exit code, rather `--version` is considered a missing file with an exit code of 0.

```
Downloads/shellcheck-0.1.0$ ./shellcheck --version
No such file: --version
Downloads/shellcheck-0.1.0$ echo $?
0
```

While version 0.2.0 has an exit status of 1:
```
Downloads/shellcheck-0.2.0$ ./shellcheck --version
No such file: --version
Downloads/shellcheck-0.2.0$ echo $?
1
```

My code does not correctly support shellcheck version 0.1.0, so I'll update the code for that case.

dojoteef: Well I guess I can check for exit status 3 and check for `empty(l:version_line)`. That should support all current versions. I'll do that.

dojoteef:
I think the versioning part should be easy to check since it's covered by tests. All the possible permutations are covered, so as long as the tests pass the code should be correct. I've pretty much just copied the appropriate lines from the [semver.org](http://semver.org) website and put them as comments above the specific test cases.

Also, what do you think of my comments regarding the use of `filetype` to check for supported shells being incorrect?

autoload/neomake/utils.vim
```diff
@@ -336,3 +336,196 @@ function! neomake#utils#hook(event, context) abort
...
+    let l:parser = copy(a:)
+    let l:parser.parsed = {
+                \ 'major': 0,
+                \ 'minor': 0,
+                \ 'patch': 0,
+                \ 'stage': [],
+                \ 'metadata': []
+                \ }
+
+
```
mhinz: Superfluous line.

autoload/neomake/makers/ft/sh.vim
mhinz: Please always use the full command names, so `echomsg` in case of `echom`.

tests/ft_sh.vader
```diff
@@ -12,3 +12,6 @@ Execute (shellcheck):
   set ft=sh
   norm i#!/usr/bin/env bash
   AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc']
+
+  norm ggcG# shellcheck shell=bash
```
mhinz: `normal!`

autoload/neomake/makers/ft/sh.vim
dojoteef: Doh, my bad. Those were debugging statements for verifying my zsh check was correctly working. I'll remove those!

tests/ft_sh.vader
dojoteef: I was just following the current style in the file, which is my general approach to code I do not own. I'm happy to change the code, though it seems inconsistent with the rest of the file. It seems to be littered with them.

Would you want me to change the other instances?

mhinz: Hmm, yea. Either change all of them in an own commit or leave that out for now and we clean it up later.

dojoteef: I would prefer to clean up later so as not to have this pull request include unrelated changes.

blueyed:
@dojoteef 
Needs to be rebased.

@mhinz 
What do you think about this?

mhinz:
@blueyed I think it needs to be rebased. :>

Lots of things changed, and yea, I would try to avoid using `executable()` for sole detection purposes. For shebangs that contain `sh`/`dash`/`bash`/`ksh`, we just use `shellcheck -fgcc`, since shellcheck reads the shebang on its own.

Adding `zsh` to that list, would make it work with shellcheck versions supporting zsh. For current releases that's not the case, though. Shellcheck would default to using the bash dialect and issue a warning about an unsupported shebang.

dojoteef:
@mhinz Sorry, I am not sure I understand what changes you are asking for. Are you saying not to use `shellcheck --version` and the version parsing at all? If so, I am not sure how to make it do that and properly detect/add zsh support. As you've stated that could result in producing an error in the quickfix/location list. Does that mean you believe showing an error to be the desired behavior? Note that I believe in some versions of shellcheck it does not error when provided an invalid flag.

To be honest I do not use zsh and simply took the time to download and test for shellcheck compatibility since I need to make a change to the shellcheck maker. If you want me to remove any zsh related stuff, I'd be happy to.

As for rebasing, will do. Though it seems like something to do after the details of this change are settled before checking in, unless there are large changes that would affect this code.

mhinz:
Not sure how to handle zsh. Should we support features from old versions? I mean, as far as I understand it, zsh was only supported for some time and was removed again in recent versions.

blueyed:
I'd also say that we should in general only support the most recent versions of checkers - just for simplicity.
But OTOH the code to query versions is provided here now already and can become useful in other places, too.

autoload/neomake/makers/ft/sh.vim
```diff
@@ -4,14 +4,35 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
...
+    let pattern = '^#\s*shellcheck\s*shell='
+    let line = matchstr(getline(1, 2), l:pattern)
+    return matchstr(l:line, l:pattern . '\zs[^ \t]*$')
+endfunction
 
 function! neomake#makers#ft#sh#shellcheck() abort
     let args = ['-fgcc']
     let shebang = matchstr(getline(1), '^#!\s*\zs.*$')
-    if !len(shebang)
+    if !len(shebang) && !len(s:shellcheck_getshell())
```
mhinz: I think it's better to use `empty()` instead of `!len()`.

mhinz:
Yea, I'm not opposed to the changes. What's done is done. And calling `system('shellcheck --version')` only once on `ft=sh` is fine with me for detection purposes.

blueyed:
A problem with the current `sh` maker is btw that it will enable `shellcheck` also for a `#!/usr/bin/env zsh` shebang (`ft=sh`), resulting in an error from shellcheck that it is not supported.
That won't change with this PR, will it?
For this the check should be done in `neomake#makers#ft#sh#EnabledMakers`, or the shellcheck maker should have a way to skip its run conditionally?!

mhinz:
`neomake#makers#ft#sh#shellcheck()` was changed quite a bit lately to make it always do the Right Thing.

For the sake of getting on, I suggest removing the actual zsh support from this PR and get it merged it quickly. Afterwards we add a commit that does the dection and sets `s:shellcheck_supports_zsh` in `autoload/neomake/makers/ft/sh.vim` which again get used in `neomake#makers#ft#sh#shellcheck()`. Then we have all the buffer processing in the same place.

dojoteef:
@blueyed @mhinz Okay, this change is severely out of date. I propose I essentially scrap this and simply put the one change I really care about which is to respect a shellcheck directive for the shell if the shebang is missing.

Basically this line:
```
if match(getline(1), '\v^#!.*<%(sh|dash|bash|ksh)') >= 0
```

Will become:
```
if match(getline(1), '\v^#!.*<%(sh|dash|bash|ksh)') >= 0
        \ || match(getline(1), '\v^#\s+shellcheck\s+shell=') >= 0
```

Does that sound reasonable? If so, then I will code it up and add a test.

blueyed:
Sounds reasonable (to me at least).

mhinz:
I don't know that directive, but I'm sure it exists. ;] And please create a new PR for it. Closing.

(And it should be `\=`!)

PR #757, opened 2016-11-10
Add proselint for mail filetype

KyleOndy:
Very similar to the [text](https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/text.vim) filetype maker, but for the mail filetype

### DIFF
--- a/autoload/neomake/makers/ft/mail.vim
+++ b/autoload/neomake/makers/ft/mail.vim
@@ -1,0 +1,9 @@
+function! neomake#makers#ft#mail#EnabledMakers()
+    return ['proselint']
+endfunction
+
+function! neomake#makers#ft#mail#proselint()
+    return {
+                \ 'errorformat': '%f:%l:%c: %m'
+                \ }
+endfunction
--- a/autoload/neomake/makers/ft/mail.vim
+++ b/autoload/neomake/makers/ft/mail.vim
@@ -4,6 +4,6 @@ endfunction
 
 function! neomake#makers#ft#mail#proselint() abort
     return {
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/mail.vim
```diff
@@ -0,0 +1,9 @@
+function! neomake#makers#ft#mail#EnabledMakers()
+    return ['proselint']
+endfunction
+
+function! neomake#makers#ft#mail#proselint()
+    return {
+                \ 'errorformat': '%f:%l:%c: %m'
```
blueyed: Should specify a type explicitly, e.g. by prefixing it with `%E` for errors.

Also please add " abort" to the function definitions (check ":Neomake vint" ;)).

KyleOndy:
A lot of the current makers don't have explicit types or abort on their functions.

Should they? If so I can put in some PRs.

blueyed:
> Should they? If so I can put in some PRs.

Yes, the linter complains.
It gets currently done only with changed codes.
If you create a PR, that would cause conflicts for existing / pending ones probably.

blueyed:
> I believe a warning makes more sense then a hard error. Prose is subjective

:+1: 
See also https://github.com/neomake/neomake/issues/584.

blueyed:
Thanks!

blueyed:
Please add it to the wiki: https://github.com/neomake/neomake/wiki/Makers

KyleOndy:
Added to wiki. Thanks for the fast and easy PR.

PR #762, opened 2016-11-13
[RFC] Make `:Neomake! <maker>` great again

mhinz:
Without this, `:Neomake! <tab>` would correctly list potential makers, but using
any of them would result in: "Maker not found".

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -947,17 +947,11 @@ endfunction
 
 function! neomake#Make(file_mode, enabled_makers, ...) abort
     let options = a:0 ? { 'exit_callback': a:1 } : {}
-    if a:file_mode
-        let options.enabled_makers = len(a:enabled_makers) ?
-                    \ a:enabled_makers :
-                    \ neomake#GetEnabledMakers(&filetype)
-        let options.ft = &filetype
-        let options.file_mode = 1
-    else
-        let options.enabled_makers = len(a:enabled_makers) ?
-                    \ a:enabled_makers :
-                    \ neomake#GetEnabledMakers()
-    endif
+    let options.file_mode = a:file_mode
+    let options.ft = &filetype
+    let options.enabled_makers = len(a:enabled_makers)
+                    \ ? a:enabled_makers
+                    \ : neomake#GetEnabledMakers(a:file_mode ? &filetype : '')
     return s:Make(options)
 endfunction
 


autoload/neomake.vim
```diff
@@ -957,6 +957,7 @@ function! neomake#Make(file_mode, enabled_makers, ...) abort
         let options.enabled_makers = len(a:enabled_makers) ?
                     \ a:enabled_makers :
                     \ neomake#GetEnabledMakers()
+        let options.ft = &filetype
```
blueyed: Could be moved out of the if/else then.

mhinz: Okay, I refactored the whole function instead.

autoload/neomake.vim
```diff
@@ -947,17 +947,11 @@ endfunction
...
-    else
-        let options.enabled_makers = len(a:enabled_makers) ?
-                    \ a:enabled_makers :
-                    \ neomake#GetEnabledMakers()
-    endif
+    let options.file_mode = a:file_mode
+    let options.ft = &filetype
+    let options.enabled_makers = len(a:enabled_makers)
+                    \ ? a:enabled_makers
+                    \ : neomake#GetEnabledMakers(a:file_mode ? &filetype : '')
```
mhinz: I know there's a subtle difference now, since `neomake#GetEnabledMakers()` was changed to `neomake#GetEnabledMakers('')`, but the outcome is the same.

blueyed:
Awesome, thanks!

blueyed:
Well, I think this is wrong.

`Neomake!` is meant to run project makers (the ones not in `ft/`).
The completion in this case was just wrong.

With this change, `Neomake! cabal` results in "Maker not found".

The code responsible for this is: https://github.com/neomake/neomake/blob/850ebc88b5b353a763a04c1fc0b1556b2a148888/autoload/neomake.vim#L269-L282.

blueyed:
I could imagine removing the implicit "if there are fts passed into GetMaker this is filetype" - but am not sure about this.
I think it should be possible to run makers also on filetypes that are not configured after all.

See https://github.com/neomake/neomake/pull/781 for proper completion of `Neomake!` / `NeomakeProject`.

blueyed:
@mhinz 
Check out https://github.com/neomake/neomake/pull/790 :)

Issue #761, opened 2016-11-13
Error with pandoc files: markdownlint not found


zyphrus:
When using `:Neomake` in pandoc files an error occurs,  markdownlint not found (more below).

However there are no issues when running in a markdown file.

### Expected behaviour

To run `:Neomake` in a pandoc file without getting errors.

#### Neomake debug information

Async support: 1
Current filetype: pandoc

##### Enabled makers

```
Neomake: Maker not found: markdownlint
                                      Error detected while processing function neomake#DisplayInfo[7]..neomake#GetEnabledMakers[33]..neomake#utils#AvailableMakers[1]..neomake#utils#MakerIsAvailable:
line    8:
E716: Key not present in Dictionary: exe)
E116: Invalid arguments for function executable(maker.exe)
E15: Invalid expression: executable(maker.exe)
For the current filetype (with :Neomake): -1
```

### Additional Info

Tried on Linux and OSX with neovim 0.1.6 and vim 7.4 and markdownlint is not installed on any of them

Use with [vim-pandoc/vim-pandoc-syntax](https://github.com/vim-pandoc/vim-pandoc-syntax) and [full config here](https://github.com/zyphrus/nvim)

blueyed:
Thanks for your report, should be fixed in #763 for now.

I wonder though, if `mdl` should be used at all - and if so, why `markdownlint` is not used (or is this an oversight (missing autoload function).
/cc @Alok (who submitted https://github.com/neomake/neomake/pull/746).

zyphrus:
Thanks for getting this resolved so quickly. Just testing it again as a follow up from the latest master and getting a message for markdownlint not found.Not sure if this is suppose to happen but it is not as bad as it was before.

Normal markdown file type is still fine, this is just in pandoc file type.

### Output of the ":verb NeomakeInfo" command

#### Neomake debug information
Async support: 1
Current filetype: pandoc


##### Enabled makers
Neomake: Maker not found: markdownlint
For the current filetype (with :Neomake): []
NOTE: you can define g:neomake_pandoc_enabled_makers to configure it (or b:neomake_pandoc_enabled_makers).
For the project (with :Neomake!): []
NOTE: you can define g:neomake_enabled_makers to configure it.


##### Settings

```
shell: /bin/zsh
shellcmdflag: -c
Windows: 0
```


#### :version

```
NVIM 0.1.6
Build type: RelWithDebInfo
[...]

PR #763, opened 2016-11-13
Fix neomake#utils#MakerIsAvailable with empty result from GetMaker

blueyed:
Fixes https://github.com/neomake/neomake/issues/761.

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -23,7 +23,7 @@ function! neomake#utils#LogMessage(level, msg, ...) abort
     let verbose = get(g:, 'neomake_verbose', 1)
     let logfile = get(g:, 'neomake_logfile')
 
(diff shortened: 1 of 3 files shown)


autoload/neomake/utils.vim
```diff
@@ -181,6 +181,9 @@ function! neomake#utils#MakerIsAvailable(ft, maker_name) abort
     endif
     if !has_key(s:available_makers, a:maker_name)
         let maker = neomake#GetMaker(a:maker_name, a:ft)
+        if maker == {}
```
mhinz: `maker == {}` or `empty(maker)`? :-)

I have no real favourite, but I guess we should decide for one and be consistent with all new changes.

blueyed: I don't now either.. `empty` seems better marginally?!

blueyed: Amended.  There was one other occurrence.

PR #768, opened 2016-11-14
Fix post processing of jobs when changing windows

blueyed:
This removes s:{loc,qf}list_nr, where s:loclist_nr was using the
(unstable) winnr of the maker.


blueyed:
Fails for Vim-async.. :/

PR #770, opened 2016-11-14
Allow stylelint for .less

partounian:
As stated in the other PR, there must be a better way rather than checking if stylelint is executable. This was modeled to be as similar as possible to the PR #665 . Also, thank you to @rstacruz for mentioning stylelint.


blueyed:
Thanks!

PR #771, opened 2016-11-14
return markdown#markdownlint settings for pandoc#markdownlint

davidosomething:
also add aborts


blueyed:
Thanks!
I see that markdown files get set to ft=pandoc by default via https://github.com/vim-pandoc/vim-pandoc/blob/9fe82eb56ff0ebbc22705e8fe911df7d3503f4ff/plugin/pandoc.vim#L101 now.

PR #772, opened 2016-11-14
Add write-good support

Alok:
I accidentally had `neomake#makers#ft#text#EnabledMakers` in  `autoload/neomake/makers/ft/text.vim` return `neomake#makers#ft#markdown#EnabledMakers()` instead of just `['proselint']`, but I think you can edit that line. Sorry about that.


blueyed:
@Alok 
You should really create PRs from separate branches..!

Alok:
I realized that, let me fix it and add (yet another) PR. Sorry for the trouble.

PR #777, opened 2016-11-16
Refactor neomake#makers#ft#sh#shellcheck()

### LINKED ISSUE #776: shellcheck on *.bash
Not all bash scripts will have a shebang (e.g. sourced files ending in `.bash`, which might be sourced from a `.bashrc` and therefore should not have +x or a shebang to prevent spawning)

They should be assumed to be bash if the file extension is `.bash` and there is no shebang

I'm too lazy to PR right now, but here's my local fix:
https://github.com/davidosomething/dotfiles/blob/5732298ca163e6c4d0afc784c606eadc254bb9c4/vim/plugin/plug-neomake.vim#L321-L332

Perhaps a `g:neomake_shellcheck_perfer_posix` can be introduced if someone really wants to assume `.bash` should be `.sh` and POSIX compliant

mhinz:
Simplify the function and also consider filename extensions and global variabes
from `:h ft-sh-syntax`.

Closes #776.

### DIFF
--- a/autoload/neomake/makers/ft/sh.vim
+++ b/autoload/neomake/makers/ft/sh.vim
@@ -4,20 +4,23 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
     return ['sh', 'shellcheck']
 endfunction
 
-" IDEA: could be detected once from "shellcheck --help" (since older versions
-" support zsh, and newer might do so again).
-let s:shellcheck_supported = ['sh', 'bash', 'dash', 'ksh']
-
 function! neomake#makers#ft#sh#shellcheck() abort
-    let args = ['-fgcc']
-    let shebang = matchstr(getline(1), '^#!\s*\zs.*$')
-    if !len(shebang)
-        if index(s:shellcheck_supported, &filetype) != -1
-            let args += ['-s', &filetype]
-        endif
+    let ext = expand('%:e')
+    let shebang = getline(1)
+
+    if ext ==# 'dash' || shebang =~# 'dash'
+        let ft = 'dash'
+    elseif exists('g:is_sh') || ext ==# 'sh' || shebang =~# '\<sh'
+        let ft = 'sh'
+    elseif exists('g:is_kornshell') || exists('g:is_posix') || ext ==# 'ksh'
+        \ || shebang =~# 'ksh'
+        let ft = 'ksh'
+    else
+        let ft = 'bash'
     endif
+
     return {
-        \ 'args': args,
+        \ 'args': ['-fgcc', '-s', ft],
         \ 'errorformat':
             \ '%f:%l:%c: %trror: %m,' .
             \ '%f:%l:%c: %tarning: %m,' .
--- a/tests/ft_sh.vader
+++ b/tests/ft_sh.vader
@@ -1,14 +1,48 @@
 Include: include/setup.vader
 
 Execute (shellcheck):
-  AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc']
-  set ft=sh
-  AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc', '-s', 'sh']
-  set ft=bash
-  AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc', '-s', 'bash']
-  set ft=nadda
-  AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc']
+  " By variable...
+  let g:is_sh = 1
+  AssertEqual ['-fgcc', '-s', 'sh'], neomake#makers#ft#sh#shellcheck().args
+  unlet g:is_sh
+  let g:is_kornshell = 1
+  AssertEqual ['-fgcc', '-s', 'ksh'], neomake#makers#ft#sh#shellcheck().args
+  unlet g:is_kornshell
+  let g:is_posix = 1
+  AssertEqual ['-fgcc', '-s', 'ksh'], neomake#makers#ft#sh#shellcheck().args
+  unlet g:is_posix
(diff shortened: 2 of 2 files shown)


mhinz:
Can someone explain the build issues to me? :>

mhinz:
Hm, I'm not sure about that. As far as I see, `b:is_bash` and friends are only used internally by `$VIMRUNTIME/syntax/sh.vim`.

E.g. when you put this in your vimrc:

```vim
autocmd FileType sh let b:is_bash = 1
```

..and then open a file `foo.sh`, both, `b:is_{ba,}sh`, will be set. It's not documented either.

@davidosomething Do you use the `b:` in practice?

davidosomething:
ah you're right, the `b:` vars can be ignored as they're computed once and used only on first run of the syn file

blueyed:
@mhinz 
There's an assertion failure in the tests: https://travis-ci.org/neomake/neomake/jobs/176248873#L779

You can run these tests using `make tests/ft_sh.vader`.

The failure comes from the first assertion:

>   AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc']

mhinz:
Okay, I rewrote all shellcheck tests. :>

p.s. Too bad that Vader doesn't list failing assertions at the end. So what I've ended up with was grepping the output for `(X)`. Is there a better approach?

blueyed:
> p.s. Too bad that Vader doesn't list failing assertions at the end. So what I've ended up with was grepping the output for (X). Is there a better approach?

Yes, it's bad in that regard - but our automatic `Log` calls make it especially worse.
I tend to search the scrollback in my terminal for `(X)`.

I also think it would be nice to also have some color, e.g. the `(X)` lines colored in red - if that is possible through the redirection/output.
A summary of failures would be nice, too.  Also, the line number for `AssertX` failures.

tests/ft_sh.vader
```diff
@@ -1,14 +1,58 @@
...
-  set ft=sh
-  AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc', '-s', 'sh']
-  set ft=bash
-  AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc', '-s', 'bash']
-  set ft=nadda
-  AssertEqual neomake#makers#ft#sh#shellcheck()['args'], ['-fgcc']
+  " By variable...
+  let g:is_sh = 1
+  AssertEqual ['-fgcc', '-s', 'sh'], neomake#makers#ft#sh#shellcheck().args
+  unlet g:is_sh
```
blueyed: You could use Vader's `Save`/`Restore` here, but it's only relevant to not mess state when running Vader inside of Vim I'd say.

autoload/neomake/makers/ft/sh.vim
```diff
@@ -4,20 +4,23 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
...
+    if ext ==# 'dash' || shebang =~# 'dash'
+        let ft = 'dash'
+    elseif exists('g:is_bash') || ext ==# 'bash' || shebang =~# 'bash'
+        let ft = 'bash'
+    elseif exists('g:is_kornshell') || exists('g:is_posix') || ext ==# 'ksh'
+        \ || shebang =~# 'ksh'
+        let ft = 'ksh'
+    else
+        let ft = 'sh'
     endif
```
blueyed: There should maybe be another fallback here to not use `-s` at all (if there is any other shebang), and leave the detection to `shellcheck` in that case?!  That was the thing covered by the failing test IIRC.
But just a remark, nothing required really probably.

mhinz: Well, I could change it to `bash` instead. That's what shellcheck does if `-s` is not specified.

So, what about this logic?

```vim
    if ext ==# 'dash' || shebang =~# 'dash'
        let ft = 'dash'
    elseif exists('g:is_sh') || shebang =~# '\<sh'
        let ft = 'sh'
    elseif exists('g:is_kornshell') || exists('g:is_posix') || ext ==# 'ksh'
        \ || shebang =~# 'ksh'
        let ft = 'ksh'
    else
        let ft = 'bash'
    endif
```

Note that I left out `ext ==# 'sh'` on purpose, because lots of people use `.sh` extension but use if for non-bourne shell scripting.

tests/ft_sh.vader
mhinz: What would be the net gain here? As far as I see `Save`/`Restore` are used for variables only. (Vader's documentation could be better. :-P)

mhinz:
The build for Vim 8.0.69 has 2 failing tests that I can reproduce locally. Has nothing to do with this PR, though.

blueyed:
Fixed by https://github.com/neomake/neomake/commit/4955a749c77fd2190a420ea71a0315ddb345b125?
Please try rebasing.

mhinz:
@blueyed Yup, that fixed it!

mhinz:
@blueyed It's _not_ fixed yet. It's just not reliably reproducible.

autoload/neomake/makers/ft/sh.vim
```diff
@@ -4,20 +4,23 @@ function! neomake#makers#ft#sh#EnabledMakers() abort
...
-    if !len(shebang)
-        if index(s:shellcheck_supported, &filetype) != -1
-            let args += ['-s', &filetype]
-        endif
+    let ext = expand('%:e')
+    let shebang = getline(1)
+
+    if ext ==# 'dash' || shebang =~# 'dash'
+        let ft = 'dash'
+    elseif exists('g:is_sh') || ext ==# 'sh' || shebang =~# '\<sh'
```
blueyed: Should match `'\<sh$'` for the shebang here (end-of-line).

Issue #782, opened 2016-11-18
Cargo maker not found in rust files


pengowen123:
I open a rust file inside of a project that is built with cargo, and run :Neomake! cargo. It outputs:
Neomake: Maker not found: cargo

### Expected behavior

I expect the above command to run the cargo maker on my project.

### Steps to reproduce

Install neovim and vimplug. Use vimplug to install neomake. Open a rust file, and run :Neomake! cargo.

#### Neomake debug information
Async support: 1
Current filetype: rust

##### Enabled makers
For the current filetype (with :Neomake): ['rustc']
NOTE: you can define g:neomake_rust_enabled_makers to configure it (or b:neomake_rust_enabled_makers).
For the project (with :Neomake!): []
NOTE: you can define g:neomake_enabled_makers to configure it.

##### Settings
```


shell: /bin/bash
shellcmdflag: -c
Windows: 0
```

#### :version
```

NVIM 0.1.6
Build type: RelWithDebInfo
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -Wconversion -O2 -g -DDISABLE_LOG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -D_GNU_SOURCE -I/build/neovim/src/build/config -I/build/neovim/src/neovim-0.1.6/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/build/neovim/src/build/src/nvim/auto -I/build/neovim/src/build/include
Compiled by builduser

Optional features included (+) or not (-): +acl   +iconv    +jemalloc +tui      
For differences from Vim, see :help vim-differences

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

```

#### :messages
```
"main.rs" 6L, 61C

NVIM 0.1.6
Build type: RelWithDebInfo
[...]

PR #784, opened 2016-11-18
neomake#Make: only set ft for file_mode

blueyed:
This partly reverts 02e9912a (https://github.com/neomake/neomake/pull/762).
Fixes https://github.com/neomake/neomake/issues/782.


mhinz:
Okay, I try to follow the development..

With this reverted, and #781 getting merged, `Neomake!` would have show the proper project-wide completions (only 4 at the moment), but the "Maker not found" error would still persist, right?

PR #788, opened 2016-11-19
Puppet: Add errorformat with column

lbischof:
The output isn't parsed correctly, because the errorformat is missing the column. 

https://github.com/vim-syntastic/syntastic/blob/master/syntax_checkers/puppet/puppet.vim


blueyed:
Thanks!

Issue #597, opened 2016-09-08
Improve CONTRIBUTING.md


blueyed:
Should include best practices for e.g submitting new makers.

I can think of e.g. linking to good examples and a relevant section in the documentation. 

Ref: https://github.com/neomake/neomake/pull/595#issuecomment-245736252

davidosomething:
things to mention:
- https://github.com/neomake/neomake/wiki/Running-Neomake's-test-suite and other wiki pages
- vimhelplint
- vim-vint, vimlint

define some coding standards (indentation), maybe introduce and .editorconfig rc file

Kd-Here:
Hello geeks can you give me chance to contribute for documentation.
It will be my first documentation open soucre contribution.
Could you please help me with it

PR #789, opened 2016-11-19
Indentation

poohzrn:
Indented all `.vim` files and inserted `vim: ts=4 sw=4 et `where missing

### DIFF
--- a/autoload/neomake/makers/ft/cuda.vim
+++ b/autoload/neomake/makers/ft/cuda.vim
@@ -6,10 +6,10 @@ endfunction
 
 function! neomake#makers#ft#cuda#nvcc()
     return {
-        \ 'exe': 'nvcc',
-        \ 'errorformat':
-            \ '%f\(%l\): %trror: %m,'.
-            \ '%f\(%l\): %tarning: %m,'.
-            \ '%f\(%l\): %m',
-        \ }
+                \ 'exe': 'nvcc',
+                \ 'errorformat':
+                \ '%f\(%l\): %trror: %m,'.
+                \ '%f\(%l\): %tarning: %m,'.
+                \ '%f\(%l\): %m',
+                \ }
 endfunction
--- a/autoload/neomake/makers/ft/d.vim
+++ b/autoload/neomake/makers/ft/d.vim
@@ -45,15 +45,15 @@ function! s:DmdStyleMaker(args)
     call s:UpdateDub()
     let l:args = ['-w', '-wi', '-c', '-o-', '-vcolumns'] + a:args + s:dubImports
     return {
-        \ 'args': l:args,
-        \ 'errorformat':
-        \     '%f(%l\,%c): %trror: %m,' .
-        \     '%f(%l): %trror: %m,' .
-        \     '%f(%l\,%c): %tarning: %m,' .
-        \     '%f(%l): %tarning: %m,' .
-        \     '%f(%l\,%c): Deprecation: %m,' .
-        \     '%f(%l): Deprecation: %m,',
-        \ }
+                \ 'args': l:args,
+                \ 'errorformat':
+                \     '%f(%l\,%c): %trror: %m,' .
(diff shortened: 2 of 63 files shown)


poohzrn:
changes made with:
`nvim --headless $(find . -name '*.vim') -c "argdo :normal :gg=G" -c "x"`

blueyed:
Thanks.
I've planned to do this also, but it causes unnecessary conflicts with all my local stashes/branches (and pending PRs).
Therefore I think it's not really worth it currently.

Apart from that:

1. I'd prefer to use `sw=2`
2. I would remove all modelines.  IMHO they only add noise, and this should be configured for the project, e.g. by using an `.editorconfig` file.

autoload/neomake/makers/ft/d.vim
```diff
@@ -45,15 +45,15 @@ function! s:DmdStyleMaker(args)
...
-        \ }
+                \ 'args': l:args,
+                \ 'errorformat':
+                \     '%f(%l\,%c): %trror: %m,' .
+                \     '%f(%l): %trror: %m,' .
+                \     '%f(%l\,%c): %tarning: %m,' .
+                \     '%f(%l): %tarning: %m,' .
+                \     '%f(%l\,%c): Deprecation: %m,' .
+                \     '%f(%l): Deprecation: %m,',
+                \ }
```
davidosomething: I prefer this style, where there is inline spacing to show that the following string is the value of the above key

autoload/neomake/makers/ft/cuda.vim
```diff
@@ -6,10 +6,10 @@ endfunction
...
-        \ 'errorformat':
-            \ '%f\(%l\): %trror: %m,'.
-            \ '%f\(%l\): %tarning: %m,'.
-            \ '%f\(%l\): %m',
-        \ }
+                \ 'exe': 'nvcc',
+                \ 'errorformat':
+                \ '%f\(%l\): %trror: %m,'.
+                \ '%f\(%l\): %tarning: %m,'.
+                \ '%f\(%l\): %m',
```
davidosomething: imo this is ugly, needs inline spacing to distinguish key and value

poohzrn:
1. I don't have any preferences regarding `sw=2` vs. `sw=4`
2. `.editorconfig` is a great idea, however, I don't know any ways to to that other than using a [plugin](http://editorconfig.org/)
    To enforce this some a step to the ci stuff should be added - or do you have any suggestions?

autoload/neomake/utils.vim
```diff
@@ -332,7 +332,7 @@ function! neomake#utils#hook(event, context) abort
     if exists('#User#'.a:event)
         let g:neomake_hook_context = a:context
         call neomake#utils#DebugMessage('Calling User autocmd '.a:event
-                                      \ .' with context: '.string(a:context))
+                    \ .' with context: '.string(a:context))
```
davidosomething: this should have been caught by vint -- needs some spacing around the `.`

davidosomething:
there are several plugins, dahu and sgur have vimL only implementations of editorconfig as well

for the sw, I prefer 2 also since most VimL as well as the helpfiles are formatted for tw of 80

poohzrn:
I think we should just close the pull request and ping #597

autoload/neomake/utils.vim
blueyed: @davidosomething 
It's only style, not an error?!  AFAIK vint does not handle this, does it?

blueyed:
Ok.

Let's go with `sw=2` then.

What about the modelines?
https://github.com/embear/vim-localvimrc is also useful for project-wide settings.  Maybe we could add such a file, additionally to an editorconfig?
I don't use an editorconfig plugin myself - the last time I've looked it seemed like too much overhead.
@davidosomething 
Do you use one?  Can you recommend one?

> To enforce this some a step to the ci stuff should be added

Yes, would be nice.  What could we use here?

Issue #589, opened 2016-09-07
Pass parameter to makeprg


sakhnik:
How could I run make, make clean, make check with `:Neomake!`? I'd like to do `set makeprg=make\ -C\ BUILD` once, and then to do my development cycle continuously.

Tim Pope's Dispatch allowed me to call `:Make clean` etc, for example.

blueyed:
Yes, the interface for that is not good.

Setting `:set makeprg=make\ -C\ BUILD` should be used by `:Neomake!` though?!
Make sure to check with `:copen`.

I think the following would be nice to have:

    :Neomake maker --foo

If the "maker" is omitted, the first one would be used, and with `:Neomake!` that would be `&makeprg`/'make' then.
So `:Neomake! clean` would work.

But it conflicts with passing multiple makers to `:Neomake`, e.g. `:Neomake python pylama`.
OTOH that use case is not very common I would say, and could be moved to e.g. `:NeomakeMultiple` or something similar?!

Related: https://github.com/neomake/neomake/issues/175

I consider/use Neomake only for the linting/checker stuff, and still use vim-dispatch otherwise quite often.

davidosomething:
from https://github.com/neomake/neomake/pull/651#pullrequestreview-1889695

> btw: with NeomakeSh I was thinking to force the errorformat to accept everything %+G, so you will get all the output.
> 
> As for #589 defining your own maker for those one-off cases is cumbersome.. that's why I think a new command and or support for e.g. Neomake! -- clean (calls &makeprg foo) or Neomake linter --ignore=foo (calls the linter maker with --ignore=foo added to args) should be added.

+1 for `:Neomake! [optional_maker_name] -- [pass-through rest of string as args for maker.exe]`

wookayin:
+1, for `:Neomake <target>` invokes something similar to `$ make <target>` in the shell.
Especially, `:Neomake clean` does not work out-of-box.

blueyed:
@wookayin 
That requires a new command I guess, since currently this expects a maker name (and even multiple) there.

blueyed:
FWIW I would agree with some API breaking change in this regard.

Most people are using it through an autocommand, and there it could be `NeomakeFileLint` after all.
Relevant: https://github.com/neomake/neomake/projects/1.
If we would have decent default config, we could control all those things much better.

PR #790, opened 2016-11-19
[RFC] Use file-mode/filetype makers for project mode

blueyed:
This allows for `:Neomake! vint` to use the (file-mode) "vint" maker on
all Vim files in the project (via `glob('**/*.vim', 0, 1)`).

Currently this will only look at makers for the current filetype, so `:Neomake! vint` does not work when the filetype is not "vim".  This could be changed to work always?!

What about something like `:Neomake! vint **/*.vim` to specify the files being handled?
This should be made possible through #589.

TODO:
 - [ ] doc
 - [x] fix/adjust command completion
 - [ ] the mapping from &ft to `.ext` (to get the files to act on) does not work in a lot of cases, e.g. `ft=python` => `.py`, `ft=javascript` => `.js`.

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -220,7 +220,8 @@ endfunction
 
 " Get a setting by key, based on filetypes, from the buffer or global
 " namespace, defaulting to default.
-function! neomake#utils#GetSetting(key, maker, default, fts, bufnr) abort
+function! neomake#utils#GetSetting(key, maker, default, fts, bufnr, ...) abort
+  let expected_type = a:0 ? a:1 : -1
   let maker_name = has_key(a:maker, 'name') ? '_'.a:maker.name : ''
   if len(a:fts)
     for ft in a:fts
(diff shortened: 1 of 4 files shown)


autoload/neomake/utils.vim
```diff
@@ -342,3 +357,20 @@ function! neomake#utils#hook(event, context) abort
...
+    endif
+    " Incomplete!  Right approach?!
+    let ext_by_ft = {
+                \ 'cpp': ['c', 'h'],
+                \ 'python': ['py'],
+                \ 'javascript': ['js'],
+                \ }
+    let exts = has_key(ext_by_ft, a:ft) ? ext_by_ft[a:ft] : [a:ft]
+    return map(copy(exts), "'**/*.'.v:val")
+endfunction
```
blueyed: This won't scale really, will it?
At least it allows for settings to provide this already.
Should it be a callback in the filetype maker file, similar to GetEnabledMakers?

blueyed: Maybe we should just require files as arguments in this case? `:Neomake! vint **/*.vim`
It could still be set as a setting, but would not have defaults.

blueyed:
This worked quite good today on a project using `Neomake! luacheck`, but then it also included too many files (the build directory etc) - after all using the files filtered by `git ls-files` would have helped here (but that is getting more specialized right there).

blueyed:
I'd appreciate some feedback on this.. /cc @davidosomething @mhinz 

I think it would be good to move more makers to project makers (which means they can act on the current directory), and then get something similar to this PR in, which allows to run file makers on a subset of files.

With regard to passing in args to makers (https://github.com/neomake/neomake/issues/589), I could imagine also to only allow a single maker name (for some new command possibly?), so we could have `Neomake! luacheck src/**/*.lua` - without the extra dashes (although that's not too bad after all).

After all, we should also consider to refactor makers into "objects", so that they could derive from others more easily, and provide an attribute to indicate if they can run on a directory, too.

davidosomething:
I think #589 is still useful, but this ticket replicates existing behavior that linters and build tools already provide, and in a much more robust way.

For the vint case, user should run vint via a makefile/shell script that does the globbing, and configure neomake to run `make`.
Globbing and file subsetting -- `**/*` -- should be handled by the shell or build tool that they are running (make, or node, or python glob2, etc). For instance, the luacheck case you used could be handled by configuring `.luacheckrc` `include` and `exclude` files, and the maker would just be luacheck

mhinz:
Hmm.

Usually we can't know what the user wants to test. Unconditional recursive checking of _every_ file of a filetype  in a directory is kind of offensive and will most likely include false positives.

When using `make` via Neomake to run a linter on multiple files, does Neomake set the `errorformat` to a proper value? That would cause mayhem and basically defeat the purpose of the plugin which is being clever about using the right errorformat for each tool.

Maybe instead of checking a whole directory, `Neomake! vint` could vint all Vim files currently loaded in Vim?

blueyed:
Closing for now.

PR #235, opened 2015-11-23
go: add support for gometalinter

buztard:
This adds support for https://github.com/alecthomas/gometalinter
if available

PR #795, opened 2016-11-22
Go: add gometalinter

bradleyfalzon:
Go has a quite a number of static analysis tools outside of the standard compiler, lint and vet checks.

Gometalinter help manage (install and update) and standardise the output of the checks to a single format.

I've added gometalinter support because a user can then use this intermediate tool to enable or disable various checkers/linters themselves without having to further customise neomake.

WIP:

- ~I'd like users to be able to adjust the `args` as they see fit, the ones I've proposed show few false positives and usually refer to issues that really do need to be fixed (or have left dead code around and can be safely removed).~ Fixed in ad531ac7473a732b8a7c39b3e02e2e8b88b026ab
- ~Can users disable the gometalinter if they do not want to use it, even though it is installed?~ Yes they can by doing something like `let g:neomake_go_enabled_makers = ['go', 'golint']`
- ~What happens if gometalinter isn't installed, does neomake handle this gracefully?~ Yes it does
- ~Can someone show me an example commit with a basic test that I could use to help ensure these are tested?~

Currently the arguments are just my opinion as to what should 

Relates to #235 which also tried to implement the same functionality.

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#go#EnabledMakers() abort
-    return ['go', 'golint', 'govet']
+    return ['go', 'golint', 'govet', 'gometalinter']
 endfunction
 
 " The mapexprs in these are needed because cwd will make the command print out
@@ -27,9 +27,10 @@ function! neomake#makers#ft#go#go() abort
 endfunction
 
 function! neomake#makers#ft#go#golint() abort
+    " golint's issues are informational, as they're stylistic (not bugs)
     return {
         \ 'errorformat':
-            \ '%W%f:%l:%c: %m,' .
+            \ '%I%f:%l:%c: %m,' .
             \ '%-G%.%#'
         \ }
 endfunction
@@ -48,55 +49,16 @@ function! neomake#makers#ft#go#govet() abort
         \ }
 endfunction
 
-" This comes straight out of vim-go.
-function! neomake#makers#ft#go#Paths() abort
-    let dirs = []
-
-    if !exists('s:goroot')
-        if executable('go')
-            let s:goroot = substitute(system('go env GOROOT'), '\n', '', 'g')
-        else
-            let s:goroot = $GOROOT
-        endif
-    endif
-
-    if !empty(s:goroot) && isdirectory(s:goroot)
-        let dirs += [s:goroot]
-    endif
-
-    let workspaces = split($GOPATH, neomake#utils#path_sep())
-    if workspaces != []
-        let dirs += workspaces
-    endif
-
-    return dirs
-endfunction
-
-" This comes straight out of vim-go.
-function! neomake#makers#ft#go#ImportPath(arg) abort
-    let path = fnamemodify(resolve(a:arg), ':p')
-    let dirs = neomake#makers#ft#go#Paths()
-
-    let workspace = ''
-    for dir in dirs
-        if !empty(dir) && match(path, dir) == 0
-            let workspace = dir
-        endif
-    endfor
-
-    if empty(workspace)
-        return -1
-    endif
-
-    let srcdir = substitute(workspace . '/src/', '//', '/', '')
-    return substitute(path, srcdir, '', '')
-endfunction
-
-function! neomake#makers#ft#go#errcheck() abort
-    let path = neomake#makers#ft#go#ImportPath(expand('%:p:h'))
+function! neomake#makers#ft#go#gometalinter() abort
+    " Only run a subset of gometalinter for speed, users can override with:
+    " let g:let g:neomake_go_gometalinter_args = ['--disable-all', '--enable=X', ...]
+    "
+    " All linters are only warnings, the go compiler will report errors
     return {
-        \ 'args': ['-abspath', path],
+        \ 'args': ['--disable-all', '--enable=errcheck', '--enable=gosimple', '--enable=staticcheck', '--enable=unused'],
         \ 'append_file': 0,
-        \ 'errorformat': '%E%f:%l:%c:\ %m, %f:%l:%c\ %#%m'
+        \ 'cwd': '%:h',
+        \ 'mapexpr': 'neomake_bufdir . "/" . v:val',
+        \ 'errorformat': '%W%f:%l:%c:%m',
         \ }
 endfunction


bradleyfalzon:
Ok, pushed a second commit to allow customising of arguments passed to gometalinter.

I still feel the defaults I've provided are better than the defaults gometalinter provides for people that want linting on save as it reduces the amount of linters to those are the most important and therefore faster because there's less to execute and more suited to lint on save workflows. Although varcheck and structcheck could be removed as they don't lead to incorrect code just unused code.

autoload/neomake/makers/ft/go.vim
```diff
@@ -47,3 +47,19 @@ function! neomake#makers#ft#go#govet()
             \ '%-G%.%#'
         \ }
 endfunction
+
+let g:neomake_go_gometalinter_args =
+    \ get(g:, 'neomake_go_gometalinter_args', ['--disable-all',
+        \ '--enable=varcheck', '--enable=structcheck', '--enable=ineffassign',
+        \ '--enable=unconvert', '--enable=gosimple', '--enable=staticcheck'])
```
blueyed: `neomake_go_gometalinter_args` should work already out of the box.
You just need to specify the defaults below with the `'args'` key.

autoload/neomake/makers/ft/go.vim
```diff
@@ -47,3 +47,19 @@ function! neomake#makers#ft#go#govet()
...
+
+function! neomake#makers#ft#go#gometalinter()
+    return {
+        \ 'exe': 'gometalinter',
+        \ 'args': g:neomake_go_gometalinter_args,
+        \ 'append_file': 0,
+        \ 'cwd': '%:h',
+        \ 'mapexpr': 'neomake_bufdir . "/" . v:val',
```
blueyed: Well done for the `mapexpr`! :)

autoload/neomake/makers/ft/go.vim
```diff
@@ -47,3 +47,19 @@ function! neomake#makers#ft#go#govet()
...
+        \ 'errorformat': '%W%f:%l:%c:%m',
```
blueyed: Are there only warnings really?

autoload/neomake/makers/ft/go.vim
```diff
@@ -47,3 +47,19 @@ function! neomake#makers#ft#go#govet()
...
 endfunction
+
+let g:neomake_go_gometalinter_args =
+    \ get(g:, 'neomake_go_gometalinter_args', ['--disable-all',
+        \ '--enable=varcheck', '--enable=structcheck', '--enable=ineffassign',
+        \ '--enable=unconvert', '--enable=gosimple', '--enable=staticcheck'])
+
+function! neomake#makers#ft#go#gometalinter()
+    return {
+        \ 'exe': 'gometalinter',
```
blueyed: `exe` is by default the name, so you remove it.

autoload/neomake/makers/ft/go.vim
```diff
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#go#EnabledMakers()
-    return ['go', 'golint', 'govet']
+    return ['go', 'golint', 'govet', 'gometalinter']
```
blueyed: In case `gometalinter` is supposed to replace the others, you should only return it here, in case it is available:

```vim
return executable('gometalinter') ? ['gometalinter'] : ['go', 'golint', 'govet']
```

But I can see that `go` might just catch syntax errors etc.

autoload/neomake/makers/ft/go.vim
bradleyfalzon: Excellent, thanks.

autoload/neomake/makers/ft/go.vim
bradleyfalzon: Will do.

autoload/neomake/makers/ft/go.vim
bradleyfalzon: Totally copy/pasted ! But I'll take it :)

autoload/neomake/makers/ft/go.vim
bradleyfalzon: I would say yes, gometalinter simply copies the output from the various tools and all those tools have their own probability of false positives - if it compiles, then I think everything else is a warning.

bradleyfalzon:
Thanks, I'll update this over the next few days.

bradleyfalzon:
OK, so I've tested pushed through an update, can you please re-review?

1. I've made golint informational, they're stylistic issues which can be safely ignored.
1. I'm only running a very small subset of gometalinter, this is to focus on a good enough without being too slow experience. I'm happy to be challenged on this.
1. For Go projects, it's best to enable `neomake_serialize_abort_on_error` as the linters won't usually work without the code compiling (a few can, but they're limited without a type checker and that requires the program to compile) and then the linters themselves echo the same error as the go check already does. Is there a way to turn on abort on error by default for all Go files? Is that something we'd like to do globally? 
1. Is there somewhere appropriate I can write some docs about customising these options, in particular, the `neomake_go_gometalinter_args`, run on save and `neomake_serialize_abort_on_error`? To allow a user only interested in Go to get up a running quicker?

mhinz:
Sorry for the delay. If you're still with us, I'll make sure to get this merged quickly.

1. OK.
2. Sounds reasonable to me.
3. If `golint` and `govet` check for compilable code anyway, what about removing the `go` maker entirely then? Choosing the right set of enabled default makers is highly subjective anyway. I also think that this would be a sound default:

    ```go
    return executable('gometalinter') ? ['gometalinter'] : ['golint', 'govet']
    ```

4. People who disagree with defaults will simply have to read the documentation. I think it's not feasible to add filetype-specific sections to it.

tweekmonster:
> Is there somewhere appropriate I can write some docs about customising these options

It might be reasonable to add a section for maker-specific configs just above the FAQ.  Though, Neomake has a lot of configs on its own, so it'll probably get a little long if that section grows.  A new doc file just for maker-specific configs is another option.

In either case, it'll be a good idea to add links to that section (or file) at the top of certain sections (`neomake-configuration`, `neomake-options`, etc.) for people still figuring things out.

There's a couple makers that probably should have documentation (I'm just mentioning it for posterity, not saying it should be done in this PR).  From a quick grep, it looks like `java`, `d`, `sh`, and `php` have non-standard configs that should be documented as well.  I can't figure out the `sh` variables though.  `g:is_sh` are a strange choice.

...and after I typed that I find `runtime/filetype.vim` and `runtime/syntax/sh.vim`

<sup>*`@tweekmonster shakes his fist at the screen`*</sup>

autoload/neomake/makers/ft/go.vim
```diff
@@ -47,3 +48,17 @@ function! neomake#makers#ft#go#govet()
...
+    " Only run a subset of gometalinter for speed, users can override with:
+    " let g:let g:neomake_go_gometalinter_args = ['--disable-all', ....]
+    "
+    " All linters are only warnings, the go compiler will report errors
+    return {
+        \ 'args': ['--disable-all', '--enable=gosimple', '--enable=staticcheck'],
+        \ 'append_file': 0,
+        \ 'cwd': '%:h',
+        \ 'mapexpr': 'neomake_bufdir . "/" . v:val',
+        \ 'errorformat': '%W%f:%l:%c:%m',
```
novas0x2a: Here's a slightly better efm: `'%W%f:%l:%c:%t%*[^:]: %m',`
It distinguishes between errors and warnings (the only two severities gometalinter supports natively right now)

bradleyfalzon: Thanks @novas0x2a, how exactly is this an improvement? (Serious question, isn't meant to sound sarcastic).

bradleyfalzon:
Note, I am going to get back to this, holiday season and all...

autoload/neomake/makers/ft/go.vim
novas0x2a: ah, just what i said above, it lets vim know whether it's an error or a warning. Example:
before:
```
version.go|16 error| syntax error: non-declaration statement outside function body
version.go|6 col 2 warning| warning: "fmt" imported but not used (errcheck)
version.go|6 col 2 warning| warning: "fmt" imported but not used (errcheck)
version.go|16 col 1 warning| warning: expected declaration, found 'IDENT' wat (golint)
version.go|16 col 1 warning| error: expected declaration, found 'IDENT' wat (vet)
version.go|16 col 1 warning| warning: expected declaration, found 'IDENT' wat (errcheck)
```

after:
```
version.go|16 error| syntax error: non-declaration statement outside function body
version.go|6 col 2 warning| "fmt" imported but not used (errcheck)
version.go|6 col 2 warning| "fmt" imported but not used (errcheck)
version.go|16 col 1 warning| expected declaration, found 'IDENT' wat (golint)
version.go|16 col 1 error| expected declaration, found 'IDENT' wat (vet)
version.go|16 col 1 warning| expected declaration, found 'IDENT' wat (errcheck)
```

Differences to note:
1) gometalinter's duplicated severity prefix ("error:" or "warning:") is stripped out of the error message, leaving only vim's error/warning category (which can be syntax-styled as usual)
2) vim's error/warning designation is now correct on line 5 (gometalinter reports it as an error, but the previous efm marks everything as a warning because that's the default error type (`%t`); after my change, vim and gometalinter agree)

bradleyfalzon: Yes, I see what you mean, but IMHO static analysis tools aren't errors, they're all warnings, the only real errors are from the Go compiler.

Thoughts?

novas0x2a: The name "gometalinter" is kind of a misnomer, it's not really a linter, it's more of a consistent interface for quite a lot of tools, some of which happen to be linters. GMT already [classifies](https://github.com/alecthomas/gometalinter/blob/master/config.go#L134) the severity of the outputs, and the defaults are quite sensible; it also allows for user overrides of these (`--errors`, `--severity` etc). If you don't like that it reports things as errors, you can override the severities in the default args- I don't think it's possible to pick a set of default args that work for everyone, anyway- personally, I use 
```
    \     '--disable-all',
    \     '-E', 'vet',
    \     '-E', 'golint',
    \     '-E', 'errcheck',
    \     '--message-overrides', 'errcheck:{message}',
    \     '--sort', 'line',
    \     '--tests',
    \     '--exclude', 'bindata',
    \     '--exclude', '.pb.',
    \     '--enable-gc'
```
so optimizing for changing the args and capturing all information via efm is more important to me personally.

blueyed:
@bradleyfalzon 
Gently ping.. :)

bradleyfalzon:
OK, just to catch up, it's clear exactly what is required.

I'll first further respond to @mhinz https://github.com/neomake/neomake/pull/795#issuecomment-268335550

> 3. If golint and govet check for compilable code anyway, what about removing the go maker entirely then?

They may complain about some of the same errors, but they're can't complain about them all. So it's best to defer to go compiler for the actual errors, and linters/tools as warnings.

> 4. People who disagree with defaults will simply have to read the documentation. I think it's not feasible to add filetype-specific sections to it. 

This I didn't completely understand exactly what you would like me to do.

But ignore all that, I've been using this branch for a good couple of months now, and when there's a Go compiler error, all the linters etc are useless. The main issue, is the compiler will be complaining about a different part of the file but the linters show additional compiler warnings, so it's a little misleading where the error actually is.

In summary, I believe `neomake_serialize_abort_on_error` should be enabled by default on all Go projects, if we can't do this, then I think we should suggest it to the user and if we're going to suggest it to the user, perhaps also suggest how to customise neomake's go integration by settings appropriate `neomake_go_gometalinter_args` values.

I don't mind what the decision is on the above, and sorry if it was already made clear.

In the meantime, I'll fix the merge conflict.

bradleyfalzon:
Also, my rebase essentially reverts @higebu's 12b5f33cd3f662de32d92cda34019fc0603336ba I've chosen to do this because gometalinter can run errcheck as well, and I've added that as a default. I chose to implement gometalinter as this allows neomake to not need to implement every linter/tool available, and instead use gometalinter to manage and normalise these checks for it.

blueyed:
> neomake_serialize_abort_on_error

It is also used with the python maker, so I think it is good.

bradleyfalzon:
Actually, since rebasing, I've noticed the `golint` checks now highlight the first letter, whereas `go vet` checks aren't? See the `f` in `func`

screen shot 2017-01-30 at 9 19 04 am

bradleyfalzon:
Another issue i've been having, but seems to be more often since the rebase (but definitely happened before), sometimes I get this error:

```
Error detected while processing function neomake#highlights#ShowHighlights[4]..<SNR>30_ResetHighlights:
line    6:
E803: ID not found: 8
```

I can reproduce it with:
```bash
cat > main.go <<EOF
package main

import "fmt"

func main() {
        fmt.Println("Hello %s", "fred") // go vet warning
}
EOF

mkdir sub

cat > sub.go <<EOF
package sub

import "fmt"

func Main() {  // golint warning
        fmt.Println("vim-go %s", "fred") // go vet warning
}
EOF
```

```
vim main.go
:w
<wait for go vet warning to appear>
tabnew sub/sub.go
:w
<wait for go vet warning to appear>
:q
```

vimrc:
```
" Run Neomake on save

let g:neomake_go_gometalinter_args = ['--disable-all', '--enable=gosimple', '--enable=unused', '--enable=staticcheck']

"let g:neomake_verbose = 3
"let g:neomake_logfile = '/tmp/neomake'

" serialize and abort on error has the effect of not running linters when
" there's a compiler error
let g:neomake_serialize = 1
let g:neomake_serialize_abort_on_error = 1
```

tested with vim 8.0.271 and the branch from this PR, I've disabled all other plugins (except Vundle which was loading neomake). Above is a snippet of my vimrc, hopefully that's enough to reproduce the error.

If you don't have Go installed, you might need to:

- Get https://storage.googleapis.com/golang/go1.7.5.linux-amd64.tar.gz
- Extract to /usr/local/go
- mkdir ~/go
- export GOPATH=~/go
- Put the test files (main.go and sub/sub.go) in: $GOPATH/src/test

blueyed:
@bradleyfalzon 
Thanks for the test case, which could be transferred into a Vader test probably.
Before investigating, I suggest to rebase this on master though first.

As for the highlighting: it will just have a column by default, and `length` needs to be added by a postprocessing function.  https://github.com/neomake/neomake/pull/1030 is meant to improve this in a generic way.

bradleyfalzon:
Thanks @blueyed, I've rebased.

Having said this, and assuming the tests pass, is there any further changes/thoughts?

autoload/neomake/makers/ft/go.vim
```diff
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#go#EnabledMakers() abort
-    return ['go', 'golint', 'govet']
+    return ['go', 'golint', 'govet', 'gometalinter']
```
blueyed: Should gometalinter be preferred over other makers, if it is available/executable?

autoload/neomake/makers/ft/go.vim
```diff
@@ -48,55 +49,16 @@ function! neomake#makers#ft#go#govet() abort
...
+    " let g:let g:neomake_go_gometalinter_args = ['--disable-all', '--enable=X', ...]
+    "
+    " All linters are only warnings, the go compiler will report errors
     return {
-        \ 'args': ['-abspath', path],
+        \ 'args': ['--disable-all', '--enable=errcheck', '--enable=gosimple', '--enable=staticcheck', '--enable=unused'],
         \ 'append_file': 0,
-        \ 'errorformat': '%E%f:%l:%c:\ %m, %f:%l:%c\ %#%m'
+        \ 'cwd': '%:h',
+        \ 'mapexpr': 'neomake_bufdir . "/" . v:val',
```
blueyed: Should use single quotes.

autoload/neomake/makers/ft/go.vim
bradleyfalzon: I don't think it should be preferred, it just runs those same tools.

My other change, is to set golint errors as informational, as they are stylistic. Separating golint from gometalinter makes this check straightforward.

It's certainly possible to combine them, but I don't really see a lot of the value, especially if you want to support both.

It maybe possible to drop golint and combine that into gometalinter, but I'm not see a lot of the value (other than potentially removing `neomake#makers#ft#go#golint`).

As go vet is available with all Go commands, if a user does not have gometalinter installed, they at least get go vet out of the box.

bradleyfalzon:
Rebased but the build failures do seem to be unrelated to my change.

JelteF:
Anything holding this up?

blueyed:
Needs to be rebased, which I could do.

@bradleyfalzon 
Is it ready from your side?

bradleyfalzon:
If by debase you mean squashing those commits, then I'm fine if you want to do it on merge. If you prefer the merge commit, I can do it in a few hours. But as far as I'm concerned this is ready and I've been actively using it since proposed, so happy with the behaviour.

blueyed:
There was a conflict, so it had to be rebased.
I've squashed it into a single commit, and will merge it once it's green on Travis.
Thanks!

autoload/neomake/makers/ft/go.vim
blueyed: I think this makes sense, and that the errorformat should be adjusted according to what @novas0x2a suggested (`'%W%f:%l:%c:%t%*[^:]: %m'`).
@bradleyfalzon 
Do you still think everything should be hardcoded as warning?!

blueyed:
> In summary, I believe neomake_serialize_abort_on_error should be enabled by default on all Go projects

For all linters, or just `neomake#makers#ft#go#go`?
Should we keep `go` then even when preferring `gometalinter`?
Does `gometalinter` report compile errors after all?

autoload/neomake/makers/ft/go.vim
blueyed: > I don't think it should be preferred, it just runs those same tools.

But it means that checks are then run twice / duplicated, no?
It really sounds like `gometalinter` should be enabled alone by default, if it is installed?! (given that it also reports compile errors)

autoload/neomake/makers/ft/go.vim
blueyed: Added to https://github.com/neomake/neomake/pull/1213.

autoload/neomake/makers/ft/go.vim
bradleyfalzon: They won't always report compile errors, as some of them will just say "can't load project", others will point out the exact line number - but in all cases, only the go compiler is authoritative on what's a compile error - the others use similar but not exactly the same tools.

blueyed: > the go compiler

Is that what `neomake#makers#ft#go#go` runs?

bradleyfalzon: Yes.

blueyed: Then the first two commits of https://github.com/neomake/neomake/pull/1213 make sense..  let's discuss it there further.

bradleyfalzon:
I'd say just for `neomake#makers#ft#go#go`.

In summary, some, but not all linters can report compile errors. Only the compiler should report compile errors, the other linters ran by gometalinter aren't always exact - and a user might have configured gometalinter to use only certain tools that might not report compile errors.

It's for this reason that I also only consider compile errors an error. All other linters are simply warnings that need a humans attention, and golint itself is just stylistic (hence the issues are informational).

Once there is a compile error, all linters should be stopped, and just the compile errors should be addressed.

This is why I believe `let g:neomake_serialize = 1` and `let g:neomake_serialize_abort_on_error = 1` should be the default.

PR #797, opened 2016-11-22
python: fix pydocstyle (former pep257) errorformat

blueyed:


blueyed:
You should still be able to use it, it's just not enabled by default anymore.

We could use `executable('pydocstyle') ? 'pydocstyle' : 'pep257'` though for `GetEnabledMakers`?

mhinz:
> You should still be able to use it, it's just not enabled by default anymore.

But it would try using `pydocstyle` instead of `pep257`.

And according to your commit both executables use different errorformats, so having 2 makers for them makes sense to me.

Maybe we need a way to mark outdated makers as deprecated. So in the case of `pep257` a warning is emitted: "pep257 is outdated, switch to pydocstyle instead`.

autoload/neomake/makers/ft/python.vim
```diff
@@ -118,15 +118,19 @@ function! neomake#makers#ft#python#Pep8EntryProcess(entry) abort
 endfunction
 
 function! neomake#makers#ft#python#pydocstyle() abort
-    return {
+  if !exists('s:_pydocstyle_exe')
+    let s:_pydocstyle_exe = executable('pydocstyle') ? 'pydocstyle' : 'pep257'
```
mhinz: Does the same errorformat work for both of them?

blueyed: Yes, it is the same tool, which was renamed.. :)

blueyed:
> And according to your commit both executables use different errorformats, so having 2 makers for them makes sense to me.

?

mhinz:
The commit removed `pep257` and introduced a new errorformat, so I assumed that `pydocstyle` was a new version of `pep257` with different output. But apparently not.

LGTM. :]

blueyed:
Sorry for the confusion, and thanks for the review! :)

Issue #783, opened 2016-11-18
question regarding alex maker for markdown


mhartington:
Hi, I'm working on adding a new maker for markdown using [alexjs](http://alexjs.com/)
So far, testing locally before making a PR. Right now I have this .

```
  let g:neomake_markdown_alex_maker = {
                \ 'errorformat':
                \ '%W%f:%l:%c: warning: %m'
                \}
  let g:neomake_markdown_enabled_makers = ['alex']

```

But using the verbose output, it seems that neomake is removing the message as "invalid entry"?

### Expected behavior
Errors/warning pop up, of course 😄 

### Steps to reproduce

```
npm install -g alex
touch test.md
```

I copied the sample text from the alex site and used that as the content for `test.md`

### Output of the ":verb NeomakeInfo" command

```
Neomake [2.392]: [1] Running makers: ['alex']
Neomake [2.393]: [1] Starting async job: ['alex', '/Users/mhartington/test.md']
Neomake [3.170]: [1.6] stderr: alex: ['test.md', '   1:5-1:14  warning  `boogeyman` may be insensitive, use `boogey` instead boogeyman-boogeywoman  retext-equality', '  1:42-1:48  warning  `master` / `slaves` may be insensitive, use `primary` / `replica` instead  master-slave retext-equality', '  2:52-2:54  warning  `he` may be insensitive, use `they`, `it` instead he-she retext-equality', '  2:61-2:68  warning  `cripple` may be insensitive, use `person with a limp` instead cripple retext-equality', '', '⚠ 4 warnings', '']
Neomake [5.994]: [1.6] exit: alex: 1
Neomake [5.995]: [1.6] alex: processing 7 lines of output.
Neomake [5.995]: [1.6] Removing invalid entry: test.md ({'lnum': 0, 'bufnr': 0, 'col': 0, 'valid': 0, 'vcol': 0, 'nr': -1, 'type': '', 'maker_name': 'alex', 'pattern': ''})
Neomake [9.610]: [1.6] Removing invalid entry:    1:5-1:14  warning  `boogeyman` may be insensitive, use `boogey` instead boogeyman-boogeywoman  retext-equality ({'lnum': 0, 'bufnr': 0, 'col': 0, 'valid': 0, 'vcol': 0, 'nr': -1, 'type': '', 'maker_name': 'alex', 'pattern': ''})
[...]

PR #799, opened 2016-11-23
feat(): add alex maker

mhartington:
Ref #783


mhinz:
:+1:

We can add it to the default makers for Markdown when it gets more mature.

mhinz:
Merged (with some slight changes).

PR #802, opened 2016-11-26
clangcheck: removed flags not accepted by clang-check

ichigo663:
The extra cmd line flags "-- -Wall -Wextra" are not accepted and
processed by clang-check, moreover they actually prevent it from working
correctly.


blueyed:
Thanks!

PR #807, opened 2016-12-02
Add bufnr into the hook context of NeomakeFinished

wookayin:
Let the `g:neomake_hook_context` dictionary on the event 'NeomakeFinished' contain `bufnr`.

Tests and docs are updated accordingly.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -532,10 +532,10 @@ function! s:Make(options, ...) abort
         endif
     endfor
     if !len(job_ids)
-        let l:hook_context = {'file_mode': file_mode,
-                            \ 'bufnr': buf,
-                            \}
-        call neomake#utils#hook('NeomakeFinished', l:hook_context)
+        call neomake#utils#hook('NeomakeFinished', {
+                    \ 'file_mode': file_mode,
+                    \ 'bufnr': buf,
+                    \ })
     endif
     return job_ids
 endfunction
@@ -874,10 +874,10 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
         " Trigger autocmd if all jobs for a s:Make instance have finished.
         if neomake#has_async_support()
             if !len(filter(copy(s:jobs), 'v:val.make_id == jobinfo.make_id'))
-                let l:hook_context = {'file_mode': maker.file_mode,
-                                    \ 'bufnr': get(jobinfo, 'bufnr', -1),
-                                    \}
-                call neomake#utils#hook('NeomakeFinished', l:hook_context)
+                call neomake#utils#hook('NeomakeFinished', {
+                            \ 'file_mode': maker.file_mode,
+                            \ 'bufnr': get(jobinfo, 'bufnr', -1),
+                            \ })
             endif
         endif
     endif
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -532,8 +532,10 @@ function! s:Make(options, ...) abort
         endif
     endfor
     if !len(job_ids)
-        call neomake#utils#hook('NeomakeFinished', {
-                    \ 'file_mode': file_mode})
+        let l:hook_context = {'file_mode': file_mode,
+                            \ 'bufnr': buf,
+                            \}
+        call neomake#utils#hook('NeomakeFinished', l:hook_context)
     endif
     return job_ids
 endfunction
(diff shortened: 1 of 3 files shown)


autoload/neomake.vim
```diff
@@ -532,8 +532,10 @@ function! s:Make(options, ...) abort
         endif
     endfor
     if !len(job_ids)
-        call neomake#utils#hook('NeomakeFinished', {
-                    \ 'file_mode': file_mode})
+        let l:hook_context = {'file_mode': file_mode,
+                            \ 'bufnr': buf,
+                            \}
+        call neomake#utils#hook('NeomakeFinished', l:hook_context)
```
blueyed: I'd prefer to not use local vars for this.

autoload/neomake.vim
```diff
@@ -872,8 +874,10 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
         " Trigger autocmd if all jobs for a s:Make instance have finished.
         if neomake#has_async_support()
             if !len(filter(copy(s:jobs), 'v:val.make_id == jobinfo.make_id'))
-                call neomake#utils#hook('NeomakeFinished', {
-                            \ 'file_mode': maker.file_mode})
+                let l:hook_context = {'file_mode': maker.file_mode,
+                                    \ 'bufnr': get(jobinfo, 'bufnr', -1),
+                                    \}
+                call neomake#utils#hook('NeomakeFinished', l:hook_context)
```
blueyed: Same: I'd prefer to not use a local var for this.

wookayin: Okay, I just presumed some optional attributes might be added conditionally and dynamically, but it is not yet. I will modify a little so that the dictionary is directly passed as an argument.

autoload/neomake.vim
```diff
@@ -873,7 +875,9 @@ function! neomake#MakeHandler(job_id, data, event_type) abort
         if neomake#has_async_support()
             if !len(filter(copy(s:jobs), 'v:val.make_id == jobinfo.make_id'))
                 call neomake#utils#hook('NeomakeFinished', {
-                            \ 'file_mode': maker.file_mode})
+                            \ 'file_mode': maker.file_mode,
+                            \ 'bufnr': get(jobinfo, 'bufnr', -1),
```
blueyed: It is bad that there is no test that gets the -1.  Can it happen?  Otherwise we should remove the default from `get()`.
Do you feel like adding one?

wookayin: I am not sure whether `jobinfo.bufnr` always exists, as I am not familiar with the core internals of neomake (e.g. the anatomy of `jobinfo`). Could you please clarify to me?

In L855-L857, if `file_mode` is 0, `jobinfo.bufnr` is not used, which is the reason why I coded so. Please see also L693-695 where the original code reads `get(maker, 'bufnr', -1)`; it is about `maker.bufnr`.

However, it looks that `jobinfo.bufnr` is set whenever I tested, even if `file_mode` is 0. If you believe `jobinfo` always contains `bufnr`, I would fix by removing default as you suggested.

wookayin:
Moreover, it needs to be checked whether the callback-ed `bufnr` corresponds *exactly* to the buffer that `:Neomake` is invoked on (I observed some discrepancy in some incidental cases). A more sophiscated unit tests are needed as well.

I will investigate this soon. Please do not merge until we get convinced on this behavior.

blueyed:
Yes, I have some changed in some deep-down stashes in this regard IIRC.. :/

blueyed:
FWIW: I've used just `jobinfo` for the new `NeomakeJobFinished` hook, and moved `file_mode` from the maker to `jobinfo`.
`jobinfo` is basically meant to hold information about the job, which includes the maker - and I consider `bufnr` part of the job, not the `maker` object (which should be more static).
I am not so sure about `file_mode` - but it's possible that a maker can be run both in project and file mode, at least in theory.  Therefore => jobinfo.

blueyed:
@wookayin 
Any update on this?

blueyed:
I will likely go with passing in jobinfo to all hook, when it is available.

wookayin:
Now it is available with `context.options.bufnr` (just a memo).

blueyed:
Related #1587.

I'm seeing sometimes errors where `g:neomake_hook_context` was unset already (from parallel handling then).

Issue #336, opened 2016-03-05
statusline functions return nothing


nhooyr:
The functions for the status line return nothing for me, no matter what. No clue what's wrong. The counts functions all return empty as well even though there are errors in the loclist and quick fix list.

```vim
call plug#begin('.')
Plug 'benekastah/neomake'
call plug#end()
set statusline+=%{neomake#statusline#QflistStatus('qf:\ ')}%{neomake#statusline#QflistStatus('qf:\ ')}
```

I just tested by opening the neovim source, deleting a few lines for some errors and nothing changed on the status line after I ran `:Neomake!`

motdef:
I have the same problem. ```neomake#statusline#LoclistStatus``` returns an empty string even if I deliberately introduce errors to populate the location list.

SageEx:
Same here.

SageEx:
Any updates on this ?

thalesmello:
I suspect it has to do with the maker syntax.

I get error counts for errors in Python files, but not error counts for JavaScript files (using 'standard' maker).

Analyzing the maker syntax for standard:
```
function! neomake#makers#ft#javascript#standard()
    return {
        \ 'errorformat': '  %f:%l:%c: %m'
        \ }
endfunction
```

Analyzing the syntax for pep8:
```
function! neomake#makers#ft#python#python()
    return {
        \ 'args': [ '-c',
            \ "from __future__ import print_function\r" .
            \ "from sys import argv, exit\r" .
            \ "if len(argv) != 2:\r" .
            \ "    exit(1)\r" .
            \ "try:\r" .
            \ "    compile(open(argv[1]).read(), argv[1], 'exec', 0, 1)\r" .
            \ "except SyntaxError as err:\r" .
            \ "    print('%s:%s:%s: %s' % (err.filename, err.lineno, err.offset, err.msg))"
        \ ],
        \ 'errorformat': '%E%f:%l:%c: %m',
        \ }
endfunction
```
The error message stards with a `%E`.
In fact, if I place

```
let g:neomake_javascript_standard_maker = { 'errorformat': '%E %f:%l:%c: %m' }
```

In my Vimfile, the statusline functions start working.

Therefore, I think there is a problem with Neomake regarding the case when there is no Warning nor Error indication in the `errorformat`. Maybe it should default to error or something like that.

SageEx:
The development seems to have stated again. Please fix this issue.
Thanks.

blueyed:
@SageEx 
[...]

PR #810, opened 2016-12-04
tslint maker: use %E in errorformat

Freak613:
Same as here: https://github.com/neomake/neomake/pull/532
I was able to get airline support only when enabled %E in error format
And it's related to https://github.com/neomake/neomake/issues/336


blueyed:
Thanks!

Issue #805, opened 2016-12-01
Visual selection gets lost when error list gets updated


blueyed:
The visual selection gets lost when a maker finishes and the location list gets opened / changed.

This seems to be more an issue with Vim/Neovim (which has no command to really only open the window without going there), but we might want to postpone the update of the errors until there is no visual selection anymore?!

TEST CASE:
 - `let g:neomake_open_list = 2`
 - In a file with errors: `Neomake | norm! V`

bounceme:
strange, how about just `norm! gv` after the errors are updated?

blueyed:
Does not work.. it will only have a partial selection then.
I am trying to fix this in https://github.com/neomake/neomake/pull/831: the current idea being to only process output when in normal and insert mode.

PR #814, opened 2016-12-06
`g:neomake_open_list`: Prevent original window from jumping around

nfnty:


blueyed:
Thanks!

How and when is it jumping around for you, and what is your config for `neomake_open_list`?

Given the async nature however, we should be extra careful here (and not jump also for `open_val` already).
Instead of the `wincmd p`, what about `exe w "wincmd w"`, where w is the window ID, if available?
See `win_getid()`, `win_gotoid()` etc.
Normal window numbers and window IDs should be compatible (the latter starting at 1000).

A test for this would be nice, but if you can provide a manual test case, we can come up with a Vader test together.

See also https://github.com/neomake/neomake/issues/805, and https://github.com/neomake/neomake/projects/1.

nfnty:
The problem I'm referring to is when lines of the original window jump to a new position with respect to the terminal window when opening new splits.

This should probably be fixed by new autocommand events like `WinResizePre` and `WinResizePost`. Not by implementing this in every plugin that opens a new window.

blueyed:
So I guess you are not using `set noequalalways` then - which seems to be a very non-default setting, but I am using it.

nfnty:
That's not the problem. Here you can see what I'm referring to:

***Without this PR***
vim_bad

***With this PR***
vim_good

blueyed:
I would need to investigate - I am not using `neomake_open_list` myself, and `set noequalalways`, which might be relevant in this context: can you try it?

Apart from that I am OK with merging this for now as-is.

nfnty:
Don't merge this yet, I'm working on making a real fix for neovim instead.

blueyed:
Is it only a problem with Neovim then, but not Vim?

nfnty:
No, this is a problem that's always been there, I've just not used quickfix windows extensively until now.

blueyed:
I've tried to reproduce this today, and for me the view was jumping back after closing the location/quickfix list (after going up when it opened)?!  (but my config might be involved, similar to yours)
This PR however would improve the behavior, but it really looks more like a Vim/Neovim and/or config issue.

I'm looking forward to your findings.. :)

blueyed:
This is likely fixed/improved by https://github.com/neomake/neomake/commit/d786a43c88b801ac5bf123733b8ffbad7d08d3b9.

PR #816, opened 2016-12-07
added crystal as a maker for crystal

sbdchd:


autoload/neomake/makers/ft/crystal.vim
```diff
@@ -0,0 +1,16 @@
+function! neomake#makers#ft#crystal#EnabledMakers()
```
mhinz: Please put this modeline at the top of the file: `" vim: ts=4 sw=4 et`

blueyed: I think a lot of the files have it at the bottom, and we do not have a policy for that.
IMHO we should not have them in general, but rather a .editorconfig / .lvimrc file.

blueyed: Oh, I thought it was placed wrongly.
So, LGTM without a modeline.

mhinz: `.editorconfig`? `.lvimrc`? That probably requires extra plugins?

blueyed: https://github.com/embear/vim-localvimrc is awesome - and I like it for project specific settings.

https://github.com/editorconfig/editorconfig-vim#readme would be for `.editorconfig`, but I do not like the overhead.

We have really messy style in the maker vim files, and therefore it does not really matter (currently) what the style there is.
Adding a modeline, which takes a lot of relative space in general, does not help here.
We should rather enforce some linting here, if that becomes an issue.

blueyed:
Thanks, @sbdchd!

PR #822, opened 2016-12-09
Fix style with autoload/neomake/makers/ft/{s,}css.vim

blueyed:


davidosomething:
that's the way i like it :)

davidosomething:
although i prefer to put the concat operator on the next line so you can delete the line at will without having a stray `.`

blueyed:
@davidosomething 
I think you can pick any other ft maker plugin file and do the same.. ;) - I just reviewed those and then did it.

blueyed:
> although i prefer to put the concat operator on the next line so you can delete the line at will without having a stray `.`

I'll keep that in mind and tend to do it myself, but I did not want to refactor it.. ;)

blueyed:
@davidosomething 
The downside however is that you need to indent the first line to align the errorformat entries, so I am rather -0 on this in this context.

PR #825, opened 2016-12-09
maker: elixir#mix: remove getcwd()

blueyed:
The current directory is the default, and should rather be set during
evaluation.


blueyed:
Thanks!

Issue #741, opened 2016-10-26
Quickfix/location list should not be cleared when starting makers


blueyed:
I think that especially with long running makers it's better to keep the old results around, until the first maker finishes.

Should this be made configurable?!

teto:
Or trigger an autocmd NeomakeStarted ? I could use that on projects that I know take a long time to display into i3bar for instance "Compiling".

blueyed:
> Or trigger an autocmd NeomakeStarted

Good idea.  This would be triggered instead of the NeomakeCountsChanged that is being used currently (when resetting, and counts get reset because of this).

blueyed:
I think this should be fixed/addressed before #831 to fix the first TODO there.

blueyed:
Can you test #981, please?

PR #831, opened 2016-12-11
Refactor output processing to only be done in insert/normal mode

blueyed:
https://github.com/neomake/neomake/issues/805

TODO
 - [x] only trigger NeomakeFinished after output has been processed, and/or add a new User autocommand?  My use case is automatically closing the location/quickfix list if they are empty.  But currently they will be in case of delayed processing, since they were cleared in the beginning.  The fix for this could also be https://github.com/neomake/neomake/issues/741.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -16,6 +16,7 @@ let s:need_errors_cleaning = {
 let s:maker_defaults = {
             \ 'buffer_output': 1,
             \ 'remove_invalid_entries': 1}
+let s:project_job_output = []
 
 function! neomake#has_async_support() abort
     return has('nvim') ||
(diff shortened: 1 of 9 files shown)


autoload/neomake.vim
```diff
@@ -737,15 +770,40 @@ function! s:GetTabWinForJob(job_id) abort
...
 function! s:RegisterJobOutput(jobinfo, lines, source) abort
-    let lines = copy(a:lines)
     let maker = a:jobinfo.maker
 
     if !get(maker, 'file_mode')
-        call s:ProcessJobOutput(a:jobinfo, lines, a:source)
-        call neomake#signs#PlaceVisibleSigns()
+        if s:CanProcessJobOutput()
+            call s:ProcessJobOutput(a:jobinfo, a:lines, a:source)
+            call neomake#signs#PlaceVisibleSigns()
```
blueyed: I've noticed that `call neomake#highlights#ShowHighlights()` might be missing here?!
/cc @paretje (via ca87544)

Issue #837, opened 2016-12-13
Feature request: maker value to not have args shellescaped


bounceme:
`emacs path-to-broken.js --quick --batch --eval="(progn(package-initialize)(require 'js2-mode)(js2-mode)(js2-reparse t)(js2-display-error-list)(princ(with-current-buffer \"*js-lint*\"(buffer-substring-no-properties(point-min)(point-max)))))" 2>/dev/null` 

If you run that command, and you have emacs with js2-mode installed the errors printed to stdout will look like:
```
line 2: Undeclared variable or function 'f'
line 2: missing ; after statement
line 2: Undeclared variable or function 'exports'
line 2: Undeclared variable or function 'module'
line 3: Undeclared variable or function 'module'
line 4: syntax error
```
here is my attempt at a maker (which failed):
```
let g:neomake_javascript_emacs_maker = {
      \ 'args': ['%:p','--quick','--batch',
      \ '--eval="(progn(package-initialize)(require ''js2-mode)(js2-mode)(js2-reparse t)(js2-display-error-list)(princ(with-current-buffer \"*js-lint*\"(buffer-substring-no-properties(point-min)(point-max)))))"',
      \ '2>/dev/null'],
      \ 'errorformat': '%f line %l: %m',
      \ 'mapexpr': 'expand("%:p") . " " . v:val',
      \ }
```
any ideas?

bounceme:
maybe some extra shell quoting happens to the args? it doesn't seem to invoke emacs at all

mhinz:
Yup, `'--eval="(progn(package-initialize)(require ''js2-mode)(js2-m....` looks suspicious.

Should probably be something like `"--eval=\"(progn(package-initialize)(require '...`.

bounceme:
thanks, i'll try later

bounceme:
`      \ "--eval=\"(progn(package-initialize)(require 'js2-mode)(js2-mode)(js2-reparse t)(js2-display-error-list)(princ(with-current-buffer \\\"*js-lint*\\\"(buffer-substring-no-properties(point-min)(point-max)))))\"",
`
that didn't work either. my next guess is the elisp function not terminating output with a `\n`

bounceme:
that isn't it but i fixed the newline issue ( if it is an issue ):
```
[...]

PR #840, opened 2016-12-14
jobstart argv as string

bounceme:
this doesn't address non-async however.


bounceme:
#837

bounceme:
hmm, this would be a more substantial refactoring I think

PR #841, opened 2016-12-14
add option to prevent additional arg quoting

bounceme:
#837

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -231,10 +231,10 @@ function! s:MakeJob(make_id, maker) abort
         else
             call neomake#utils#DebugMessage('Running synchronously')
             if has_args
-                if neomake#utils#IsRunningWindows()
+                if neomake#utils#IsRunningWindows() || noquote
                     let program = exe.' '.join(map(args, 'v:val'))
                 else
-                    let program = exe.' '.join(map(args, noquote ? 'v:val' : 'shellescape(v:val)'))
+                    let program = exe.' '.join(map(args, 'shellescape(v:val)'))
                 endif
             else
                 let program = exe
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -143,6 +143,8 @@ function! s:MakeJob(make_id, maker) abort
         endtry
     endif
 
+    let noquote = get(a:maker,'noquote')
+
     try
         let has_args = type(args) == type([])
         let error = ''
@@ -161,7 +163,7 @@ function! s:MakeJob(make_id, maker) abort
                     call neomake#utils#LoudMessage(printf(
                                 \ 'Starting async job: %s',
                                 \ string(argv)), jobinfo)
-                    let job = jobstart(argv, opts)
+                    let job = jobstart(noquote ? join(argv) : argv, opts)
                 catch
                     let error = printf('Failed to start Neovim job: %s: %s',
                                 \ string(argv), v:exception)
@@ -232,7 +234,7 @@ function! s:MakeJob(make_id, maker) abort
                 if neomake#utils#IsRunningWindows()
                     let program = exe.' '.join(map(args, 'v:val'))
                 else
-                    let program = exe.' '.join(map(args, 'shellescape(v:val)'))
+                    let program = exe.' '.join(map(args, noquote ? 'v:val' : 'shellescape(v:val)'))
                 endif
             else
                 let program = exe


bounceme:
will add documentation as needed

bounceme:
forgot vim-async part

blueyed:
What about not quoting strings, but only arrays?  (i.e. if `args` is already a string leave it as-is)

bounceme:
i'm just concerned about other makers, I tried something like that and broke vint

bounceme:
I think some makers depend on the extra quoting, vint for example

bounceme:
I can't find a way to get vim-async to run my js2 maker.

blueyed:
Hmm, vint returns an array for `args`?
To have it return a string you would have to change it:
```diff
diff --git i/autoload/neomake/makers/ft/vim.vim w/autoload/neomake/makers/ft/vim.vim
index e6d2e8e..00e719a 100644
--- i/autoload/neomake/makers/ft/vim.vim
+++ w/autoload/neomake/makers/ft/vim.vim
@@ -13,7 +13,7 @@ function! neomake#makers#ft#vim#vint() abort
     endif
 
     return {
-        \ 'args': l:args,
+        \ 'args': join(map(l:args, 'shellescape(v:val)')),
         \ 'errorformat': '%I%f:%l:%c:style_problem:%m,%f:%l:%c:%t%*[^:]:%m'
         \ }
 endfunction
```


And as for the change I had in mind:

```diff
diff --git i/autoload/neomake.vim w/autoload/neomake.vim
index b0b7b97..82f8964 100644
--- i/autoload/neomake.vim
+++ w/autoload/neomake.vim
@@ -121,12 +121,16 @@ function! s:MakeJob(make_id, maker) abort
     else
         let args = a:maker.args
     endif
-    let append_file = a:maker.file_mode && index(args, '%:p') == -1 && get(a:maker, 'append_file', 1)
-    if append_file
-        call add(args, '%:p')
-    endif
+    let args_list = type(args) == type([])
 
-    call neomake#utils#ExpandArgs(args)
+    if a:maker.file_mode && get(a:maker, 'append_file', 1)
+        if args_list
+            call add(args, '%:p')
+            call neomake#utils#ExpandArgs(args)
+        else
+            let args .= ' '.fnameescape(expand('%:p'))
+        endif
+    endif
 
     if has_key(a:maker, 'cwd')
         let old_wd = getcwd()
@@ -144,14 +148,14 @@ function! s:MakeJob(make_id, maker) abort
     endif
 
     try
-        let has_args = type(args) == type([])
         let error = ''
         if neomake#has_async_support()
-            let argv = [exe]
-            if has_args
-                let argv += args
+            if args_list
+                let argv = [exe] + args
+            else
+                let argv = exe . (len(args) ? ' ' . args : '')
             endif
-            if has('nvim')
+            if has("nvim")
                 let opts = {
                     \ 'on_stdout': function('neomake#MakeHandler'),
                     \ 'on_stderr': function('neomake#MakeHandler'),
@@ -228,14 +232,17 @@ function! s:MakeJob(make_id, maker) abort
             let r = jobinfo.id
         else
             call neomake#utils#DebugMessage('Running synchronously')
-            if has_args
-                if neomake#utils#IsRunningWindows()
-                    let program = exe.' '.join(map(args, 'v:val'))
+            let program = exe
+            if len(args)
+                if args_list
+                    if neomake#utils#IsRunningWindows()
+                        let program = exe.' '.join(map(args, 'v:val'))
+                    else
+                        let program = exe.' '.join(map(args, 'shellescape(v:val)'))
+                    endif
                 else
-                    let program = exe.' '.join(map(args, 'shellescape(v:val)'))
+                    let program .= ' '.args
                 endif
-            else
-                let program = exe
             endif
 
             call neomake#utils#LoudMessage('Starting: ' . program)
```

Tests are passing (`make docker_test DOCKER_VIM=vim-master`), but the Windows case is not covered.

Should I create a new PR based on that, or do you want to pick it up from here?

bounceme:
I am kind of unsure how we should do this.
```
jobstart({cmd}[, {opts}])				{Nvim} *jobstart()*
		Spawns {cmd} as a job.  If {cmd} is a |List| it is run
		directly.  If {cmd} is a |String| it is processed like this: >
		  :call jobstart(split(&shell) + split(&shellcmdflag) + ['{cmd}'])
<		NOTE: read |shell-unquoting| before constructing any lists 
		with 'shell' or 'shellcmdflag' options.  The above call is 
		only written to show the idea, one needs to perform unquoting 
		and do split taking quotes into account.
```
from what i can see all args are being additionally quoted, here is debug with vint:
`Neomake [3.268]: Starting: vint '--style-problem' '--no-color' '-f' '{file_path}:{line_number}:{column_number}:{severity}:{description} ({pol
icy_name})' '/Users/mac/.vimrc'`

I don't know whether or not that is fine with some shell syntax, but the maker using emacs failed when it had those quoted ( and also failed with your patch ).

blueyed:
As for vint this is fine (and worked for me).
Only the argument to `-f` needs to be quoted really, but it's not meant to use a string for `vint` anyway - I was just trying it out.

> the maker using emacs failed when it had those quoted ( and also failed with your patch ).

Can you construct a minimal maker where it fails?  Would be good to use just `echo` or something more portable than emacs for it.
You are not using Windows, are you?

bounceme:
I use osx and bash

maybe this helps?

```
let g:neomake_echo_maker = {
      \ 'exe': 'echo',
      \ 'args': ['in','array','in array'],
      \ 'errorformat': '%f %m',
      \ }
```

```
Neomake [3.470]: [1] Running makers: ['echo']
Neomake [3.472]: [1] Starting async job: ['echo', 'in', 'array', 'in array', '']
Neomake [3.476]: [1.1] stdout: echo: ['in array in array ', '']
Neomake [3.477]: [1.1] exit: echo: 0
Neomake [3.477]: [1.1] echo: processing 1 lines of output.
Neomake [3.478]: Could not place signs for 1 entries without line number.
Neomake [3.479]: [1.1] echo: completed with exit code 0.
Neomake [3.479]: File-level errors cleaned in buffer 1
```

bounceme:
I think I get your patch now. would it be able to expand '%:p' in the string?

blueyed:
@bounceme 
Not sure if I get your question, but the idea is to only expand `%:p` if `args` is a list, otherwise you would have to to it when returning the string.

btw: https://github.com/neomake/neomake/pull/845

bounceme:
```
    if a:maker.file_mode && get(a:maker, 'append_file', 1)
        if args_list
            call add(args, '%:p')
            call neomake#utils#ExpandArgs(args)
        elseif args =~# '%:p'
            let args = substitute(args,'%:p','\=expand(submatch(0))','g')
        else
            let args .= ' '.fnameescape(expand('%:p'))
        endif
    endif

```

blueyed:
This is straight from my patch, right?

I'll create a PR based on that, and we work from there.

blueyed:
In case you have not seen it already, from Vim's `:h job_start`:

> {command} can be a String.  This works best on MS-Windows.  On
> Unix it is split up in white-separated parts to be passed to
> execvp().  Arguments in double quotes can contain white space.
> 
> {command} can be a List, where the first item is the executable
> and further items are the arguments.  All items are converted
> to String.  This works best on Unix.

blueyed:
@bounceme 
Please try https://github.com/neomake/neomake/pull/848.

PR #848, opened 2016-12-14
Allow for maker.args to be a string, skipping quotation/escaping

blueyed:
Fixes https://github.com/neomake/neomake/issues/837.
Closes https://github.com/neomake/neomake/pull/841.

TODO
 - [ ] probably does not work as expected on Windows?!  Could somebody test it?

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -121,12 +121,16 @@ function! s:MakeJob(make_id, maker) abort
     else
         let args = a:maker.args
     endif
-    let append_file = a:maker.file_mode && index(args, '%:p') == -1 && get(a:maker, 'append_file', 1)
-    if append_file
-        call add(args, '%:p')
-    endif
+    let args_is_list = type(args) == type([])
 
-    call neomake#utils#ExpandArgs(args)
+    if a:maker.file_mode && get(a:maker, 'append_file', 1)
+        if args_is_list
+            call neomake#utils#ExpandArgs(args)
+            call add(args, expand('%:p'))
+        else
+            let args .= ' '.fnameescape(expand('%:p'))
+        endif
+    endif
 
     if has_key(a:maker, 'cwd')
         let old_wd = getcwd()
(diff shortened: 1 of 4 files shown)


autoload/neomake.vim
```diff
@@ -121,12 +121,16 @@ function! s:MakeJob(make_id, maker) abort
...
-    if append_file
-        call add(args, '%:p')
-    endif
+    let args_is_list = type(args) == type([])
 
-    call neomake#utils#ExpandArgs(args)
+    if a:maker.file_mode && get(a:maker, 'append_file', 1)
+        if args_is_list
+            call neomake#utils#ExpandArgs(args)
+            call add(args, expand('%:p'))
```
bounceme: what about 
```
       elseif args =~# '%:p'
            let args = substitute(args,'%:p','\=expand(submatch(0))','g')
```

bounceme: in addition

bounceme:
sorry can't test now. I'll be back at a computer in a few hours

autoload/neomake.vim
bounceme: oh right, forgot it can be a funcref

blueyed: Yes, and also with `append_file`, it will get appended still, of course.

bounceme:
```
function s:EF()
  return join([expand('%:p'),'--quick','--batch',
        \"--eval=\"(progn(package-initialize)(require 'js2-mode)(js2-mode)(js2-reparse t)(js2-display-error-list)(princ(with-current-buffer \\\"*js-lint*\\\"(buffer-substring-no-properties(point-min)(point-max))))(terpri))\"" ,
        \ '2>/dev/null'])
endfunction
let g:neomake_javascript_emacs_maker = {
      \ 'args': function('s:EF'),
      \ 'append_file': 0,
      \ 'errorformat': '%f line %l: %m',
      \ }
```
working fine!

bounceme:
and vim-async magically works!

blueyed:
Great.

I will merge this then (soonish).

btw: it looks like you have to do a lot to work around not getting `stderr`, and after all we should probably not consider `stderr` for the `errorformat` processing, but rather use it as an indication that the maker failed.  I am creating an issue for that.

bounceme:
thanks!

PR #858, opened 2016-12-17
Improve init/define for signs and highlights

blueyed:
/cc @davidosomething as per https://github.com/neomake/neomake/pull/497


blueyed:
Where do you get a Vim without `+signs`?  The configure option `--disable-signs` does not seem to be available (anymore?).  Probably would need to be a version where this feature was not available yet at all?!

PR #860, opened 2016-12-17
allow dict function for maker.args

bounceme:
for example (by the way, i can make a pr for this maker if you want):
```
let g:neomake_javascript_emacs_maker = {
      \ 'append_file': 0,
      \ 'errorformat': '%f line %l: %m',
      \ }
function! g:neomake_javascript_emacs_maker.args()
  return join([expand('%:p'),'--quick','--batch',
        \ "--eval=\"(progn(package-initialize)(require 'js2-mode)(setq js2-include-node-externs t js2-include-rhino-externs t js2-include-browser-externs t js2-strict-missing-semi-warning nil)(js2-mode)(js2-reparse t)(js2-display-error-list)(princ(with-current-buffer \\\"*js-lint*\\\"(buffer-substring-no-properties(point-min)(point-max))))(terpri))\"" ,
        \ '2>/dev/null'])
endfunction
```

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -149,7 +149,7 @@ function! s:MakeJob(make_id, options) abort
         let exe = maker.exe
     endif
     if type(maker.args) == type(function('tr'))
-        let args = call(maker.args, [])
+        let args = call(maker.args, [], maker)
     elseif type(maker.args) == type({})
         let args = call(maker.args.fn, [], maker.args)
     else
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -142,7 +142,7 @@ function! s:MakeJob(make_id, options) abort
 
     " Resolve exe/args, which might be a function or dictionary.
     if type(maker.exe) == type(function('tr'))
-        let exe = call(maker.exe, [])
+        let exe = call(maker.exe, [], maker)
     elseif type(maker.exe) == type({})
         let exe = call(maker.exe.fn, [], maker.exe)
     else
--- a/tests/args.vader
+++ b/tests/args.vader
@@ -26,13 +26,27 @@ Execute (neomake#GetMakers):
 
 Execute (dict function given as maker args value):
   let g:makerdict = {
-        \ 'exe': '/usr/bin/printf',
-        \ 'errorformat': '%f: %m',
-        \ 'append_file': 0,
-        \ }
+    \ 'exe': '/usr/bin/printf',
+    \ 'errorformat': '%f: %m',
+    \ 'append_file': 0,
+    \ }
   function! g:makerdict.args()
     return ['%s: %s', 'file1', 'exe: '.self.exe]
   endfunction
   call neomake#Make(1, [makerdict])
   NeomakeTestsWaitForFinishedJobs
   AssertEqual map(copy(getloclist(0)), 'v:val.text'), ['exe: /usr/bin/printf']
+
+Execute (exe function given as maker exe value):
+  function! g:Eref() dict
+    return '/usr/bin/printf'
+  endfunction
+  let g:makerexe = {
+    \ 'exe': function('g:Eref'),
+    \ 'args':['%s %s\n', 'curfile:', 'exe'],
+    \ 'errorformat': '%f: %m',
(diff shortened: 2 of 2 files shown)


autoload/neomake.vim
```diff
@@ -118,7 +118,7 @@ function! s:MakeJob(make_id, maker) abort
         let exe = a:maker.exe
     endif
     if type(a:maker.args) == type(function('tr'))
-        let args = call(a:maker.args, [])
+        let args = call(a:maker.args, [], a:maker)
```
blueyed: Should be done also for `exe` above.

bounceme:
probably obvious, but, what is wrong with the test?

bounceme:
sorry this pr is so messed up haha

bounceme:
???

tests/args.vader
```diff
@@ -23,3 +23,18 @@ Execute (neomake#GetMakers):
...
+        \ 'errorformat': '%f: %m',
+        \ 'append_file': 0,
+        \ 'mapexpr': 'expand("%:p") . ": " . v:val',
+        \ }
+  function g:makerdict.args(...) dict
+    return "'%d\\n' " . (self == g:makerdict)
+  endfunction
+  call neomake#Make(1, [makerdict])
+  NeomakeTestsWaitForFinishedJobs
+  AssertEqual getloclist(0).text, '1'
```
blueyed: >     (25/60) [EXECUTE] (X) Vim(call):E730: using List as a String

Should be `getloclist(0)[0].text`.
Or better: `AssertEqual map(copy(getloclist(0)), 'v:val.text'), ['1']`.

blueyed:
One step closer.. ;) (I've committed from the GitHub UI).
But now it returns/adds `0` instead of `1` (https://travis-ci.org/neomake/neomake/jobs/184754388#L740).

btw: `make test VADER_ARGS=tests/args.vader` locally might work better.. ;)

bounceme:
> btw: make test VADER_ARGS=tests/args.vader locally might work better.. ;)
your right! was just using `make`

I still don't know what is wrong with the `self` comparison though. maybe some other condition instead?

blueyed:
Fixed the test, but unfortunately it passes without your patch also?!
Ah, with an older Vim it fails (without your patch):

> Vim(let):E725: Calling dict function without Dictionary: 1

What Vim/Neovim are you using?

blueyed:
Great, now Neovim on Travis is crashing?!
(I have seen segfaults while playing around with the test, e.g. when trying to use `string(self)`).

bounceme:
I use neovim release 1.7

bounceme:
should i wait till the tests pass before adding the exe support?

blueyed:
The tests won't magically work (usually) ;) - and I give it back to you for now.
If you find usability issues with the tests, please mention them.. :)

bounceme:
I assume vim 73, or vader is causing problems. I was trying and getting syntax errors randomly

bounceme:
Is this a vader issue @junegunn ?

blueyed:
@bounceme 
It comes from here: https://github.com/junegunn/vader.vim/blob/d2c838a62f5ce01c54bd91a5c261600af3cf21db/autoload/vader/parser.vim#L158-L161, and it is caused by using a tab instead of spaces here.
Might be worth reporting and fixing this for Vader maybe (to allow for tabs instead of spaces, or have the error message clearer).

blueyed:
btw: I have some WIP to allow for the `maker` itself have a `fn` function, which then gets called initially (like a constructor), and would allow to change anything dynamically..

bounceme:
you still want to merge this ?

blueyed:
No, the plan is to have a `maker.fn` instead via https://github.com/neomake/neomake/pull/867.
See https://github.com/neomake/neomake/pull/867/commits/973ca80963377417ad93567e1bc9e5505eea8bac.

I also plan to deprecate all the other custom callbacks, but they can stay for now also..

I plan to cherry-pick the tests from here though (if they apply), so I'll leave it open for now.

blueyed:
#867 has been merged finally.

PR #868, opened 2016-12-17
Makefile: highlight lines containing '(X)' in red

blueyed:
It does not seem to be possible to get colors through from Vim itself, so this just wraps `sed` around it to add some highlights.

/cc @junegunn


blueyed:
Meh, this causes `sed: couldn't write X items to stdout: Resource temporarily unavailable` for me for long-running tests..

blueyed:
> sed: couldn't write X items to stdout: Resource temporarily unavailable

`sed --unbuffered` helps here.

PR #870, opened 2016-12-18
C: treat notes as info messages instead of errors

### LINKED ISSUE #396: Parse gcc/clang "note:" as Info, not Error.
More than an Issue is just a question.
I have copied to my own maker, the errorformat from neomake clang maker (#36), it works nice, spotting warnings and errors with different symbols. But the "note: xxxxxx' output is parsed as an error, and I would like other symbol.

It is not clear to me how neomake differentiate warning from errors, do I have to modify errorformat of the maker? Or the postprocess?

tweekmonster:
Fixes #396


tweekmonster:
I wasn't sure if this needs to be applied to `clangtidy` as well.

blueyed:
Thanks!

PR #873, opened 2016-12-18
Add jsonlint-py maker

spk:
ref http://deron.meranda.us/python/demjson/
It is installed as `jsonlint-py` on Debian like to not conflict with PHP `jsonlint`


blueyed:
Merged in 4428d7c, thanks!

blueyed:
Thanks!

PR #794, opened 2016-11-20
Add writegood maker for text, markdown, pandoc

blueyed:
Closes https://github.com/neomake/neomake/pull/773.

TODO:
 - [ ] Needs an unreleased version?!
https://github.com/btford/write-good/commit/a4e52ae33cff729e78e8f502a4ce37e7f7aa9887#commitcomment-19895355
 - [ ] does not have all errors/warning on a single line (https://github.com/btford/write-good/commit/a4e52ae33cff729e78e8f502a4ce37e7f7aa9887#commitcomment-19895374).

PR #874, opened 2016-12-19
Added writegood

bitboxer:
This is adding writegood support for markdown files. Sadly we still have to wait for the next release of writegood. I will keep my eye on this. Just wanted to finish this already so that we can merge this as soon as it is released :wink:


blueyed:
Thanks, but I'd rather re-open https://github.com/neomake/neomake/pull/794 - which adds it for more filetypes etc.

bitboxer:
I can't reopen that because I have no write premission here 😉 . But I can add those filetypes here, too.

blueyed:
Re-opened (and rebased) #794.

Do you think we should merge it already?  It seems like you might be aware of `writegood` issues when using it..

PR #884, opened 2016-12-22
Javascript: Fix flow argument and add support for entry.length

JelteF:
Flow support was totally broken `'--from vim'` did not work, it should either be `'--from', 'vim'` or like I did `'--from=vim'`. This also improves the error message in general and adds support for the length of underline.

### DIFF
--- a/autoload/neomake/makers/ft/javascript.vim
+++ b/autoload/neomake/makers/ft/javascript.vim
@@ -58,13 +58,19 @@ function! neomake#makers#ft#javascript#semistandard() abort
 endfunction
 
 function! neomake#makers#ft#javascript#flow() abort
-    " Replace "\n" by space.
-    let mapexpr = 'substitute(v:val, "\\\\n", " ", "g")'
     return {
-        \ 'args': ['--from vim'],
-        \ 'errorformat': '%E%f:%l:%c\,%n: %m',
-        \ 'mapexpr': mapexpr,
+        \ 'args': ['--from=vim', '--show-all-errors'],
+        \ 'errorformat': '%EFile "%f"\, line %l\, characters %c-%m,%C%m,%Z%m',
+        \ 'postprocess': function('neomake#makers#ft#javascript#FlowProcess')
         \ }
+endfunction
+
+function! neomake#makers#ft#javascript#FlowProcess(entry) abort
+    let l:lines = split(a:entry.text, '\n')
+    if len(l:lines)
+        let a:entry.text = join(l:lines[1:])
+        let a:entry.length = l:lines[0] - a:entry.col + 1
+    endif
 endfunction
 
 function! neomake#makers#ft#javascript#xo() abort


mhinz:
@rafaelrinaldi 

Does this PR work for you, too? I'd like to have a confirmation from another Flow user.

rafaelrinaldi:
@JelteF What Flow version are you using?

JelteF:
I'm using the latest one (0.37.4)

rafaelrinaldi:
@mhinz @JelteF That's weird because it works fine for me and it's actually what Facebook's Vim plugin does too. [See](https://github.com/flowtype/vim-flow/blob/3bd879dd7060f13a78e9238669c2e1731e098607/plugin/flow.vim#L41).

JelteF:
Might be operating system related, that it renders to a  quoted string on
Linux. But the error format was definitely broken. I based mine on the one
from vim-flow, with some tweaks to get the full message.

blueyed:
FB's Vim plugin appears to use a string for the command though: https://github.com/flowtype/vim-flow/blob/3bd879dd7060f13a78e9238669c2e1731e098607/plugin/flow.vim#L50.

@rafaelrinaldi 
So you are using Windows then?
Vim or Neovim?

blueyed:
On Windows and Vim all arguments will be joined into a string anyway: https://github.com/neomake/neomake/blob/master/autoload/neomake.vim#L227-L231.

rafaelrinaldi:
@blueyed I'm on a Mac (macOS Sierra), running Neovim (`v0.2.0-150-g33319b1b`).

blueyed:
@rafaelrinaldi 
Then `'args': ['--from vim']` shouldn't work for you, too?!
Can you show the output of a run with `:let g:neomake_verbose = 3`, in case it works (as in "reports errors correctly")?

blueyed:
@JelteF 
Have you seen https://github.com/neomake/neomake/pull/590?

rafaelrinaldi:
@blueyed But it _does_ work for me.
screen shot 2016-12-22 at 4 29 34 pm

blueyed:
Yeah, the shell is different.

Try `sh -c 'flow "--from vim"'`, where the `--from vim` is passed in as a single argument.
Or `flow '--from vim``.

(It is more important in this context how Neomake handles it (the verbose log output) than your shell or how it looks in screenshots)

blueyed:
I could still imagine than `flow` parses args differently on MacOS that Linux though - e.g. by splitting single args on spaces, but that would be unusual.

JelteF:
@rafaelrinaldi I don't doubt that it works on whatever platform you are running. It just completely fails on mine. I'm quite sure the quoting behaviour for the `args` array is just be different on different platforms. With it being as simple as this:

So for linux:
```
['--from vim']    -> flow '--from vim'
['--from=vim']    -> flow '--from=vim'
['--from', 'vim'] -> flow '--from' 'vim'

```
and on mac
```
['--from vim']    -> flow --from vim
['--from=vim']    -> flow --from=vim
['--from', 'vim'] -> flow --from vim

```

So any of them should work on mac (like you said), but the first one fails on linux because it interprets it interprets it as a string literal. 

@blueyed 
I did not see #590 and used [this](https://github.com/flowtype/vim-flow/blob/3bd879dd7060f13a78e9238669c2e1731e098607/plugin/flow.vim#L39) from vim-flow as a base and changed it a bit so it concatenates all explanation lines (sometimes there are two explanation lines)

@rafaelrinaldi 
Is your (yellow line) output actually from neomake and not from vim-flow. I've never seen an error formatted like that before. That would also explain why it "works" for you, because it's not actually neomake :wink:

PS. I'm using neovim 0.1.7 on arch linux (in xfce4-terminal).

blueyed:
Merged it, thanks!

We can keep on discussing and trying to figure out the behavior @rafaelrinaldi is seeing though still, of course.

rafaelrinaldi:
@JelteF Oh sorry if it sounded that I didn't think this was an issue. I want everybody to be able to use Neomake too, regardless of platform! I just find this to be _interesting_. It might be something silly on the way Flow is parsing stdin args. Will take a look at that and will let you know about my findings here...

Oh and the yellow thing is from my theme configuration. For some reason one day I thought it was a great idea but seeing it now it look ugly as hell.

PR #893, opened 2016-12-24
Add g:neomake_place_signs_immediately

### LINKED ISSUE #834: [Enhancement] New option to show sign column when error lines haven't been displayed yet
On vim with neomake, the sign column is only visible when there have been errors/warnings on one of the displayed lines. For example let's say that I just open a file which has an error on line 95 and only lines 1 to 10 have been displayed so far, then run neomake. What happens is that the sign column won't appear until you navigate and display the error line for the first time.

The same example with syntastic would have shown the sign column from the beginning, without needing to come across the error line. The importance of this issue is that seeing the sign column from the beginning tells you that there is an error/warning somewhere in the document, whereas with the current vim/neomake setup you can't tell.

I don't know if the current behavior is desired, maybe developers are looking forward to display the sign column only when there are errors/warnings on visible lines and hide the sign column otherwise. Maybe it's a consequence of the (great) async mechanism. However I think that if possible, it would be a good idea to provide an option to draw the sign column whenever there is an error/warning line anywhere on the document.

I am using vim 8 but I haven't tested this on neovim.

rhelmot:
Resolves #834 

This works by, at the end of PlaceVisibleSigns, if we notice that there are signs in the queue but none in the current buffer, pluck the first one in the file out of the queue and make it visible. This should make it fast on large files with a huge number of errors. The option is enabled by default.

This is something that has annoyed me for a long time, until today I got fed up and found that issue, read through the surrounding discussions, and learned enough vimscript to do it myself :)

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -127,12 +127,21 @@ function! neomake#signs#PlaceVisibleSigns() abort
         endif
         let topline = line('w0')
         let botline = line('w$')
+        let placed_one = 0
         for ln in range(topline, botline)
             if has_key(s:sign_queue[type][buf], ln)
                 call neomake#signs#PlaceSign(s:sign_queue[type][buf][ln], type)
                 unlet s:sign_queue[type][buf][ln]
+                let placed_one = 1
             endif
         endfor
+        " if this buffer has signs pending but none have been placed, force
+        " the placement of at least one to show the sidebar
+        if !placed_one && !has_key(s:placed_signs[type], buf)
+            let first_ln = keys(s:sign_queue[type][buf])[0]
+            call neomake#signs#PlaceSign(s:sign_queue[type][buf][first_ln], type)
+            unlet s:sign_queue[type][buf][first_ln]
+        endif
         if empty(s:sign_queue[type][buf])
             unlet s:sign_queue[type][buf]
         endif


blueyed:
Thanks!

But wouldn't https://github.com/neomake/neomake/pull/471 fix it also already?

autoload/neomake/signs.vim
```diff
@@ -136,6 +136,15 @@ function! neomake#signs#PlaceVisibleSigns() abort
         if empty(s:sign_queue[type][buf])
             unlet s:sign_queue[type][buf]
         endif
+        if g:neomake_place_signs_immediately
```
blueyed: It also could be skipped in case another one was displayed above already, right?

doc/neomake.txt
```diff
@@ -334,6 +334,11 @@ This setting enables the placement of signs next to items from the location
 and quickfix list (i.e. errors/warnings etc recognized from the
 |'errorformat'|). Defaults to 1.
 
+*g:neomake_place_signs_immediately*
+This setting causes the sign bar to appear immediately when there are
+errors/warnings/etc to display, but none of them are visible on the current
+screen.
```
blueyed: I think there should not be a new setting for this.

rhelmot:
Force-pushed to clean up the commit in response to the suggestions.

You're right, #471 does accomplish the same thing, but it appears to be abandoned or at least stagnated and I would like this to be fixed now :)

rhelmot:
Also, could someone help me understand what the CI failures are supposed to indicate? I can't find anything in the build log telling me what's wrong, and I really don't think this should have broken anything...

blueyed:
Well, I'd rather like to get #471 on track, instead of adding more complexity.
If you want to get this in now, I'd suggest to adopt #471 and submit it as a new PR.

I could investigate with the build failures, but it's not worth it currently probably.

blueyed:
Closing as per previous comment.
I hope you agree!

Issue #896, opened 2016-12-27
Stopped working with PHP


still-dreaming-1:
When I open a .php file, type random stuff in with missing semicolons, and run `:Neomake`, nothing happens. It used to work, I'm not sure what changed other than I updated to the latest Neovim. I tried using Neomake with JavaScript, and that is working fine. I reproduced this with a minimum Neovim config file, so it is not interference from another plugin or anything in my config. I also tried updating Neomake, so I should be on the latest code. I tried rebooting my system as well.

### Expected behavior

The left side of the screen should visually indicate which lines have lint errors, and the bottom of the screen should show the error message when the cursor is on that line.

#### Neomake debug information
Async support: 1
Current filetype: php

##### Enabled makers
For the current filetype (with :Neomake):
 - php
   postprocess: function('neomake#makers#ft#php#PhpEntryProcess')
   errorformat: '%m\ in\ %f\ on\ line\ %l,%-GErrors\ parsing\ %f,%-G'
   args: ['-l']
   exe: 'php'
NOTE: you can define g:neomake_php_enabled_makers to configure it (or b:neomake_php_enabled_makers).

For the project (with :Neomake!): None.
NOTE: you can define g:neomake_enabled_makers to configure it.

Default maker settings:
 - remove_invalid_entries: 0
 - buffer_output: 1

##### Settings
```
g:neomake_place_signs = 1
g:neomake_verbose = 3

shell: /bin/zsh
shellcmdflag: -c
Windows: 0
```

#### :version
```
NVIM 0.2.0-dev
Build type: RelWithDebInfo
Compilation: /usr/bin/x86_64-linux-gnu-gcc -g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -DDIS
ABLE_LOG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector --param ssp-buffer-size=4 -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -D_GN
U_SOURCE -I/build/neovim-YeT38K/neovim-0.1.5ubuntu1+git201612261631+2984+22~ubuntu14.04.1/build/config -I/build/neovim-YeT38K/neovim-0.1.5ubuntu1+git201612261631+2984+22~ubuntu14.04.1/src -I
[...]

PR #900, opened 2016-12-27
Update makers/ft/php.vim from Syntastic

blueyed:
Fixes https://github.com/neomake/neomake/issues/896

### DIFF
--- a/autoload/neomake/makers/ft/php.vim
+++ b/autoload/neomake/makers/ft/php.vim
@@ -1,27 +1,34 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#ft#php#EnabledMakers()
+function! neomake#makers#ft#php#EnabledMakers() abort
     return ['php', 'phpmd', 'phpcs']
 endfunction
 
-function! neomake#makers#ft#php#php()
+function! neomake#makers#ft#php#php() abort
     return {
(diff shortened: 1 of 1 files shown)


still-dreaming-1:
I want to try this out, but I don't quite see.. where it is. Doesn't a pull request normally come from a forked repository?

mhinz:
@still-dreaming-1 They can come from any branch. Also from the same repo. At the bottom of each PR page there's a small message like this one:

```
Add more commits by pushing to the fix-php branch on neomake/neomake
```

So.. it's here! 

```
$ git fetch origin refs/pull/900/head && git checkout FETCH_HEAD 
```

blueyed:
Yep.. and for the full documentation/help about this see https://help.github.com/articles/checking-out-pull-requests-locally/.

still-dreaming-1:
Ah, I got it now I. I was confused because I did actually look for that branch, but I just didn't see if for some reason. Anyway, this works great, thanks! I find it interesting that telling php to show the errors fixed my issue. I did try running `php -l` in the terminal and got the same generic error. I wonder what changed as I did not have this problem until now?

blueyed:
> I wonder what changed as I did not have this problem until now?

A PHP update with new defaults?!
Some config change of your own?

PR #905, opened 2017-01-03
[pylint] fix column off-by-one

novas0x2a:
pylint uses 0-indexed columns, vim uses 1-indexed columns

This bug might exist in flake8, too, but i'm not sure (I don't use it).


blueyed:
Thanks!

PR #688, opened 2016-10-07
Fix tcd changing dir

hkupty:
This way the `tcd` won't be overriden to a `cd`, messing with tab-local directory settings.

PR #867, opened 2016-12-17
[RFC] Fix serializing/chaining

blueyed:
Ref: https://github.com/neomake/neomake/issues/577.

TODO:
 - [ ] doc: maker/jobinfo: https://github.com/neomake/neomake/issues/888
 - [ ] hook instead of .fn?! https://github.com/tweekmonster/deoplete-clang2/commit/872ff514168a719d33b8e588bc64962fa75100dd#commitcomment-20282587

PR #906, opened 2017-01-04
[RFC] more powerful filename functionality

novas0x2a:
This is an RFC for some new functionality; I haven't documented it or written unit tests for it yet, pending approval from upstream :)

This is one approach to solving the problems I list below. Another solution is actually just to set args to a function, so this may be overly complex, but I figured I'd send it off anyway.. Copying the comment from the commit:

This adds a new maker key, `make_filename`, which is expected to
return the filename to append (when append_file is set).

This approach allows two problems to be solved:

1) Running pylint on a single file is useful, but weak. Running pylint
   on a whole project (using a project maker) can be quite slow. A
   good middle ground is to run pylint on the module; in this case,
   make_filename can return something like expand('%:p:h'). This also
   applies to makers for other languages with strong
   module-as-directory semantics (such as go).

2) The current go makers rely on cwd changes, which can cause weird
   problems (#688) particularly if two linters are in use with
   different cwds.

   make_filename allows for a different approach- if make_filename is
   `return './' . expand('%:.:h')`, no cwd changes are necessary.

   Note: this in particular is why make_filename takes a function and
   not an expand() directive; the go tool requires the ./ prefix on
   the directory name or it searches the GOPATH for the argument,
   rather than interpreting it as a directory.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -127,6 +127,10 @@ function! s:AddJobinfoForCurrentWin(job_id) abort
     call settabwinvar(tabpagenr, winnr, 'neomake_jobs', win_jobs)
 endfunction
 
+function! s:DefaultFilename() abort
+    return expand('%:p')
+endfunction
+
 function! s:MakeJob(make_id, options) abort
     let job_id = s:job_id
     let s:job_id += 1
@@ -158,11 +162,12 @@ function! s:MakeJob(make_id, options) abort
     let args_is_list = type(args) == type([])
 
     if a:options.file_mode && get(maker, 'append_file', 1)
+        let MakeFilename = get(maker, 'make_filename', function('s:DefaultFilename'))
         if args_is_list
             call neomake#utils#ExpandArgs(args)
-            call add(args, expand('%:p'))
+            call add(args, MakeFilename())
         else
-            let args .= ' '.fnameescape(expand('%:p'))
+            let args .= ' '.fnameescape(MakeFilename())
         endif
     endif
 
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks!
But instead having more options, what about using `maker.fn` (in https://github.com/neomake/neomake/pull/867) for this?
You could set `append_file` explicitly and append the art itself (similar to the command maker).

As for 2. I think it should not require a workaround, but should get fixed instead.. IIRC #867 might do so already.
However, skipping the `cd` back and forth is good in general!

novas0x2a:
My vimscript is a little weak, but let me see if understand. Right now (against master) to configure pylint I'm doing:
```viml
let g:neomake_python_enabled_makers = ['pylint']
let g:neomake_pylint_append_file = 0
let g:neomake_pylint_args = {}
function g:neomake_pylint_args.fn()
    return neomake#makers#ft#python#pylint()['args'] + [GetPylintRCArgs()] + [expand('%:p:h')]
endfunction
```

If I understand #867 correctly, the way I'd do that once it lands is
```viml
let g:neomake_python_enabled_makers = ['pylint']
let s:pylint_maker = neomake#GetMaker('pylint', 'python')
function s:pylint_maker.fn() dict
    let self.args = extend(self.args, [GetPylintRCArgs(), expand('%:p:h')])
    let self.append_file = 0
    return self
endfunction
```
If my understanding is correct, yeah, I'd be just fine with #867 instead of this pr.

blueyed:
Since https://github.com/neomake/neomake/pull/867 has been merged, I'm closing this.
Thanks!

Issue #910, opened 2017-01-05
Scss-lint works but Neovim doesn't show hints


caneta:
I cannot make scss-lint display error messages on my Neovim interface.

### Expected behavior

I expect to have some sort of hint when saving a scss with errors, or when I run `:Neomake`

### Steps to reproduce
Go into a directory with an scss file: running `scss-lint mybadfile.scss` works ad reports errors.
If I open neovim and run Neaomake, nothing happens.
Looking at `:messages` it seems working, but nothing is displayed on Neovim interface.

### Output of the ":verb NeomakeInfo" command
```

#### Neomake debug information
Async support: 1
Current filetype: scss


##### Enabled makers
For the current filetype (with :Neomake):
 - scsslint
   errorformat: '%A%f:%l:%v [%t] %m'
   exe: 'scss-lint'
   args: []
NOTE: you can define g:neomake_scss_enabled_makers to configure it (or b:neomake_scss_enabled_makers).


For the project (with :Neomake!): None.
NOTE: you can define g:neomake_enabled_makers to configure it.


Default maker settings:
 - remove_invalid_entries: 0
 - buffer_output: 1


##### Settings
g:neomake_verbose = 3
g:neomake_error_sign = {'text': '❌'}
g:neomake_warning_sign = {'text': '⚠️'}
g:neomake_java_javac_maker = {'args': ['cp', '~/Scrivania/liferay-ce-portal-src-7.0-ga2']}
g:neomake_style_warning_sign = {'text': '💩'}
g:neomake_place_signs = 1
g:neomake_style_error_sign = {'text': '⁉️'}


shell: /bin/bash
shellcmdflag: -c
Windows: 0


#### :version

NVIM 0.2.0-dev
Build type: RelWithDebInfo
[...]

PR #520, opened 2016-07-25
Update the SCSS linter errorformat

icatalina:
Seems like the scss-lint format has changed, this updated rule should
cope with the new schema.

PR #911, opened 2017-01-05
scss: scsslint: errorformat: handle missing column

blueyed:
Column information was added in scss-lint 0.49.0.

Fixes https://github.com/neomake/neomake/issues/910.


blueyed:
For reference: the errorformat was changed in https://github.com/neomake/neomake/pull/520.

PR #924, opened 2017-01-12
[WIP] Custom formatted location list

tweekmonster:
Replaces the location list buffer with better aligned content:

- Filename is removed and displayed in `w:quickfix_title`
- Adds each item's sign to the sign column
- Adds a fake gutter that displays the maker name, line number, and
  column

This is similar to the example I posted in #325

Currently, there's a couple things that need to be addressed:

- Documentation.
- This is enabled through `neomake#quickfix#enable()`.  The idea is that there would also be a `disable()` method.  It probably should be enabled/disabled with a variable.
- There can be list items that point to a different buffer.  I think those items should just be prefixed with something like `[buf:shortname]` instead of adding a column for all items.  However, I only saw this happening with `govet` which looks like it was using a temporary buffer.
- The quickfix window is not reformatted.  It might be nice to format the quickfix list so that the columns line up nicely and displays the makers.
- If the location list is already visible, neomake's `s:current_errors` isn't populated for the current buffer when this script runs.  I haven't figured out why that is.

The main thing blocking this is I can't make this only run after Neomake finishes.  Something I thought about was possibly adding a dummy entry at the end of the list that isn't added to the buffer.  That way it can be distinguished from location lists created by other scripts.

## Preview

Before:
screen-shot-2017-01-12-17-32-06

After:
screen-shot-2017-01-12-16-43-59

Before:
screen-shot-2017-01-12-17-33-58

After:
screen-shot-2017-01-12-16-41-39

Before:
screen-shot-2017-01-12-17-35-41

After:
screen-shot-2017-01-12-16-40-06


autoload/neomake.vim
```diff
@@ -42,6 +42,13 @@ function! neomake#GetJob(job_id) abort
     return s:jobs[a:job_id]
 endfunction
 
+function! neomake#GetErrors(bufnr) abort
+    if !a:bufnr
+        return s:current_errors['project']
+    endif
+    return get(s:current_errors['file'], a:bufnr, {})
+endfunction
+
```
blueyed: The (public) API here should be better.. 
See https://github.com/vim-airline/vim-airline/pull/1275#issuecomment-271122146 for some discussion / link.

As for this PR: it is only used once, so keep it inlined for now?!

autoload/neomake/quickfix.vim
```diff
@@ -0,0 +1,148 @@
...
+        setlocal breakindent linebreak
+        let &breakindentopt = 'shift:'.(b:neomake_start_col + 1)
+    endif
+
+    for item in signs
+        call neomake#signs#PlaceSign(item, 'file')
+    endfor
+
+    runtime! syntax/neomake/qf.vim
+    execute 'runtime! syntax/neomake/'.getbufvar(src_buf, '&filetype').'.vim'
```
blueyed: `getbufvar(src_buf, '&filetype')` might be empty?!

syntax/neomake/qf.vim
```diff
@@ -0,0 +1,5 @@
+syntax clear
```
blueyed: Doesn't this clear other custom configs?
Is it required for resetting/reloading?

blueyed:
> The quickfix window is not reformatted. It might be nice to format the quickfix list so that the columns line up nicely and displays the makers.

Yes, also possibly applying `pathshorten()` and some other things in the end.

> If the location list is already visible, neomake's s:current_errors isn't populated for the current buffer when this script runs. I haven't figured out why that is.

Likely a bug then.  Might depend on the setting to open the quickfix/location list?!

blueyed:
> The main thing blocking this is I can't make this only run after Neomake finishes. Something I thought about was possibly adding a dummy entry at the end of the list that isn't added to the buffer. That way it can be distinguished from location lists created by other scripts.

What about using the (badly named) NeomakeCountsChanged hook, either directly or by also looking for `qf` FileType events and then act accordingly?

autoload/neomake.vim
tweekmonster: I'm not sure what you mean.  Are you saying that you want it to be more flexible and compatible with the goal of that airline issue?

blueyed:
I think I'd prefer the line/col first (before the maker): It is more compact in length (for the column width, in case of multiple makers) there, and also the maker's name is a good prefix for the error already.

In https://github.com/neomake/neomake/pull/325 my idea was to append it at the end ` [maker]`: it is useful to see quickly, but not really a category most of the time.

You might have put it to the beginning for some reason though?! :)

syntax/neomake/qf.vim
tweekmonster: I suppose it would.  But, as stated above, this would ideally run *only* for Neomake.  For Neomake-specific syntax, custom configs would live in `&rtp/syntax/neomake/qf.vim`.

The reason I did it like this is so that it wouldn't clash with any existing `syntax/qf.vim` configs, as well as allow maker-specific highlighting like in the example for Python.

autoload/neomake.vim
blueyed: Yes, and the statusline API in general.
My plan is to clean up my local config some day for that..

syntax/neomake/qf.vim
blueyed: I like the approach of using `:runtime` here.

tweekmonster:
> Likely a bug then. Might depend on the setting to open the quickfix/location list?!

My theory is that during the async post-processing, the buffer focus is changing temporarily due to the list changing.  So, this script is running beforehand.

I forgot to mention that this only seemed to happen with the Python maker.   Note to self: Now that I'm thinking about it, I'm using `impsort.vim` that does async highlighting and it might changing the buffer focus unexpectedly.

> What about using the (badly named) NeomakeCountsChanged hook, either directly or by also looking for qf FileType events and then act accordingly?

As you mentioned, it would require also looking out for `Filetype qf`.  The combination is too magical for my liking.  Future changes to how the hooks are handled could break this too easily.

I'm actually starting to like my idea of adding a dummy entry to the list.  If this feature is enabled, the last entry would have an item with `pattern` set to `{neomake}` as the marker/signal, and the text could be a comma delimited list of the makers, or possibly even a string representation of a dict that contains extra information that can't be added to the `setloclist()` function.  This script would simply not add a line for it.

> I think I'd prefer the line/col first (before the maker): It is more compact in length (for the column width, in case of multiple makers) there, and also the maker's name is a good prefix for the error already.

I actually want the fake gutter to be customizable.  I didn't go through with doing all that because I wanted to make sure this PR was acceptable in general first :P

An idea to make it more compact is to just truncate the maker name to 3-4 characters.  Then an optional `maker_name_short` key for the maker configs could be used if manual shortening is needed to make it less ambiguous.

> In #325 my idea was to append it at the end [maker]: it is useful to see quickly, but not really a category most of the time.

I thought about this, but, it didn't seem right that if you wanted to know what maker emitted the message, you had to actually look for it instead of just knowing where to look in a designated space.  That would be annoying if the message is long and off screen, or if the surrounding text is dense enough to make it hard to see.

> You might have put it to the beginning for some reason though?! :)

I was just keeping the order as `signs | fold | number`, where `fold` is the maker name.  No real thought went into the ordering :D

tweekmonster:
I'm feeling the pressure to finish this now 😜   I'll work on this a bit more today.

blueyed:
Any updates?
Please rebase it at least to make it easier to dogfood.

tweekmonster:
This now supports quickfix as well!

I went ahead with the dummy quickfix entry since it seemed like the most reliable way to pass along the maker names.  It resolved the issue I mentioned above with the python maker.

I haven't done the gutter customization yet because I couldn't think of a good way to do this without a lot of explanation.  I was thinking about using python-style formatting:

```
%(maker)s %(lnum)s:%(col)s
```

Then I would get the order by removing the `(word)`s from the string and replacing them with `*` so that it can be used in the `printf()`.

blueyed:
> This now supports quickfix as well!

Awesome!

Quick feedback/issue: probably because of the dummy entry in the end, my automatic resizing of the qf list is off by one now.. i.e. normally it would be resized for one entry, but now by the time of the resizing there are likely 2.  I am doing that mainly on `FileType qf`, but have not investigated yet.

tweekmonster:
Yeah that sounds like the dummy entry.  I like the dummy entry because it's a foolproof way to pass info around, but I'm not committed to the idea.  The alternative is maintaining that info in a script var and querying for it later.  The problem is that there wouldn't be an easy way to know if `getqflist()` is returning a list that's from Neomake.

We could just use the `pattern` key for each entry to hold the maker name (e.g. `neomake:flake8`), but I'm not exactly clear on what that field does and if using it will cause other issues.

tweekmonster:
I made it so that the maker name is attached to the list entry's text, but only if the custom qf is enabled.

I wanted to use the `pattern` field since that wouldn't be useful for Neomake, but it turns out that `pattern` overrides `lnum`.  I was going to try using something like `\%15l\%(maker\)\?` for the pattern field, but it seemed a little overboard to make Vim parse regex to jump to a line.

tweekmonster:
@blueyed Requesting a review.  I think this is working pretty well where it is, but I wanted to get your thoughts about attaching the maker info to the text.

autoload/neomake.vim
blueyed: Can we move this to some private API for now?
It also does not seem to be used (anymore)?!

autoload/neomake.vim
```diff
@@ -658,10 +665,12 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
     let ignored_signs = 0
     let s:postprocess = get(maker, 'postprocess', function('neomake#utils#CompressWhitespace'))
     let debug = get(g:, 'neomake_verbose', 0) >= 3
+    let custom_qf = get(g:, '_neomake_qf_enabled', 0)
```
blueyed: This should be a `s:` var maybe?!
But since it's used in `autoload/neomake/quickfix.vim` it could become a function/property there instead.

autoload/neomake/quickfix.vim
```diff
@@ -0,0 +1,192 @@
+" vim: ts=4 sw=4 et
+scriptencoding utf-8
+
+let s:maker_match_id = 997
+let s:gutter_match_id = 998
+let s:cursor_match_id = 999
```
blueyed: Could those numbers conflict with other plugins etc?

autoload/neomake/quickfix.vim
```diff
@@ -0,0 +1,192 @@
...
+                    \ s:cursor_match_id,
+                    \ s:cursor_match_id)
+    endif
+endfunction
+
+
+function! s:reset(buf) abort
+    call neomake#signs#ResetFile(a:buf)
+    execute 'sign unplace * buffer='.a:buf
+    call clearmatches()
```
blueyed: Might be nice to only clear matches that were defined by ourselves?!

autoload/neomake/quickfix.vim
```diff
@@ -0,0 +1,192 @@
...
+        let b:neomake_qf = 'file'
+        let src_buf = qflist[0].bufnr
+    else
+        let b:neomake_qf = 'project'
+    endif
+
+    runtime! syntax/neomake/qf.vim
+    if src_buf
+        let ft = getbufvar(src_buf, '&filetype')
+        if ft != ''
```
blueyed: Use `!=#` to make `vint` happy.
Or just `len(ft)`.

autoload/neomake/quickfix.vim
```diff
@@ -0,0 +1,192 @@
...
+            for name in split(ft, '\.')
```
blueyed: There are methods for splitting / handling multiple fts already, which _could_ be used.. looks OK for me as-is though.

autoload/neomake.vim
tweekmonster: It's actually unused now.  Should I just remove it?

autoload/neomake.vim
tweekmonster: `neomake#quickfix#is_enabled()`?

autoload/neomake/quickfix.vim
tweekmonster: Not sure.  Is there a base ID I should use?

autoload/neomake/quickfix.vim
tweekmonster: or `empty()`

autoload/neomake/quickfix.vim
tweekmonster: Oh in this plugin?  I'll double check!

autoload/neomake/quickfix.vim
blueyed: Not sure either.. let's keep it like this then for now.
For signs base ids of 5000 and 7000 are used.

autoload/neomake.vim
tweekmonster: Actually, would it be more consistent to use`neomake#quickfix#IsEnabled()`?

autoload/neomake/quickfix.vim
blueyed: See https://github.com/neomake/neomake/blob/2ad3768f176fd5537831107e989c4ae6d63974e9/autoload/neomake/utils.vim#L230-L242 - but this might be overkill/wrong after all.

autoload/neomake.vim
blueyed: In some sense, yes.  Although I prefer the former style, and we could move slowly towards it instead.

tweekmonster: Sounds good to me.  Given the scope of this plugin, it could also be defined as `neomake#TitleCase` functions are documented, while `neomake#snake_case` is intended to be undocumented/subject to change.

blueyed: Agreed.

autoload/neomake/quickfix.vim
tweekmonster: Ahh, hmm.  Now, I'm not sure if this is right at all.  Should the syntaxes be based on the maker names instead of `&filetype`?

blueyed: Yeah, it should be a property of the maker(s): all makers contributed to the list should be able to define the syntax.  But then the defaults (like here) could come from the filetype / `supertype`.
(could be done for settings in general, but then settings are maker-specific in general)

And the whole setup for syntax should be move to a function for easier testing.

tweekmonster: Alright so, that means the maker names should be truncated later.  Otherwise, there's no way to get that information.

If it's alright, I'm going to stick with just using the maker names for now and deal with this if issues arise from it.

blueyed: How so?
When using a callback/setting on the maker, it would scale with settings we have already.
With maker names alone, those are not unique, and could be used for different filetypes (different linters).

tweekmonster: I think I got lost in the conversation.  What I'm saying above is that I have to attach the maker name to `entry.text` as `{neomake:maker_name}` since getting that info through a function is unreliable (particularly for quickfix).  But, currently it's the short form of the name that's attached to the text since I made it an optional maker property (`short_name`).

I also think we can worry about syntax for maker supersets a little later since this is a completely new feature.

blueyed: OK.

tweekmonster:
@blueyed I'll fix the lint error.  Last detail for this is to add tests.  Do you have any suggestions for this?

blueyed:
> Last detail for this is to add tests. Do you have any suggestions for this?

Not really.
Probably a new .vader file makes sense for this feature (include it in neomake.vader then).
`integration.vader` has the most diverse set of tests, where you might find a good base one.
Also rebasing makes sense, since tests have been improved a bit.

I recommend also to use my Vader fork for better tracebacks/error messages (https://github.com/neomake/neomake/pull/974).

tweekmonster:
Well I was wondering if Vader will have an issue with opening the quickfix window to get the contents.  I also just noticed that the maker config in the entry text is leaking into the text displayed in the message area.  Need to fix that.

blueyed:
> Well I was wondering if Vader will have an issue with opening the quickfix window to get the contents.

Should be good.
Both as in opening it really (`:copen`) or accessing it through `cgetlist()`.

blueyed:
You can run `make tests/foo.vader` to only run a particular file - that's what I do with a set of new tests usually.

And for the better Vader: you can symlink to it in `tests/vim/plugins/vader.override`.

blueyed:
Rebased on master.

tweekmonster:
Thanks!  I haven't had a a ton of time to work on the tests.  One day 😬

tweekmonster:
@blueyed I had some changes that I didn't push and a forced update was easier than sorting out the rebase.

tweekmonster:
@blueyed Sorry I sat on this for so long.  I added a couple basic tests.  It needs to improve, but I think it can be dealt with later since the tests at least ensure it doesn't disrupt the normal quickfix display.

Notes:

- I couldn't test the fake gutter's syntax highlighting since that doesn't seem to be possible.
- Found it strange that `normal!` didn't trigger `CursorMoved`.
- Needs tests for multiple makers.
- Needs tests for checking truncated maker names.

tweekmonster:
I guess I need to fix up the test.  Forgot to include it in `neomake.vader`.

tweekmonster:
I'll let you take a look at the tests.  Everything I try ends up not working or breaking other tests.  I keep getting this:

```
[EXECUTE] (X) Vim(let):E121: Undefined variable: g:neomake_test_postprocess_count
```

tweekmonster:
@blueyed I simplified the tests so that only the rendering is checked.  I couldn't get the tests to pass as mentioned above.

Looks like I still need to make some fixes for older versions of Vim, though.

tests/customqf.vader
```diff
@@ -0,0 +1,92 @@
...
+      \ }])
+
+  lopen
+
+  AssertEqual getline(1, '$'),
+    \ ['Alph 1:- Line 1 ', 'Alph 2:5 Line 2']
+
+  normal! 0
+  " :normal! ignores autocmd's here?
+  doautocmd CursorMoved
```
blueyed: Not supported by Vader apparently: https://github.com/junegunn/vader.vim#some-events-may-not-be-triggered - doing it manually is the way to go.

blueyed:
It's good to have the tests simplified, although some integration test would be useful, too.
The previous ones looked to verbose though.
It should be much easier with https://github.com/neomake/neomake/pull/1050 - I'll wait until that is in.

Seems to be only one issue with vim73:
> (112/150) [EXECUTE] (X) Vim(sign):E474: Invalid argument
      > function neomake#quickfix#FormatQuickfix..<SNR>47_reset, line 2

tweekmonster:
> Seems to be only one issue with vim73:

Also one issue with 7.4 because of `matchaddpos()`.  It's a problem for 7.3, too.  It's concealed by the sign error.

tweekmonster:
Whats the best way to clear signs from the quickfix buffer?  I'm using `sign unplace * buffer=` because `neomake#signs#ResetFile()` seemed to always leave a sign behind on line 1.

autoload/neomake/quickfix.vim
```diff
@@ -33,10 +33,18 @@ function! s:cursor_moved() abort
...
+                        \ [[line('.'), (b:neomake_start_col - b:neomake_number_len) + 2, b:neomake_number_len]],
+                        \ s:cursor_match_id,
+                        \ s:cursor_match_id)
+        else
+            call matchadd('neomakeCursorListNr',
+                        \  '\%' . line('.') . 'c'
+                        \. '\%' . ((b:neomake_start_col - b:neomake_number_len) + 2) . 'c'
+                        \. '.\{' . b:neomake_number_len . '}',
+                        \ s:cursor_match_id, s:cursor_match_id)
+        endif
```
blueyed: It's used `autoload/neomake/highlights.vim` also, so there could be a wrapper.
OTOH I've noticed that using `matchadd` (with `\%Xv` (not `c` like above)) seems to be the sanest currently, given that other methods use byte offsets.  See https://github.com/neovim/neovim/issues/6270#issuecomment-286403626.
So a wrapper would make it easier to handle this in the future, but it's fine to keep this here for now.
However, I think `v` with `matchadd` might be the best regarding multibyte strings.
If keeping `matchaddpos`, then `nvim_buf_add_highlight` might be an option for Neovim, too.

blueyed:
> Whats the best way to clear signs from the quickfix buffer?

The signs interface is weird, too (and I've local changes for it somewhere already).. :/
Which sign is left behind?  `neomake#signs#ResetFile()` gets used to indicate that signs are being replaced IIRC.  You then have to call CleanOldSigns (or similar).  IIRC some tests use it.

That should not really affect the qf buffer though, does it?
Why do you care about clearing (and placing) signs here?

tweekmonster:
Aww man, I actually wrote out a response for this but I must've closed the window before posting it.

I'm adding signs to the qf buffer opposed to using `E`, `W`, etc.  I also didn't want to add the sign characters in the buffer.  Would've required editing `signs.vim` to get the configured signs.  But, I mainly didn't want to deal with the highlighting for it.

> You then have to call CleanOldSigns (or similar). IIRC some tests use it.

Thanks I'll look into this later.

blueyed:
Is this using a dummy entry still, or rather `b:neomake_qf`?
I am wondering since I would like to make Neomake (optionally?) not change any existing location (and quickfix?) list that is not owned by it, i.e. only if there is no location list yet, or it is owned by Neomake (see https://github.com/neomake/neomake/issues/1165#issuecomment-290394265).

blueyed:
I've rebased it against master locally, let me know if I should push it (about 3 minor conflicts, so you might want to do it yourself).

blueyed:
`:sign unplace * buffer={nr}` does not work in the Vim 7.3 build; you might have to remember the placed signs, and unplace them one by one there.
Like mentioned before, I have some improvements to the signs module, and it might be usable for this already in the current state to register and clear the signs in the qf window.

blueyed:
btw: Vim just added support for `context` to location/quickfix lists: https://github.com/vim/vim/commit/8f77c5a4ec756f3f866bd6b18feb6fca6f2a2e91.

petobens:
Any updates on this? It would be really nice to have :)

blueyed:
@tweekmonster seems to be in (another) zone currently (which is fine :)).

Since this is the oldest open PR currently, and I also like this feature very much myself, I plan to pick it up  myself eventually (but there at least 3 other big things to get in first).

@petobens 
Have you tried it?  (should be rebased on master though to get other improvements)
Do you feel like contributing and picking it up?

tweekmonster:
@petobens @blueyed I haven't been able to keep up due to personal and professional priorities.  I'd like to pick this back up, but I'm not sure when I'd be able to.  I really don't mind if someone would like to continue the work.

petobens:
@blueyed @tweekmonster I would love to pick this up but unfortunately my vimscript skils are not good enough to do so...

blueyed:
Merged master, and adjusted it to handle makers with get_list_entries/process_output.

syntax/neomake/qf.vim
blueyed: The `let b:current_syntax = 'neomake_qf'` gets cleared with `syntax clear` though, so what is the point of setting it?
Should it get moved after `syntax clear`, or is it required for `syntax clear` somehow?

blueyed:
FWIW: the "next" branch contains both "automake" and this, in case you want to try it together with "automake".. ;)

blueyed:
It needs to be profiled/optimized probably, since it appears to add a lot of sluggishness (when used often / with automaking).

blueyed:
This is next / coming soon(™) - the last time I've tried it it had a decent set of conflicts already - looking forward to solve them.
IIRC performance is an issue here also - especially since #1167 has been merged by now.

blueyed:
Yay for diff coverage reports: https://codecov.io/gh/neomake/neomake/pull/924/diff

blueyed:
I would like to get this into master soon.

The only remaining question might be how to configure it.

It might be good after all to explictly opt in via `neomake#quickfix#enable()`, and then we can enhance / integrate it from there?!

blueyed:
I would like to get this into the `neomake#configure` namespace eventually - similar to how it is used with automake, but it needs a bit of refactoring then.

blueyed:
:fireworks: 

Thanks a lot, @tweekmonster!

blueyed:
Follow up in https://github.com/neomake/neomake/issues/1698.

petobens:
Hi @blueyed. I want to try this new feature but I can't seem to get it working. I have the following settings in my vimrc:
```vim
call neomake#quickfix#is_enabled()  " enable experimental quickfix formatting
let g:neomake_python_enabled_makers = ['flake8', 'mypy']
let g:neomake_python_flake8_args = ['--ignore=E402,W503']
let g:neomake_python_mypy_args = ['--ignore-missing-imports',
            \ '--follow-imports=skip', '--python-version', '3.6']
```
However when I run neomake on the following file:
```python
def f(x: int) -> int:
    return x
print(f([5]))
```
I see only see the following line in the quickfix:
```
bar.py|5 error| Argument 1 to "f" has incompatible type "List[int]"; expected "int"
```

I should say `mypy` somewhere right? (at least that's what i gathered from the original screenshots). 

Btw if i add a long line (>80 columns) at the . top like this:
```python
# asdlksadjasldk jaskld asdjas kldasjdkl asdjasld kjsadkl saj dklasdja skld jasklas
def f(x: int) -> int:
    return x
print(f([5]))
```
and run neomake again then I only see
```
bar.py|1 col 80 warning| E501 line too long (83 > 79 characters)
```
which is the flake8 warning (i expect to see both flake8 and mypy).

Thanks in advance for the help (for your awesome work with this plugin!)

blueyed:
1. You should use `call neomake#quickfix#enable()`.

2. Not sure why you do not see flake8 + mypy.  Is it in the list, but not visible maybe? I.e. something would be wrong with auto-resizing the list (also changed recently).

petobens:
Yep, that was it (in both cases). Mmm is there any plan to improve the formatting a bit?

blueyed:
From your comparison:

1. I think the buffer name should get the same highlight (blue)
2. the mypy one looks off: it should have an error sign
3. Don't you like the line/col formatting/enhancement?

In general:
You should not use `set relativenumber` in the qf window (we should ensure that it is handled correctly, i.e. enforce `:setlocal number norelative`).

blueyed:
There is no plan to improve this, and @tweekmonster has been silent about his work since a while.
We need to progress from there/here.

blueyed:
Also the entries `nr` is not displayed at all.
I think it could be appended to the short maker name, or get its own column, e.g. between maker name and line/col info?
This would make sense for the shellcheck maker, whose short name is "SC" and therefore e.g. "SC2089" would be displayed as with the entry's text - where I would like to remove it from.

blueyed:
Feels like we should create a meta issue or project for this after all.

petobens:
Hi @blueyed I've finally revisited this one. Mmm coming back to my question:
> Is there a way to keep the same formatting and colors and just add the sign and linter?

I like the standard formatting but would simple like and indication to what linter each error/warnings corresponds to. Is there a way to achieve that? (i.e extend current formatting and attach something like `[mypy]` or `[pylint]` to each quickfix/location list entry)

blueyed:
@petobens 
Thanks for the feedback.  I've created new issues based on your comment.

> In summary: I would like to have something like `g:neomake_place_signs` but for the linter name.

I do not understand that one though.

petobens:
Thanks! 

> I do not understand that one though.

I meant that maybe we could have something like `g:neomake_show_linter_name` that controls whether to show (or not) the linter name in the same spirit that `g:neomake_place_signs` controls whether to show or not errors/warnings signs.

Issue #926, opened 2017-01-13
Expansion of '%:r.o' not supported?


derkling:
I want to defined a maker which build a single source of a Linux kernel tree.
Being a make based project, I usually use this command line:

```bash
make path/to/source.o
```

where `path/to/source.c` is the source file I want to compile. Notice that I need to specify the target to the make command (thus the usage of the `.o` extension in the previous command).

### Expected behavior

I would expect this custom maker:

```vim
let g:neomake_c_krnsrc_maker = {
        \ 'exe': 'make',
        \ 'args': ['ARCH=arm64', 'CROSS_COMPILE=aarch64-linux-gnu-', '%:r.o'], 
        \ }
```

to append the proper object file target corresponding to the current buffer when I call the maker with:

```vim
Neomake! c_krnsrc
```

AFAIU, the `!` avoid the appending of the current file name, which should be however provided by the expansion of the last arg (i.e. ` '%:r.o'`).

### Steps to reproduce

Using the previously defined custom maker and command I can see that the ` '%:r.o'` is not expanded:

```
[2017-01-13T19:18:05+0000 @1009.745, debug] [2] Running makers: c_krnsrc ([{'ft': '', 'errorformat': '%-G%f:%s:,%-G%f:%l: %#error: %#(Each undeclared identifier is reported only%.%#,%-G%f:%l: %#error: %#for each function it appears%.%#,%-GIn file included%.%#,%-G %#from %f:%l\,,%f:%l:%c: %trror: %m,%f:%l:%c: %tarning: %m,%I%f:%l:%c: note: %m,%f:%l:%c: %m,%f:%l: %trror: %m,%f:%l: %tarning: %m,%I%f:%l: note: %m,%f:%l: %m', 'args': ['ARCH=arm64', 'CROSS_COMPILE=aarch64-linux-gnu-', '%:r.o'], 'exe': 'make', 'remove_invalid_entries': 0, 'buffer_output': 1, 'name': 'c_krnsrc'}])
[2017-01-13T19:18:05+0000 @1009.747, verb ] [2] Starting async job: ['make', 'ARCH=arm64', 'CROSS_COMPILE=aarch64-linux-gnu-', '%:r.o']
[2017-01-13T19:18:05+0000 @1009.750, debug] [2.5] Vim job: {'status': 'run', 'stoponexit': 'term', 'exitval': 0, 'exit_cb': 0, 'channel': channel 5 open, 'process': 3425}
[...]

PR #930, opened 2017-01-15
Improve neomake#utils#ExpandArgs to handle %:r.o

blueyed:
Fixes https://github.com/neomake/neomake/issues/926.


blueyed:
No response, but why not.
At least it's tested (+1) and not documented (+1) = 2.

PR #938, opened 2017-01-17
Python MyPy requires --fast-parser flag for async/await support

Integralist:
I had to add the following to my `.vimrc` file to get async/await syntax support working...

```viml
let g:neomake_python_mypy_args = neomake#makers#ft#python#mypy()['args'] + ['--fast-parser']
```

It wasn't immediately obvious when I went sifting through the Neomake python maker code, until I stumbled across a blog post that mentions the need for the `--fast-parser` flag. 

I believe enough people use async/await syntax that we either add this flag by default or - as per this PR (as it's experimental support) - we at least make mention of it somewhere.

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -249,6 +249,7 @@ endfunction
 
 " Because this uses --silent-imports it requires mypy >= 0.4
 " It is annoying for new users to use MyPy without --silent-imports
+" For experimental async/await syntax support you'll need --fast-parser
 function! neomake#makers#ft#python#mypy() abort
     return {
         \ 'args': ['--silent-imports'],
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -1,5 +1,4 @@
 " vim: ts=4 sw=4 et
-let s:is_windows = has('win32') || has('win64') || has('win16') || has('dos32') || has('dos16')
 
 function! neomake#makers#ft#python#EnabledMakers() abort
     if exists('s:python_makers')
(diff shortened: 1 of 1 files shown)


blueyed:
I think we should just add the option by default then.

It is even recommended, except for Windows, so it should be done for non-Windows only then?!

> --fast-parser             enable fast parser (recommended except on Windows)

Integralist:
@blueyed also in recent MyPy version you'll see the deprecation warning:

```
--silent-imports has been replaced by --ignore-missing-imports --follow-imports=skip
```

blueyed:
@Integralist 
Ok, then I'd say to update this PR adopting those options, too.
Do you agree?

blueyed:
Please update this PR then accordingly.

Integralist:
@blueyed not really written VimL before so you'll have to excuse the code if it's not how it should be

blueyed:
Pushed a fixup: https://github.com/neomake/neomake/pull/938/commits/4bbf20e8000f7ec4a01d5b8b9797f51aa4462143.

I'll squash it when it's green.

Integralist:
Great stuff, thanks :+1:

blueyed:
@Integralist 
Thank you!

autoload/neomake/makers/ft/python.vim
```diff
@@ -251,9 +250,12 @@ endfunction
 " --fast-parser: adds experimental support for async/await syntax
 " --silent-imports: replaced by --ignore-missing-imports --follow-imports=skip
 function! neomake#makers#ft#python#mypy() abort
-    let defaults = ['--ignore-missing-imports', '--follow-imports=skip']
+    let args = ['--ignore-missing-imports', '--follow-imports=skip']
+    if neomake#utils#IsRunningWindows()
```
Integralist: @blueyed apologies but I've just come back to this and I realised there was a mistake in your refactor.

We only wanted to add `--fast-parser` for non-Windows user,  
where as your refactor looks to add it if you are a Windows user

PR #937, opened 2017-01-17
ignore general messages in perl maker

euclio:

PR #939, opened 2017-01-17
tests: include ft_perl

blueyed:
Ref: https://github.com/neomake/neomake/pull/937


euclio:
Hm, I am not able to reproduce locally.

blueyed:
Oh, me neither - will debug it.

blueyed:
The `laddexpr` was adding to an existing location list, should be fixed now.

PR #935, opened 2017-01-16
Fix "Invalid buffer id" with highlights after bwipe

blueyed:

PR #940, opened 2017-01-17
bufexists() function expects a number as a parameter.

datanoise:
Since bufnr is used as a key to various dicts like s:sign_queue,
s:placed_signs, etc. its value is converted to string. So we have to
explicitly convert it back to number when calling bufexists() function.

It fixes an issue when calling `:Neomake! clippy` doesn't clean the old signs. Introduced by #935 PR.


blueyed:
Nice find.
The problem is that dictionary keys are always strings..!
I will try to come up with a test and/or change existing ones to actually catch this.

PR #951, opened 2017-01-21
csslint: handle global warnings

blueyed:


blueyed:
We're ahead of Syntastic herre.. ;)

PR #954, opened 2017-01-22
Added support for g:neomake_php_phpmd_args_rules to override phpmd rules

codeinabox:


autoload/neomake/makers/ft/php.vim
```diff
@@ -41,8 +41,16 @@ function! neomake#makers#ft#php#phpcs() abort
...
+    else
+        call add(l:args, 'codesize,design,unusedcode,naming')
+    endif
+
     return {
-        \ 'args': ['%:p', 'text', 'codesize,design,unusedcode,naming'],
+        \ 'args': args,
         \ 'errorformat': '%E%f:%l%\s%m'
         \ }
 endfunction
```
blueyed: I think it is better to set `g:neomake_php_phpmd_args` directly?!
`let g:neomake_php_phpmd_args = ['%:p', 'text', 'your,config']`

codeinabox: But don't you want to force text output because I assume the error format is dependent on that?

blueyed: Dunno?!
If you mean that having an extra var makes it less likely to cause user errors, then that  might be a point.
However, adding per-maker config settings is not going to scale really.
I've thought about something like `g:neomaker_FT_MAKER_extra_args` that would be appended (since people tend to wrongly overwrite `_args` or just `_maker` itself accidentally, and getting the defaults is long (`neomake#makers#ft#php#phpmd().args`).

`tox` has an interesting concept of a `posargs` setting, which can have a default.
In this case it would default to default list, and could be changed by a user.
OTOH with other makers, you would want to configure `--exclude`, `--include` etc instead (and not just a single arg).

codeinabox: I agree, I saw a per-maker configuration for phpcs and thought they were okay but I think a generic post arguments is better solution and what [Syntastic uses](https://github.com/vim-syntastic/syntastic/blob/master/syntax_checkers/php/phpmd.vim)

blueyed: We should probably move this into a new issue then, and close this PR here?!

blueyed: Do you feel like working on such a generic solution by chance?

codeinabox:
Sounds good to me

autoload/neomake/makers/ft/php.vim
codeinabox: I can give it a go

blueyed:
Closing this to get if off the radar - we should have a new issue/PR for it.

autoload/neomake/makers/ft/php.vim
codeinabox: Are you thinking `extra` or `post` args? I am wondering if we need to be specific as would there be a time you need `pre` args?

blueyed: `post` sounds good to me, i.e. `neomaker_FT_MAKER_args_post`?!

It should come before any `append_file`.

blueyed: Please see https://github.com/neomake/neomake/issues/972#issuecomment-275800406.

PR #958, opened 2017-01-24
quickfix autocmds

bounceme:
for generic processing of quickfix and loclist etc.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -738,6 +738,11 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
             call setqflist(list, 'r')
         endif
     endif
+    if file_mode
+        silent doautocmd QuickFixCmdPost lgetfile
+    else
+        silent doautocmd QuickFixCmdPost cgetfile
+    endif
     if ignored_signs
         call neomake#utils#DebugMessage(printf(
                     \ 'Could not place signs for %d entries without line number.',
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -738,11 +738,6 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
             call setqflist(list, 'r')
         endif
     endif
-    if file_mode
-        silent doautocmd QuickFixCmdPost lgetfile
-    else
-        silent doautocmd QuickFixCmdPost cgetfile
-    endif
     if ignored_signs
         call neomake#utils#DebugMessage(printf(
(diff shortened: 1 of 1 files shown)


bounceme:
i expected it would work in this `if` : https://github.com/bounceme/neomake/blob/593beba3666f0f1e911dec27aafc98be99b4d5a9/autoload/neomake.vim#L734
but it didn't somehow

blueyed:
Why not also here?
https://github.com/bounceme/neomake/blob/593beba3666f0f1e911dec27aafc98be99b4d5a9/autoload/neomake.vim#L832-L838

I assume `QuickFixCmdPre` get run automatically already?

> **QuickFixCmdPost**
`:h QuickFixCmdPost` says "after a quickfix command is run, before jumping to the
first location."

Is the `lgetfile/cgetfile` arbitrary? (or what dispatch does)
I think it should match the actual command, shouldn't it?

And instead of `silent doautocmd` it should be checked if the autocommand exists, and/or it should be run with `<nomodeline>` (see `neomake#utils#hook`).

As with e.g. `grep` it gets run only at the end, which would be similar to `NeomakeFinished`, isn't it?

bounceme:
> Why not also here?
https://github.com/bounceme/neomake/blob/593beba3666f0f1e911dec27aafc98be99b4d5a9/autoload/neomake.vim#L832-L838

@blueyed you're right

 about the event type, I don't really know right now, this is still very w.i.p. I can change it to be `make` or both: `cgetfile,make`

bounceme:
strange that caddexpr can't be used with `quickfixcmdpost/pre`

bounceme:
After looking into the code more: I think the current code would work if there isn't any other quickfix list changing that happens after where i added the `doautocmd`'s. I am not quite sure what would be considered a good location for `quickfixcmdpre`. right before the `caddexpr` , or is there more going on related to quickfix lists before that is executed?

bounceme:
https://github.com/neomake/neomake/pull/958/files#diff-1ceaba5126b0cf477e72f87cb2767af0R852 might be a better place actually

bounceme:
https://github.com/tpope/vim-dispatch/blob/a54f2c5e18b8c2aad8c6f8ba474760e70fdaaca3/autoload/dispatch.vim#L836 tpope did that also, which is a convincing argument when talking about vimscript

bounceme:
I think I'm the only one who wants this. maybe it be good to have neomake isolated from vim's builtin things, with its own au cmds

PR #961, opened 2017-01-24
Fix error when running neomake in vader files in another repo

JelteF:
Running Neomake after running Vader was giving me errors in the vader.vim repo.


blueyed:
Thanks!

Issue #959, opened 2017-01-24
Wrong quoting with Vim on Windows (pylint, vint, etc)


jwmng:
Windows 10, Vim 8.0 (1-69) compiled 6/11/2016
(this is my first issue on a public project ever, I hope I'm doing this right)

### Expected behavior
 `:Neomake pylint` should run properly and put an invalid syntax error in the location list.

### Steps to reproduce
```vim
:let g:neomake_verbose = 3
:e ~/Desktop/a.py
:Neomake pylint
```

With `a.py`:
```python
print(
```
### Suspected cause
The default `--msg-template` is wrapped in quotes to handle whitespace in the parameter as set in `ft/python.vim`:
```vim
function! neomake#makers#ft#python#pylint() abort
    return {
        \ 'args': [
            \ '--output-format=text',
            \ '--msg-template="{path}:{line}:{column}:{C}: [{symbol}] {msg}"',
(...)
```

This results in the following async call (newlines added for readability), with double double quotes in `--msg-template`:

```cmd
 'C:\WINDOWS\system32\cmd.exe /c "pylint --output-format=text 
--msg-template=""{path}:{line}:{column}:{C}: [{symbol}] {msg}"" 
--reports=no C:\Users\jwm\Code\task\tasktools.py"'
```
This causes `pylint` to see the last two arguments (`{symbol}` and `{arg}`) as files to check, resulting in the output (again newlines added):

```
Neomake [252.237]: [2.1] stdout: pylint:
 ['************* Module [{symbol}]^M', '[{symbol}]:1:0:F:^M', 
  '************* Module {msg}^M', '{msg}:1:0:F:^M', 
  '************* Module a^M', 'a.py:2:0:E:^M', '']
```

And no errors show up.

### What have you tried?
To remove the space from `--msg-template` and `errorformat`, but it breaks parsing. 

#### Neomake debug information
Async support: 1
Current filetype: python

##### Enabled makers
For the current filetype (with :Neomake):
````
 - pylint
   postprocess: function('neomake#makers#ft#python#PylintEntryProcess')
   errorformat: '%A%f:%l:%c:%t: %m,%A%f:%l: %m,%A%f:(%l): %m,%-Z%p^%.%#,%-G%.%#'
   args: ['--output-format=text', '--msg-template="{path}:{line}:{column}:{C}: [{symbol}] {msg}"', '--reports=no']
   get_argv: function('4', {'get_argv': function('4')})
   exe: 'pylint'
````

[...]

Issue #724, opened 2016-10-20
Setup / fix tests for Windows


blueyed:
- [ ] Tests should work on Windows
 - [ ] CI should be run on Windows, e.g. using [AppVeyor](https://ci.appveyor.com/)

jwmng:
Vim80 Gvim, Windows 10, Neomake 8f6aba3 with the fix by @jsfaint in #964 
Using `make-3.8.1` from the gnuwin32 project.

The makefile is not written for use in windows (uses `/bin/bash` and `/dev/null`) and cygwin throws an error as well so I had to run Vader manually.

Full output including debug: [gist](https://gist.github.com/jwmng/a87bab691bbc505cd3b487f97b18e5b4)

- Success/Total: 43/99 (assertions: 224/255)
- Elapsed time: 69.020472 sec.

### `:Vader *` output
````
neomake.vader|7 error| Error with non-existing filename for 2nd maker (#23)
neomake.vader|8 error| Hook execution gets logged (#39)
neomake.vader|9 error| Test Neomake on errors.sh with one maker (#56)
neomake.vader|9 error| Test Neomake on errors.sh with two makers (#58)
neomake.vader|9 error| Neomake: handle result for current window (#70)
neomake.vader|9 error| Neomake: handle output for removed window (#82)
neomake.vader|9 error| Neomake: handle wiped buffer (#87)
neomake.vader|9 error| NeomakeSh: true (#104)
neomake.vader|9 error| NeomakeSh: echo foo (#120)
neomake.vader|9 error| NeomakeSh: non-existing command (#136)
neomake.vader|9 error| NeomakeSh!: handle unfinished output on exit (#148)
neomake.vader|9 error| NeomakeSh: buffers output (#166)
neomake.vader|9 error| NeomakeSh: project: handle removed window on exit (quit) (#182)
neomake.vader|9 error| NeomakeSh: project: handle removed window on exit (bwipe) (#198)
neomake.vader|9 error| Neomake with windo (#248)
neomake.vader|9 error| NeomakeSh: two jobs after each other (#282)
neomake.vader|9 error| Neomake#Make: two jobs after each other (#296)
neomake.vader|9 error| Neomake: non-existing and proper job (#320)
neomake.vader|9 error| Neomake: remove_invalid_entries (#336)
neomake.vader|9 error| Neomake: append_file from settings (#341)
neomake.vader|10 error| Postprocessing should update list (#345)
[...]

PR #964, opened 2017-01-25
Fix quoting of argv for Vim on Windows

blueyed:
Fixes https://github.com/neomake/neomake/issues/959

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -290,11 +290,8 @@ function! s:maker_base.get_argv(...) abort dict
         else
             let argv = exe . (len(args) ? ' ' . args : '')
         endif
-
         if !has('nvim')
-            if neomake#utils#IsRunningWindows()
-                let argv = args_is_list ? join(argv) : argv
-            elseif !args_is_list
+            if !args_is_list
                 " Have to use a shell to handle argv properly (Vim splits it
                 " at spaces).
                 let argv = [&shell, &shellcmdflag, argv]
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -290,10 +290,13 @@ function! s:maker_base.get_argv(...) abort dict
         else
             let argv = exe . (len(args) ? ' ' . args : '')
         endif
+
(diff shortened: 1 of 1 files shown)


jsfaint:
This PR Not works for me 😢 
But with the following patch on master branch 8f6aba3531c81a3d3d291721f7515f8efa2e59f8, it works now
```
diff --git a/autoload/neomake.vim b/autoload/neomake.vim
index 75436d7..2fa6035 100644
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -291,9 +291,7 @@ function! s:maker_base.get_argv(...) abort dict
             let argv = exe . (len(args) ? ' ' . args : '')
         endif
         if !has('nvim')
-            if neomake#utils#IsRunningWindows()
-                let argv = &shell.' '.&shellcmdflag.' '.shellescape(args_is_list ? join(argv) : argv)
-            elseif !args_is_list
+            if !args_is_list
                 let argv = [&shell, &shellcmdflag, argv]
             endif
         endif
```

blueyed:
@jwmng
Does this also work for you?

blueyed:
@jsfaint 
Would be good to remove this special case after all, thanks for looking into it!

I think this PR or in general I've tried to take the following into account (from `:h job_start`):
> {command} can be a String.  This works best on MS-Windows.  On Unix it is split up in white-separated parts to be passed to execvp().  Arguments in double quotes can contain white space.

But if the shell workaround works in general, it might be good.

Any chance that you or @jwmng could run the test suite?
See https://github.com/neomake/neomake/issues/724 - it would be nice to have CI for it after all.

blueyed:
Updated this PR/branch according to @jsfaint's patch.

jwmng:
Running the test suite using Vader in gvim, will put results at #724

blueyed:
@jwmng 
And does the updated PR fix it for you already (when using, not with the tests)?

jwmng:
Forgot to mention that yes it works fine

blueyed:
Thanks!

PR #967, opened 2017-01-25
fixes MakerIsAvailable function for makers with funcref as exe

jreinert:
before these changes MakerIsAvailable would fail when trying to test maker.exe if it is a funcref or a dict with a funcref

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -204,14 +204,21 @@ function! neomake#utils#MakerIsAvailable(ft, maker_name) abort
         " for our purposes
         return 1
     endif
+
     let maker = neomake#GetMaker(a:maker_name, a:ft)
     if empty(maker)
         return 0
     endif
-    if !has_key(s:available_makers, maker.exe)
-        let s:available_makers[maker.exe] = executable(maker.exe)
+
+    if (type(maker.exe) == type(function('tr')))
+        let l:executable = call(maker.exe, [])
+    elseif (type(maker.exe) == type({}))
+        let l:executable = call(maker.exe.fn, [], maker.exe)
+    else
+        let l:executable = maker.exe
     endif
-    return s:available_makers[maker.exe]
+
+    return executable(l:executable)
 endfunction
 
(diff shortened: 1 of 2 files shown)


autoload/neomake/utils.vim
```diff
@@ -204,14 +204,23 @@ function! neomake#utils#MakerIsAvailable(ft, maker_name) abort
...
+            let l:executable = call(maker.exe, [])
+        elseif (type(maker.exe) == type({}))
+            let l:executable = call(maker.exe.fn, [], maker.exe)
+        else
+            let l:executable = maker.exe
+        endif
+        let s:available_makers[maker.name] = executable(l:executable)
     endif
-    return s:available_makers[maker.exe]
+    return s:available_makers[maker.name]
```
blueyed: This somehow reverts https://github.com/jreinert/neomake/commit/2be7919475b1e8a728e32fe8b5e321ee0853aa18?!
Let's maybe just drop the caching here?!

jreinert: @blueyed you're right of course! I got rid of the caching.

blueyed:
Thanks!

1. I do not like the code duplication (with/from `s:maker_base.get_argv`).

2. Also, `maker.fn` could be used to set `.exe` after all, but is only called in `s:MakeJob` (where `maker.get_argv` is called, too).

Not sure about 2., but 1. could be moved into a function maybe?
OTOH I think we should deprecate the custom function callbacks, and that `maker.fn` should be used.

blueyed:
Thanks for the test already!

jreinert:
@blueyed so what should i do about 2. then? :)

blueyed:
I think we could call `maker.fn` here already?
But also later still, since it's meant to be dynamic.
Would result in two calls though.

Turns out that it take this opportunity to refactor quite a lot again, and tests are finally green.. will probably submit it later or tomorrow.

blueyed:
See https://github.com/neomake/neomake/pull/980.
Closing it in favor of that one.

PR #968, opened 2017-01-26
mapping credo error types to neomake

epilgrim:
Adding missing mapping of entry types from elixir credo to neomake.

Full credit to @emilsoman in https://github.com/neomake/neomake/pull/300#issuecomment-244722296

### DIFF
--- a/autoload/neomake/makers/ft/elixir.vim
+++ b/autoload/neomake/makers/ft/elixir.vim
@@ -7,7 +7,7 @@
 " D -> Software design suggestions
 " R -> Readability suggestions
 " Map structure {CredoError: NeomakeType, ...}
-let s:neomake_elixir_credo_type_map = {
+let s:neomake_elixir_credo_config_typemap = {
     \ 'F': 'W',
     \ 'C': 'W',
     \ 'D': 'I',
@@ -22,8 +22,8 @@ endfunction
 
 function! neomake#makers#ft#elixir#PostprocessCredo(entry) abort
     let type = toupper(a:entry.type)
-    let type_map = extend(s:neomake_elixir_credo_type_map,
-                \ get(g:, 'neomake_elixir_credo_type_map', {}))
+    let type_map = extend(s:neomake_elixir_credo_config_typemap,
+                \ get(g:, 'neomake_elixir_credo_config_typemap', {}))
     if has_key(type_map, type)
         let a:entry.type = type_map[type]
     endif
--- a/autoload/neomake/makers/ft/elixir.vim
+++ b/autoload/neomake/makers/ft/elixir.vim
@@ -1,5 +1,7 @@
 " vim: ts=4 sw=4 et
 
+let g:neomake_elixir_credo_type_map= get(g:, 'neomake_elixir_credo_type_map', {'F': 'W','C': 'W','D': 'I','R': 'I'})
+
 function! neomake#makers#ft#elixir#PostprocessEnforceMaxBufferLine(entry) abort
     let buffer_lines = str2nr(line('$'))
     if (buffer_lines < a:entry.lnum)
@@ -7,21 +9,18 @@ function! neomake#makers#ft#elixir#PostprocessEnforceMaxBufferLine(entry) abort
     endif
 endfunction
 
+" Credo error types
+" F -> Refactoring opportunities
+" W -> Warnings
+" C -> Convention violation
+" D -> Software design suggestions
+" R -> Readability suggestions
 function! neomake#makers#ft#elixir#PostprocessCredoErrorType(entry) abort
-    if a:entry.type ==# 'F'      " Refactoring opportunities
-        let type = 'W'
-    elseif a:entry.type ==# 'D'  " Software design suggestions
-        let type = 'I'
-    elseif a:entry.type ==# 'W'  " Warnings
-        let type = 'W'
-    elseif a:entry.type ==# 'R'  " Readability suggestions
-        let type = 'I'
-    elseif a:entry.type ==# 'C'  " Convention violation
-        let type = 'W'
-    else
-        let type = 'M'           " Everything else is a message
+    let type = toupper(a:entry.type)
+    let type_map = g:neomake_elixir_credo_type_map
+    if has_key(type_map, type)
+        let a:entry.type = type_map[type]
     endif
-    let a:entry.type = type
 endfunction
 
 function! neomake#makers#ft#elixir#EnabledMakers() abort
--- a/autoload/neomake/makers/ft/elixir.vim
+++ b/autoload/neomake/makers/ft/elixir.vim
@@ -1,6 +1,17 @@
 " vim: ts=4 sw=4 et
 
-let g:neomake_elixir_credo_type_map= get(g:, 'neomake_elixir_credo_type_map', {'F': 'W','C': 'W','D': 'I','R': 'I'})
+" Credo error types
+" F -> Refactoring opportunities
+" W -> Warnings
+" C -> Convention violation
+" D -> Software design suggestions
+" R -> Readability suggestions
+" Map structure {CredoError: NeomakeType, ...}
+let s:neomake_elixir_credo_type_map = get(
+    \ g:,
+    \ 'neomake_elixir_credo_type_map',
+    \ {'F': 'W','C': 'W','D': 'I','R': 'I'}
+\ )
 
 function! neomake#makers#ft#elixir#PostprocessEnforceMaxBufferLine(entry) abort
     let buffer_lines = str2nr(line('$'))
@@ -9,15 +20,9 @@ function! neomake#makers#ft#elixir#PostprocessEnforceMaxBufferLine(entry) abort
     endif
 endfunction
 
-" Credo error types
-" F -> Refactoring opportunities
-" W -> Warnings
-" C -> Convention violation
-" D -> Software design suggestions
-" R -> Readability suggestions
-function! neomake#makers#ft#elixir#PostprocessCredoErrorType(entry) abort
+function! neomake#makers#ft#elixir#PostprocessCredo(entry) abort
     let type = toupper(a:entry.type)
-    let type_map = g:neomake_elixir_credo_type_map
+    let type_map = s:neomake_elixir_credo_type_map
     if has_key(type_map, type)
         let a:entry.type = type_map[type]
     endif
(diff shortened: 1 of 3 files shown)


autoload/neomake/makers/ft/elixir.vim
```diff
@@ -7,6 +7,23 @@ function! neomake#makers#ft#elixir#PostprocessEnforceMaxBufferLine(entry) abort
...
+        let type = 'W'
+    elseif a:entry.type ==# 'R'  " Readability suggestions
+        let type = 'I'
+    elseif a:entry.type ==# 'C'  " Convention violation
+        let type = 'W'
+    else
+        let type = 'M'           " Everything else is a message
+    endif
+    let a:entry.type = type
+endfunction
```
blueyed: `type` `M` is something not being used yet AFAIK?.
What about `I` for it?
Or does it make sense?

epilgrim: in closer inspection, credo shouldn't give us back (at least by default) anything outside of those 5 groups. I will remove that part.
Then, I was thinking that perhaps users might have different ideas of what a translation should be. so I've created a configurable variable to override the defaults. 
Please let me know what you think

epilgrim:
I was thinking that perhaps users might have different ideas of what a translation should be. so I've created a configurable variable to override the defaults.
Please let me know what you think

autoload/neomake/makers/ft/elixir.vim
```diff
@@ -1,12 +1,28 @@
...
         let a:entry.lnum = buffer_lines
     endif
 endfunction
 
+" Credo error types
+" F -> Refactoring opportunities
+" W -> Warnings
+" C -> Convention violation
+" D -> Software design suggestions
+" R -> Readability suggestions
```
blueyed: Nice.
The documentation should be moved to the mapping above.

autoload/neomake/makers/ft/elixir.vim
```diff
@@ -1,12 +1,28 @@
 " vim: ts=4 sw=4 et
 
+let g:neomake_elixir_credo_type_map= get(g:, 'neomake_elixir_credo_type_map', {'F': 'W','C': 'W','D': 'I','R': 'I'})
```
blueyed: Please add a space before `=`.

blueyed: Also split it on two or more lines, please.

blueyed: Please use `s:` for the default, and then use it with `get()` below - it should not set this var into the global namespace.

autoload/neomake/makers/ft/elixir.vim
```diff
@@ -1,12 +1,28 @@
...
+" W -> Warnings
+" C -> Convention violation
+" D -> Software design suggestions
+" R -> Readability suggestions
+function! neomake#makers#ft#elixir#PostprocessCredoErrorType(entry) abort
+    let type = toupper(a:entry.type)
+    let type_map = g:neomake_elixir_credo_type_map
+    if has_key(type_map, type)
+        let a:entry.type = type_map[type]
+    endif
```
blueyed: This could be simplified with `map()`, but it's fine I suppose.

autoload/neomake/makers/ft/elixir.vim
```diff
@@ -23,6 +39,7 @@ function! neomake#makers#ft#elixir#credo() abort
     return {
       \ 'exe': 'mix',
       \ 'args': ['credo', 'list', '%:p', '--format=oneline'],
+      \ 'postprocess': function('neomake#makers#ft#elixir#PostprocessCredoErrorType'),
```
blueyed: Maybe just `PostprocessCredo` for the function name?

autoload/neomake/makers/ft/elixir.vim
epilgrim: sorry, but I didn't get this comment

epilgrim:
addressed some of the comments. I'll wait for clarification for the one left

blueyed:
What's left?
I've pushed a fixup which what I've meant.

autoload/neomake/makers/ft/elixir.vim
blueyed: No worries.
I've mean that `map()` could be used here.. but it's more like using `get()` is appropriate:

```diff
diff --git i/autoload/neomake/makers/ft/elixir.vim w/autoload/neomake/makers/ft/elixir.vim
index 704143e..dbe4111 100644
--- i/autoload/neomake/makers/ft/elixir.vim
+++ w/autoload/neomake/makers/ft/elixir.vim
@@ -21,12 +21,9 @@ function! neomake#makers#ft#elixir#PostprocessEnforceMaxBufferLine(entry) abort
 endfunction
 
 function! neomake#makers#ft#elixir#PostprocessCredo(entry) abort
-    let type = toupper(a:entry.type)
     let type_map = extend(s:neomake_elixir_credo_type_map,
                 \ get(g:, 'neomake_elixir_credo_type_map', {}))
-    if has_key(type_map, type)
-        let a:entry.type = type_map[type]
-    endif
+    let a:entry.type = get(type_map, toupper(a:entry.type), a:entry.type)
 endfunction
 
 function! neomake#makers#ft#elixir#EnabledMakers() abort
```

I am not sure it's more readable though, and there are no tests - so let's skip it.

blueyed: Regarding tests: there are some `tests/ft_*` cases already, in case you want to dig into it.

blueyed:
@epilgrim 
What do you think of my fixup?
Does it still work for you?

tests/ft_elixir.vader
```diff
@@ -14,3 +14,24 @@ Execute (Credo: postprocess: global configuration):
...
+
+Given (A file with 2 lines):
+  some
+  lines
+Execute (Mix: postprocess: line inside buffer is not modified):
+  function! F(entry)
+    call neomake#makers#ft#elixir#PostprocessEnforceMaxBufferLine(a:entry)
+    return a:entry
+  endfunction
+  AssertEqual F({'lnum': 1}).lnum, 1
```
blueyed: You could do both assertions in the same test..
Also for the tests above.

tests/ft_elixir.vader
```diff
@@ -14,3 +14,24 @@ Execute (Credo: postprocess: global configuration):
...
+
```
blueyed: Please remove the empty line at the end.

autoload/neomake/makers/ft/elixir.vim
```diff
@@ -1,12 +1,34 @@
...
     let buffer_lines = str2nr(line('$'))
     if (buffer_lines < a:entry.lnum)
         let a:entry.lnum = buffer_lines
     endif
 endfunction
 
+function! neomake#makers#ft#elixir#PostprocessCredo(entry) abort
+    let type = toupper(a:entry.type)
+    let type_map = extend(s:neomake_elixir_credo_type_map,
+                \ get(g:, 'neomake_elixir_credo_type_map', {}))
```
blueyed: Regarding the variable name, we might was to use something more namespacy, e.g. `g:neomake_elixir_credo_config_typemap`, i.e. `_config` for anything that is maker specific.
/cc @tweekmonster

tests/ft_elixir.vader
epilgrim: ok, updated. I'm used to one intent per test :)

autoload/neomake/makers/ft/elixir.vim
blueyed: @tweekmonster 
Any input on that from you?

tweekmonster: > Regarding the variable name, we might was to use something more namespacy, e.g. g:neomake_elixir_credo_config_typemap, i.e. _config for anything that is maker specific.

I agree.  The names should have a predictable pattern.  I actually would prefer maker configs to be dicts since that's a pretty long name, but it's a little late for that now.

> /cc @tweekmonster

Because I'm super opinionated about small details right?!  😀

blueyed: > I actually would prefer maker configs to be dicts since that's a pretty long name, but it's a little late for that now.

Please elaborate.. it's never too late, and there are not many custom maker options yet.

tweekmonster: > The names should have a predictable pattern.

To clarify what I mean by this: I think it could be useful to have a way to debug makers specifically by inspecting their configs with something like: 

```vim
getcompletion('g:neomake_<ft>_<maker>_config_.*', 'var')
```

epilgrim: well, now that we talk about that, i'm wondering if this config is actually needed. 
Perhaps mapping all the credo "errors" should be mapped to neomake warnings, without any configuration whatsoever. 
In the end, if you have something configured in your `.credo.exs` file, it's because you want  to enforce that rule.
In the beginning, I thought it might be useful, but if this is something that is not currently done in neomake, perhaps it is not the best case to introduce it

blueyed: @tweekmonster 
> `getcompletion('g:neomake_<ft>_<maker>_config_.*', 'var')`

Where's the dict there?
Wouldn't that rather be `g:neomake_<ft>_<maker>_config` then?

(TIL about `getcompletion()` though - there are some places where this could be useful (e.g. `neomake#Get(Project)Makers`).)

@epilgrim 
It's a topic that came up several times already, sorry to hijack your PR.. ;)
Apart from that I think there should be no unnecessary config options, but if it's there already we might want to keep it for now?!
But it might also be better to have something generic, that allows users to apply such mappings, e.g. a generic postprocess hook like mentioned in https://github.com/neomake/neomake/pull/977.

tweekmonster: > Where's the dict there?

I mentioned it's too late to have a dict for configs.  The variable name pattern is well established now.  I was just saying that since we're stuck it should at least be consistent for that reason.

> In the end, if you have something configured in your .credo.exs file, it's because you want to enforce that rule.

Counter argument: I do a lot of Python work in a "project" setting, so I'll have linter configs in the repo dir.  But, I have a lot more one-off scripts that are annoying to see warnings in.  I'd like to be able to pass options to the linter without creating yet another dot file in my home directory.

If it's an option you think people will find useful, it should be there.  Particularly users that aren't inclined to look at Neomake's source to figure things out.

tweekmonster:
@epilgrim @blueyed I think 2 approvals are a good enough reason to merge this.  I'll squash and merge in a couple minutes.

PR #969, opened 2017-01-26
Python: Use fix for flake8 column for redefinition error as well

JelteF:
This had a similar bug, where the col (and thus also the underline) was put at
the def or import already instead of under the actual redefinition.

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -106,19 +106,32 @@ function! neomake#makers#ft#python#Flake8EntryProcess(entry) abort
             " errors.
             let l:view = winsaveview()
             call cursor(a:entry.lnum, a:entry.col)
+            " The number of lines to give up searching afterwards
+            let l:search_lines = 5
 
-            if searchpos('from', 'cnW', a:entry.lnum)[1] == a:entry.col
+            if searchpos('\<from\>', 'cnW', a:entry.lnum)[1] == a:entry.col
                 " for 'from xxx.yyy import zzz' the token looks like
                 " xxx.yyy.zzz, but only the zzz part should be highlighted. So
                 " this discards the module part
                 let l:token = split(l:token, '\.')[-1]
+
+                " Also the searhch should be started at the import keyword.
+                " Otherwise for 'from os import os' the first os will be
+                " found. This moves the cursor there.
+                echom search('\<import\>', 'cW', a:entry.lnum + l:search_lines)
             endif
 
             " Search for the first occurrence of the token and highlight in
             " the next couple of lines and change the lnum and col to that
             " position.
-            let l:search_lines = 5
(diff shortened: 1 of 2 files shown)


autoload/neomake/makers/ft/python.vim
```diff
@@ -101,8 +101,9 @@ function! neomake#makers#ft#python#Flake8EntryProcess(entry) abort
     if strlen(l:token)
         " remove quotes
         let l:token = substitute(l:token, "'", '', 'g')
-        if a:entry.type ==# 'F' && a:entry.nr == 401
-            " The unused import error column is incorrect
+        if a:entry.type ==# 'F' && (a:entry.nr == 401 ||  a:entry.nr == 811)
+            " The unused column is incorrect for import errors and redefinition
+            " errors
```
blueyed: Dot at the end, please.

What do you think about having a test for this? :)

blueyed: Added the dot and a test.

tests/ft_python.vader
```diff
@@ -38,3 +38,31 @@ Execute (python: pylama: errorformat):
...
+  \ {'lnum': 2, 'bufnr': bufnr('%'), 'col': 1, 'valid': 1, 'vcol': 0,
+  \  'nr': 811, 'type': 'F', 'pattern': '',
+  \  'text': 'redefinition of unused ''os'' from line 1'}]
+
+  let e = copy(getloclist(0))[0]
+  call neomake#makers#ft#python#Flake8EntryProcess(e)
+  AssertEqual e, {
+  \ 'lnum': 2,
+  \ 'bufnr': bufnr('%'),
+  \ 'col': 8,
```
blueyed: This should be 19 instead of 8, i.e. it should underline the 2nd `os` (which overrides the previous name).

blueyed: OTOH this should be really fixed in flake8 after all, shouldn't it?  (I think we talked about it before..)

blueyed:
@JelteF 
ping?

tests/ft_python.vader
JelteF: The ast module (which flake8 uses) really sucks in this regard. It doesn't have allow access to the position of these sub parts of the AST. I tried to fix it there for the previous error first, but gave up because of that. I'll look into the 'os'  bug tomorrow, but it should be super rare to encounter.

blueyed: IIRC it is more about preferring the last match instead of the first in this regard.

JelteF:
I improved the matching of wrong tokens, but I think there will probably always be ones if you craft the string carefully enough. I think this should be fine now.

JelteF:
Did you mean to merge instead of just approving?

blueyed:
Wanted to squash the fixups, which I've done now.. will merge it when it's green.

blueyed:
Thanks!

PR #970, opened 2017-01-27
removed ignored readability checkers

epilgrim:
They should be ignored in the .credo.exs file if the user wants to do so


blueyed:
Thanks!

PR #974, opened 2017-01-28
Use my friendly Vader fork

blueyed:
TODO
 - [x] tests should be green first

Via https://github.com/junegunn/vader.vim/pull/107.


blueyed:
Via https://github.com/junegunn/vader.vim/pull/107.
First the tests in there should be green though.

Issue #510, opened 2016-07-16
Add support for `:verb Neomake`


blueyed:
Instead of setting `g:neomake_verbose` globally, `:verb Neomake` should enable verbose logging for the jobs started from it.
This requires to add a `verbose` attribute to jobs being run and use that for filtering messages.
It should probably use the `max()` from `g:neomake_verbose` and `&verbose` then.

PR #976, opened 2017-01-29
Add &verbose to g:neomake_verbose

blueyed:
Fixes https://github.com/neomake/neomake/issues/510.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -7,17 +7,22 @@
 ### Steps to reproduce
 
 <!--
-Include if relevant.
+Provide steps to reproduce the issue, if relevant.
 
-Please execute `:let g:neomake_verbose = 3` to enable verbose output, which
-might help showing the issue already.
-Look at `:messages` to see them.
+1. Please execute `:let g:neomake_verbose = 3` first, to enable verbose
+   output/logging in general, or set `g:neomake_logfile` to a file path.
+2. Look at `:messages` (or the logfile) for the generated output, which might
+   help revealing the issue already.
 -->
 
-### Output of the ":verb NeomakeInfo" command
+### Output from (verbose) NeomakeInfo
 
 <!--
 Paste the output from `:verbose NeomakeInfo` here.
-You can use `:redir @+> | silent verb NeomakeInfo | redir END` to copy it to
-your clipboard.
+You can use the following command to copy it directly to your clipboard:
+
+    :redir @+> | silent verb NeomakeInfo | redir END
+
+If you have used `g:neomake_logfile`, then please paste the contents of the
+file here, too.
 -->
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -555,7 +555,8 @@ function! s:Make(options) abort
         return []
     endif
     call neomake#utils#DebugMessage(printf('Running makers: %s',
-                \ join(map(copy(enabled_makers), 'v:val.name'), ', ')))
+                \ join(map(copy(enabled_makers), 'v:val.name'), ', ')),
+                \ {'make_id': s:make_id})
     let maker = {}
     while len(enabled_makers)
         let maker = remove(enabled_makers, 0)
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -54,6 +54,11 @@ function! neomake#GetStatus() abort
                 \ }
 endfunction
 
+" Not documented, only used internally for now.
+function! neomake#GetMakeOptions(...) abort
+    return s:make_info[a:0 ? a:1 : s:make_id]
+endfunction
+
 function! neomake#ListJobs() abort
     call neomake#utils#DebugMessage('call neomake#ListJobs()')
     for jobinfo in values(s:jobs)
@@ -534,6 +539,7 @@ function! s:Make(options) abort
         let s:make_id += 1
         let s:make_info[s:make_id] = {
                     \ 'cwd': getcwd(),
+                    \ 'verbosity': get(g:, 'neomake_verbose', 1) + &verbose,
                     \ }
 
         if file_mode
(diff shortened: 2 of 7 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -10,11 +10,10 @@
 Include if relevant.
 
 Please execute `:let g:neomake_verbose = 3` to enable verbose output, which
-might help showing the issue already.
-Look at `:messages` to see them.
+might help showing the issue already.  Look at `:messages` for its output.
```
blueyed: We could mention `3verb Neomake` here?!

tweekmonster: Maybe replace the paragraph with:

```
Execute `:3verb Neomake` and look at `:messages` for its output.
```

And add at the top:

```
Please ensure you're using the latest version of Neomake before posting.
```

Don't give the other option since we can consider that to be more of a development variable.

autoload/neomake.vim
```diff
@@ -531,6 +536,7 @@ function! s:Make(options) abort
         let s:make_id += 1
         let s:make_options[s:make_id] = {
                     \ 'cwd': getcwd(),
+                    \ 'verbosity': get(g:, 'neomake_verbose', 1) + &verbose,
```
tweekmonster: Why is it added?  Throwing in arithmetic complicates (slightly) a simple concept.

Shouldn't `&verbose` simply override it temporarily the same way `:verbose` does?  IMO, `:verb Neomake` is the same as `let g:neomake_verbose = 1 | Neomake | unlet! g:neomake_verbose`.

blueyed: It's not possible to detect if `:verb` is used, except for assuming that it is when it's `> 0`.
This means that `&verbose` alone cannot be used as the default, since that would be `0` then.
The default is 1 currently, but that's not set in stone, and logging levels should be reviewed anyway - currently I'm just stuffing most of it into the debug level (3).

tweekmonster: > The default is 1 currently, but that's not set in stone, and logging levels should be reviewed anyway - currently I'm just stuffing most of it into the debug level (3).

This is what I mean.  If `&verbose` is `0`, then just use `g:neomake_verbose` since its default is `1`.

```vim
let verbosity = &verbose ? &verbose : get(g:, 'neomake_verbose', 1)
```

tweekmonster: It's a little...verbose, though 😂

blueyed: I see.. :)
But then `:verb Neomake` is just using the default verbosity (1), which is what I wanted to avoid, since `:verbose` should make things more verbose after all.

tweekmonster: Well then, maybe it can be `4 - min([3, &verbose])` so that increasing the count makes it quieter?

Alternatively just make `:verbose` always cause the most verbose output regardless of the count?  Adding `:verbose` in front of the command is basically saying "tell me everything" in most of Vim's builtin commands.

In any case, I think having this feature makes `g:neomake_verbose` deprecated for users.  It would be more of a convenience for maker devs.  So, doing the addition isn't a terrible idea to me.

.github/ISSUE_TEMPLATE.md
blueyed: > Execute `:3verb Neomake` and look at `:messages` for its output.

That limits it to the actual command being used, and some issues are about autocommands etc, so I'll leave it.

autoload/neomake.vim
blueyed: > So, doing the addition isn't a terrible idea to me.

I think so, too.. I'll sleep over it.
Thanks for your valuable feedback once again! :)

jamessan: > Adding :verbose in front of the command is basically saying "tell me everything" in most of Vim's builtin commands.

Not really.  `:verbose` takes a count which acts like setting `&verbose` to that number.  That is, `:10verbose {cmd}` is like `:set verbose=10 | {cmd} | set verbose=0`.  It does make sense that `:verbose Neomake` should have more logging than typical, but that shouldn't preclude supporting multiple levels of verbosity.

tweekmonster: @jamessan I'm referring to my experience with it, which I imagine isn't unique.  Personally, I never read the doc on `:verbose` until a few days ago.  It was something I picked up along the way.  Everything I regularly use it with (`:verbose nmap`, `:verbose autocmd`, etc) appears to give you the most verbose output.  `:100verbose autocmd` doesn't make it any more informative, so I would assume that's the case for everything.

Only things I saw in the docs where the levels mattered is for debugging VimL execution, and it looks like the levels are for filtering message origins (autocmd execution, diff, exceptions), not message types.

From my point of view, `:verbose Neomake` would be something a developer asks a user to do for troubleshooting.  Wouldn't we want it to be as verbose as possible in that case?  `g:neomake_verbose` then would be for testing and passive logging in development.

blueyed: @tweekmonster 
Makes sense, and would make it easier to explain/understand.

tweekmonster:
A quick thought that has no bearing on this PR: Since the verbosity is being added, would it be useful to print the verbosity level as well?

blueyed:
> Since the verbosity is being added, would it be useful to print the verbosity level as well?

Good point.
AFAIR this is being done for verbosity=3.

tweekmonster:
> AFAIR this is being done for verbosity=3.

I was thinking that if `&verbose > 0`, it might be useful to print it regardless of the level to give an indication that `g:neomake_verbose` was augmented.  I also think it's not super important and this can be merged as-is.

blueyed:
It currently adds a timestamp for level=3: https://github.com/neomake/neomake/blob/eb1f7d495208b22680d1ef769c02929e29a94c30/autoload/neomake/utils.vim#L61-L64.

Thanks for your review!

blueyed:
@tweekmonster 
Maybe I missed your point regarding https://github.com/neomake/neomake/pull/976#issuecomment-281826908: do you mean to add a log message that logs info about the increased verbosity, or (as I've understood it) to prefix log messages with their respective level?

tweekmonster:
> add a log message that logs info about the increased verbosity

This.

PR #977, opened 2017-01-29
[RFC] Entry removal in post-processing

tweekmonster:
This allows entries to be explicitly removed by a `postprocess` function.

In the future, I want to add support for a user-defined `NeomakePostProcess(maker, entry)` function that's called after the maker's `postprocess` function.  This will reduce the cognitive load and the need for maintenance for an end user that just wants to alter/remove some entries without worrying about dealing with a built-in maker directly.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -674,8 +674,8 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
             endif
         endif
 
-        if !entry.valid
-            if maker.remove_invalid_entries
+        if entry.valid <= 0
+            if entry.valid < 0 || maker.remove_invalid_entries
                 let index -= 1
                 call remove(list, index)
                 let list_modified = 1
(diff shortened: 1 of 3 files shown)


blueyed:
Thanks!

`NeomakePostProcess` makes sense for me, too.
Could/should we do it through a hook / user command?

tweekmonster:
> Could/should we do it through a hook / user command?

Using autocmd would work, too.  In that case, it would be better to pass the entire list to reduce the number of post-processing calls if there is more than one subscriber.  But, that would mean that the list should be modified by the subscribers.  I don't know if that qualifies as "easy" from a user perspective, though.

blueyed:
> But, that would mean that the list should be modified by the subscribers. I don't know if that qualifies as "easy" from a user perspective, though.

I am not sure either, but it seems like using autocommands for hooks is somehow a well-defined interface.  OTOH passing/setting the context as global vars is a bit silly.

Let's go with this PR as-is for now?

tweekmonster:
> OTOH passing/setting the context as global vars is a bit silly.

No more silly than having `v:whatever` 😜

> Let's go with this PR as-is for now?

👍

PR #979, opened 2017-01-30
Add maker.name to neomake#ListJobs() output

sanmiguel:
As per the title: when using `:NeomakeListJobs` to manage errant jobs, it can be hard to figure out which one was which. This proposed change simply adds the `maker.name` value on the output for each job.


blueyed:
Thanks!
Makes a lot of sense.
A test would be nice, but since it broke no existing one, you're good.. ;)

blueyed:
Thanks, merged.

sanmiguel:
Awesome, thanks!

PR #980, opened 2017-01-30
Improve handling of maker.exe

blueyed:
Created based on https://github.com/neomake/neomake/pull/967.

TODO:
 - [x] drop neomake#utils#MakerIsAvailable / neomake#utils#AvailableMakers ?!

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -542,30 +542,31 @@ function! s:Make(options) abort
     let options = copy(a:options)
     call extend(options, {
                 \ 'file_mode': 0,
-                \ 'enabled_makers': [],
                 \ 'bufnr': bufnr('%'),
                 \ 'ft': '',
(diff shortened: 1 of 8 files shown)


autoload/neomake.vim
```diff
@@ -142,6 +144,23 @@ function! s:MakeJob(make_id, options) abort
...
+            let error = printf('Exe (%s) of maker %s is not executable.', maker.exe, maker.name)
+            if !has_key(s:exe_error_thrown, maker.exe)
+                let s:exe_error_thrown[maker.exe] = 1
+                throw 'Neomake: '.error
+            endif
+            call neomake#utils#DebugMessage(error)
+            return -1
+        endif
+
+        " XXX: mark it as to be skipped earlier?!
```
blueyed: The common use case is that those would be skipped always, so we should short-circuit it.

PR #991, opened 2017-02-03
Expand all % in arguments

### LINKED ISSUE #990: neomake#util#ExpandArgs requires the metacharacters to be at the start of the string
I'm unable to define an `args` that has the cmdline-special metacharacters at an arbitrary position in the string.  Specifically, I was trying to use

```vim
let g:neomake_c_lint_maker = {
    \ 'exe': 'make',
    \ 'args': ['clint', 'LINT_FILE=%'],
    \ 'append_file': 0,
    \ 'errorformat': '%f:%l: %m',
    \}
```

to perform linting when working on Neovim.

### Expected behavior

I was hoping for `:Neomake` to successfully lint the current file.  However, `:Neomake` ends up running the command `make clint LINT_FILE=%` instead of `make clint LINT_FILE=src/nvim/message.c`.

tweekmonster:
Fixes #990

I skipped fixing one of the tests since I want to wait on a consensus on this.

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -189,8 +189,7 @@ function! neomake#utils#MakerFromCommand(command) abort
     " XXX: use neomake#utils#ExpandArgs and/or remove it.
     "      Expansion should happen later already!
     " NOTE: useful when calling it from cmdline..
-    let command = substitute(a:command, '%\(:[a-z]\)*',
-                           \ '\=expand(submatch(0))', 'g')
+    let command = neomake#utils#ExpandArgs([a:command])[0]
     " Create a maker object, with a "fn" callback.
     let maker = copy(s:command_maker)
     let maker.__command = command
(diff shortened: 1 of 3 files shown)


tweekmonster:
Ping @blueyed @jamessan

One of the test items stood out as weird: `%:rt`.  Since `%` is expanded to `file1.ext`, only `:r` has an effect.  `expand('%:rt')` returns `file1.ext` and ignores the rest of the string.  Is this the intended outcome or should it become `file1.extt`?

@blueyed mentioned in #990 
> Or rather not at all, but let it get handled later on.

I totally glossed over this part.  Where should it be handled?

blueyed:
> Since % is expanded to file1.ext, only :r has an effect.
> `expand('%:rt')` returns file1.ext

`:!echo %:prt` would return `/home/user/src/neomake/file1.extrt`, i.e. with `rt` at the end.

blueyed:
@tweekmonster 
> I totally glossed over this part. Where should it be handled?

I think https://github.com/neomake/neomake/blob/bbef95f6799b1f69fbea224d102d39ba4e2cfbd3/autoload/neomake/utils.vim#L192-L193 can be skipped, and the expansion would be done later in the central place.

autoload/neomake/utils.vim
```diff
@@ -189,8 +189,7 @@ function! neomake#utils#MakerFromCommand(command) abort
     " XXX: use neomake#utils#ExpandArgs and/or remove it.
     "      Expansion should happen later already!
     " NOTE: useful when calling it from cmdline..
-    let command = substitute(a:command, '%\(:[a-z]\)*',
-                           \ '\=expand(submatch(0))', 'g')
+    let command = neomake#utils#ExpandArgs([a:command])[0]
```
blueyed: Ok, makes sense like this here I assume, since it won't be expanded later.
Should remove the `XXX`/`NOTE` here though.

tweekmonster:
@blueyed I think this is RDY.  Just want to get your final thoughts in case I didn't cover everything.

blueyed:
Thanks a lot again! 🎆

PR #993, opened 2017-02-03
Added MSBuild maker

augustomelo:
The MSBuild uses the solution file (.sln) or the .csproj to build, so to the maker works it was needed to add a function to find the .sln file. To this maker works as well it is needed to add the MSBuild on the
environmental variable PATH.

### DIFF
--- a/autoload/neomake/makers/ft/cs.vim
+++ b/autoload/neomake/makers/ft/cs.vim
@@ -1,12 +1,22 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#ft#cs#EnabledMakers()
+function! neomake#makers#ft#cs#EnabledMakers() abort
     return ['mcs']
 endfunction
 
-function! neomake#makers#ft#cs#mcs()
+function! neomake#makers#ft#cs#mcs() abort
     return {
         \ 'args': ['--parse', '--unsafe'],
         \ 'errorformat': '%f(%l\,%c): %trror %m',
+        \ }
+endfunction
+
+function! neomake#makers#ft#cs#msbuild() abort
+    return {
+        \ 'exe': 'msbuild',
+        \ 'args': ['-nologo', '-v:q', '-property:GenerateFullPaths=true', neomake#utils#FindGlobFile(expand('%:p:h'), '*.sln')],
+        \ 'errorformat': '%E%f(%l\,%c): error CS%n: %m [%.%#],'.
+        \                '%W%f(%l\,%c): warning CS%n: %m [%.%#]',
+        \ 'append_file' : 0,
         \ }
 endfunction
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -423,3 +423,22 @@ endfunction
 function! neomake#utils#path_sep() abort
     return neomake#utils#IsRunningWindows() ? ';' : ':'
 endfunction
+
+" Find a file by going up the directories from the start directory
+" and performing glob search for the file.
+function! neomake#utils#FindGlobFile(startDir, file) abort
+    let currDir = a:startDir
+    let fileFound = ''
+
+    while empty(fileFound)
+        let fileFound = globpath(currDir, a:file, 1)
+        let lastFolder = currDir
+        let currDir = fnamemodify(currDir, ':h')
+
+        if currDir ==# lastFolder
+            break
+        endif
+    endwhile
+
+    return fileFound
+endfunction
--- a/tests/ft_cs.vader
+++ b/tests/ft_cs.vader
@@ -1,0 +1,30 @@
+Include: include/setup.vader
+
+Execute (cs: msbuild: errorformat):
+  Save &errorformat
+  let &errorformat = neomake#makers#ft#cs#msbuild().errorformat
+  file FooBar.cs
+
+  lgetexpr "FooBar.cs(21,63): error CS1002: ; expected [Foo\Foobar.csproj]"
+  AssertEqual getloclist(0), [{
+    \ 'lnum': 21,
+    \ 'bufnr': bufnr('%'),
+    \ 'col': 63,
+    \ 'valid': 1,
+    \ 'vcol': 0,
+    \ 'nr': 1002,
+    \ 'type': 'E',
+    \ 'pattern': '',
+    \ 'text': "; expected"}]
+
+  lgetexpr "FooBar.cs(25,29): warning CS0168: The variable 'ex' is declared but never used [Foo\Foobar.csproj]"
+  AssertEqual getloclist(0), [{
+    \ 'lnum': 25,
+    \ 'bufnr': bufnr('%'),
+    \ 'col': 29,
+    \ 'valid': 1,
+    \ 'vcol': 0,
+    \ 'nr': 168,
+    \ 'type': 'W',
+    \ 'pattern': '',
+    \ 'text': "The variable 'ex' is declared but never used"}]
--- a/tests/neomake.vader
+++ b/tests/neomake.vader
@@ -26,6 +26,7 @@ Include (Shell): ft_sh.vader
 Include (Text): ft_text.vader
 Include (Elixir): ft_elixir.vader
 Include (PHP): ft_php.vader
+Include (Cs): ft_cs.vader
 
 * Old/unorganized tests
 Execute (neomake#GetMakers):


autoload/neomake/makers/ft/cs.vim
```diff
@@ -10,3 +10,34 @@ function! neomake#makers#ft#cs#mcs()
...
+    return {
+        \ 'exe': 'msbuild',
+        \ 'args': [ '-nologo', '-v:q', '-property:GenerateFullPaths=true', s:FindSLN() ],
+        \ 'errorformat': '\ %#%f(%l\\\,%c):\ %m',
+        \ 'append_file' : 0,
+        \ }
+endfunction
+
+" FUNCTION find *.sln file going up directory
+function! s:FindSLN()
```
blueyed: We should have a generic utils function for this, probably based on what Syntastic has already.

autoload/neomake/utils.vim
```diff
@@ -409,3 +409,22 @@ endfunction
...
+        let lastFolder = currDir
+        let currDir = fnamemodify(currDir, ':h')
+
+        if currDir ==# lastFolder
+            break
+        endif
+    endwhile
+
+    return fileFound
+endfunction
```
blueyed: What about https://github.com/vim-syntastic/syntastic/blob/2124af80abfb856e8806c3b508b3982b7157c4a3/autoload/syntastic/util.vim#L251-L259 instead?

augustomelo: I think findfile function does not support wildcards, and I need to find any file that ends with *.sln

blueyed: Ah, I see.
Then https://github.com/vim-syntastic/syntastic/blob/2124af80abfb856e8806c3b508b3982b7157c4a3/autoload/syntastic/util.vim#L261-L297..

augustomelo: Yeah I based on this one, it does a bunch of stuff that I do not know if they are needed, I tried to keep it simple, but you are the maintainer what you think ?

blueyed: It's fine when it's based on it.. I think Syntastic covers a lot of corner cases already, and we are good to pick that / build upon it.
Having it in a separate utils function is good already, and allows to improve it from there - so if you think it's fine, I'm OK.

autoload/neomake/makers/ft/cs.vim
```diff
@@ -10,3 +10,12 @@ function! neomake#makers#ft#cs#mcs()
         \ 'errorformat': '%f(%l\,%c): %trror %m',
         \ }
 endfunction
+
+function! neomake#makers#ft#cs#msbuild()
+    return {
+        \ 'exe': 'msbuild',
+        \ 'args': [ '-nologo', '-v:q', '-property:GenerateFullPaths=true', neomake#utils#FindGlobFile(expand('%:p:h'), '*.sln') ],
```
blueyed: I would remove the spaces after the `[` and before `]`.

autoload/neomake/makers/ft/cs.vim
```diff
@@ -10,3 +10,12 @@ function! neomake#makers#ft#cs#mcs()
...
+        \ 'errorformat': '\ %#%f(%l\\\,%c):\ %m',
```
blueyed: No need to escape the space here, i.e. replace `\ ` with just ` `.
Not sure about the `\\\`?!

autoload/neomake/utils.vim
```diff
@@ -409,3 +409,22 @@ endfunction
 function! neomake#utils#path_sep() abort
     return neomake#utils#IsRunningWindows() ? ';' : ':'
 endfunction
+
+" Find a file by going up the directories form the start directory
```
blueyed: s:/form/from/

blueyed: Add a reference to the Syntastic function it's based on?!
https://github.com/vim-syntastic/syntastic/blob/2124af80abfb856e8806c3b508b3982b7157c4a3/autoload/syntastic/util.vim#L261-L297.

autoload/neomake/utils.vim
```diff
@@ -409,3 +409,22 @@ endfunction
...
+" and performing glob search for the file.
+function! neomake#utils#FindGlobFile(startDir, file)
+    let currDir = a:startDir
+    let fileFound = ""
```
blueyed: Single quotes should be used/preferred.. I wonder why CI did not complain.

autoload/neomake/utils.vim
```diff
@@ -409,3 +409,22 @@ endfunction
...
+
+    while empty(fileFound)
+        let fileFound = globpath(currDir , a:file, 1)
```
blueyed: Extra space here after `currDir`, please remove it.

autoload/neomake/makers/ft/cs.vim
augustomelo: I am not sure either lol, If I still remember o took that form here: 
http://stackoverflow.com/questions/101270/vim-errorformat-for-visual-studio

augustomelo:
I made all the changes that you asked for, besides the ```\\\``` I do know why is there but if I remove it, it will stop working, maybe you can help me understand it, next you can see an error form the msbuild:
```
FooBar.cs(21,63): error CS1002: ; expected [Foo\Foobar.csproj]
```
As you can see it does not make sense the ```\\\``` on the error format, the error format should this one
```
'errorformat': '\ %#%f(%l,%c):\ %m',
```
Do you agree?

blueyed:
Yeah, the `\\\` seems to make sense there.

blueyed:
I've pushed a fixup to test/debug a test failure.. I'll take over regarding this PR - thanks for bearing with us/me.. :)

augustomelo:
I do not want to bother you, but I just did not understand why the ```\\\``` makes sense,  I have read the vim documentation and says that you have to escape the comma, but it should something like ``\,`` right ?

blueyed:
Sorry, I thought we need to escape the backslash here..

According to https://github.com/neomake/neomake/pull/993#issuecomment-277802091 it might not be necessary, too.

blueyed:
Let me debug the Travis issue (which is about not calling vint correctly), and then we can rebase it on this.

blueyed:
@augustomelo 
Any chance you could help us with https://github.com/neomake/neomake/issues/724 btw?

blueyed:
Improved the errorformat a bit, but there is room left: everything is an error now, and the "error" is let in the text.  Also you could get the error number etc.
I do not have sample output (but added yours to the test, which you could add additional ones to).

blueyed:
@augustomelo 
Would suggest to improve the errorformat through the tests.
See https://github.com/neomake/neomake/issues/724 in case tests do not run for you (assuming you're using Windows here).  But you can always use Travis for this, too - (maybe only running one test env then, by commenting/removing others from .travis.yml).

augustomelo:
@blueyed  I really appreciate your help! I will take a look when I have some time

blueyed:
Cool!
Hopefully during the weekend.. ;)

augustomelo:
Hey @blueyed sorry for the delay, finally I had some time to improve the error format, I hope that the test for the warning message is correct.  Should I make more test cases?

tests/ft_cs.vader
```diff
@@ -12,7 +12,19 @@ Execute (cs: msbuild: errorformat):
...
+  AssertEqual getloclist(0), [{
+    \ 'lnum': 25,
+    \ 'bufnr': bufnr('%'),
+    \ 'col': 29,
+    \ 'valid': 1,
+    \ 'vcol': 0,
+    \ 'nr': 168,
+    \ 'type': 'W',
+    \ 'pattern': '',
+    \ 'text': 'The variable 'ex' is declared but never used'}]
```
blueyed: The test fails:

> [{'lnum': 21, 'bufnr': 3, 'col': 63, 'valid': 1, 'vcol': 0, 'nr': 1002, 'type': 'E', 'pattern': '', 'text': 'error CS1002: ; expected'}] should be equal to [{'lnum': 21, 'bufnr': 3, 'col': 63, 'valid': 1, 'vcol': 0, 'nr': 1002, 'type': 'E', 'pattern': '^\VFoo\Foobar.csproj\$', 'text': '; expected'}]

Likely just because you got the quoting wrong (not escaped).. both in `lgetexpr` and `'text'`.

tests/ft_cs.vader
```diff
@@ -0,0 +1,30 @@
...
+    \ 'vcol': 0,
+    \ 'nr': 1002,
+    \ 'type': 'E',
+    \ 'pattern': '',
+    \ 'text': "; expected"}]
+
+  lgetexpr "Program.cs(25,29): warning CS0168: The variable 'ex' is declared but never used [Foo\Foobar.csproj]"
+  AssertEqual getloclist(0), [{
+    \ 'lnum': 25,
+    \ 'bufnr': bufnr('%'),
```
augustomelo: @blueyed this test is failing because of this line as you can see on the output from the Travis, the '''bufnr('%')''' should grab the buff number right? Maybe it was because I wrote only one test case, should I split it?
'''
[EXECUTE] (X) [{
'lnum': 25,
'bufnr': 3,
'col': 29,
'valid': 1,
'vcol': 0,
'nr': 168,
'type': 'W',
'pattern': '',
'text': 'The variable ''ex'' is declared but never used'}] 

should be equal to [{
'lnum': 25,
'bufnr': 41,
'col': 29,
'valid': 1,
'vcol': 0,
'nr': 168,
'type': 'W',
'pattern': '',
'text': 'The variable ''ex'' is declared but never used'}]
'''

blueyed: Fixed it.  It was not using the buffer's name.

blueyed:
@augustomelo 
Is this fine with you now?

Would be good to get this squashed (one commit, or two for the utils function).
Let me know if you want me to do it.

As for the fixups, they can be handled semi-automatically with `git rebase -i --autosquash origin/master`.

augustomelo:
Hey @blueyed  I could do it but only by Wednesday, if you could do it I will appreciate it  :)

blueyed:
Cool, here we go.  Will merge it when CI passed.

blueyed:
Please add it to the wiki when you find the time - with a link to the linter itself, and maybe instructions etc.
Thanks again!

Issue #972, opened 2017-01-27
Provide some way of specifying compiler flags


tbodt:
There is currently no way to specify custom include paths or defines with the C and C++ makers, which is necessary to make my project build properly.

clang_complete uses a file in the project root called `.clang_complete`, maybe could use that?

tweekmonster:
There's a conversation we had about something like this here: https://github.com/tweekmonster/deoplete-clang2/commit/872ff514168a719d33b8e588bc64962fa75100dd#commitcomment-20225451

I apparently didn't give feedback that was requested there, but the conversation can continue here.

blueyed:
`maker.fn` is in by now and gets the `jobinfo` as an argument.

tbodt:
As a user, how would I go about making this happen using `maker.fn`?

tbodt:
(I really need this and deoplete-clang2 fucks up my insert-mode tab mapping.)

blueyed:
It is not documented currently on purpose, and I am not sure what's the way to go there.

You can always configure `..._args`, but then need to include the defaults also.
There is some discussion in https://github.com/neomake/neomake/pull/954#discussion_r97237315.
We can discuss it here further..

codeinabox:
As discussed neomaker_FT_MAKER_args_post is probably a good way to do it consistently on the basis it's self explanatory and the approach Syntastic adopted

tbodt:
Is `args_post` implemented yet?

blueyed:
@tbodt 
No.
As for now I would recommend using https://github.com/embear/vim-localvimrc or something similar to configure `_args` for your project(s).

blueyed:
See https://github.com/neomake/neomake/pull/2161.

Issue #843, opened 2016-12-14
Use cargo-check in cargo maker


fuine:
Cargo just got `check` command merged, which is what we wanted. It's essentially like `-Zno-trans`, so it should work significantly faster for the syntax checks. I will open the PR changing the maker after nightly version has been released, so that people who use nightly compiler could get it running now. For the actual merge i think that we should wait until this feature hits stable, and it's supposed to be introduced in 1.15, so in approximately 7 weeks. For more information please refer to [this PR](https://github.com/rust-lang/cargo/pull/3296).

hpurmann:
Would love to see this 👍  Is there a PR I can check out, @fuine?

fuine:
@hpurmann sorry, I haven't got enough time lately. Currently i can see that there is an ongoing WIP PR to use the json output: #942. What's more is that the `check` command has not been introduced in the 1.15 version, but is currently in beta, so we need to wait 7 weeks for its stabilization. After #942 gets resolved then I will open said PR.

hpurmann:
Ok great, looking forward to that. Thank you for your work!

jonhoo:
We'll probably want to monitor https://github.com/rust-lang/cargo/issues/3431 for when we'll be able to also check tests. I know for me at least that's a very important feature.

gyscos:
`cargo-check` is now in rust stable. Tests are still not compiled, though.

fuine:
I'm going to take a closer look at this issue over the weekend, and hopefully open a new PR.

fuine:
It looks like there are two things that are relevant to this issue:
* should neomake opt into using `cargo-check` without test support, or should it stick to the current manner of calling `cargo test`? `cargo-check` is significantly faster, but on the other hand some people might want to compile their tests as well.
* I wouldn't like to create conflicts with #1123/#942. @blueyed @JelteF can you provide a rough estimation of when that PR could possibly land? To be fair the only change would modify one line, but I don't want to cause more troubles, seeing as that PR consists of quite a lot of commits.

JelteF:
With regards to your second point #924 is mergeble now as far as I can tell
and works great. The #1123 PR requires more work, especially rewriting of
the tests.

The best way for this issue to be fixed right now in my opinion is by
giving the user the choice between a cargo full and a cargo check maker.
This can simply be done by copying the cargo maker definition and changing
the command. The only thing I'm not sure about is which one to have as the
default. I'd vote for check as the speed really makes it much nicer.

[...]

PR #1003, opened 2017-02-07
Standardised PHP makers to use convention of post args

codeinabox:
This revisits my earlier PR #954 but adopts standard of post args as discussed in #972. Happy to update my PR to include some notes in the documentation about using the convention of post args.

### DIFF
--- a/autoload/neomake/makers/ft/php.vim
+++ b/autoload/neomake/makers/ft/php.vim
@@ -21,9 +21,8 @@ endfunction
 function! neomake#makers#ft#php#phpcs() abort
     let l:args = ['--report=csv']
 
-    "Add standard argument if one is set.
-    if exists('g:neomake_php_phpcs_args_standard')
-        call add(l:args, '--standard=' . expand(g:neomake_php_phpcs_args_standard))
+    if exists('g:neomake_php_phpcs_args_post')
+        call add(l:args, , g:neomake_php_phpmd_args_post))
     endif
 
     return {
@@ -35,6 +34,14 @@ function! neomake#makers#ft#php#phpcs() abort
 endfunction
 
 function! neomake#makers#ft#php#phpmd() abort
+    let l:args = ['%:p', 'text']
+
+    if exists('g:neomake_php_phpmd_args_post')
+        call add(l:args, g:neomake_php_phpmd_args_post)
+    else
+        call add(l:args, 'codesize,design,unusedcode,naming')
+    endif
+
     return {
         \ 'args': ['%:p', 'text', 'codesize,design,unusedcode,naming'],
         \ 'errorformat': '%W%f:%l%\s%\s%#%m'


autoload/neomake/makers/ft/php.vim
```diff
@@ -27,9 +27,8 @@ endfunction
 function! neomake#makers#ft#php#phpcs() abort
     let l:args = ['--report=csv']
 
-    "Add standard argument if one is set.
-    if exists('g:neomake_php_phpcs_args_standard')
-        call add(l:args, '--standard=' . expand(g:neomake_php_phpcs_args_standard))
+    if exists('g:neomake_php_phpcs_args_post')
+        call add(l:args, , g:neomake_php_phpmd_args_post))
```
blueyed: > ./autoload/neomake/makers/ft/php.vim:31:26:Error: EVP_0: unexpected token: ,

Should be `call add(l:args, g:neomake_php_phpmd_args_post)`

But then it should also take into account buffer-local settings.
This could be done with `neomake#utils#GetSetting`, i.e. `call extend(l:args, neomake#utils#GetSetting('args_post', {}, [], ['php'], bufnr('%')))`.

Similar below.

codeinabox: Thanks. I noticed there are no Vader tests for PHP, should I add some to test this?

blueyed: @codeinabox 
That would be nice, yes.

codeinabox: The shell maker tests with their argument validation should give me a rough guide to testing.

Been thinking about this more and looking back on the way Syntastic handles it (see below) and now I am rethinking the approach.

```viml
let g:syntastic_php_phpcs_args = "--standard=PSR2"
let g:syntastic_php_phpmd_post_args = "cleancode,codesize,controversial,design,naming,unusedcode"
```
Would a solution be that the user can put anything into args but the maker adds or overrides as necessary in order for it to work with the error format?

blueyed: That would require to use the `fn` constructor callback curently.
Related: https://github.com/neomake/neomake/issues/843#issuecomment-291951164.

blueyed:
@codeinabox 
ping

What is the status of this?

blueyed:
@codeinabox 
Sorry for the late reply: there is no standard yet.
Feel free to roll your own for now I'd say (but better to avoid if possible).

codeinabox:
@blueyed are you thinking that arguments should be merged with certain defaults or should they replace all arguments? Some default arguments that are needed for the error format make sense

Syntastic for example has separates out arguments required for errorformat and default ruleset, so you can override just the latter eg https://github.com/vim-syntastic/syntastic/blob/master/syntax_checkers/php/phpmd.vim

blueyed:
@codeinabox 
Yes, I think it makes sense to have something like "base args", and others get appended to it.
The appended ones then could still override the existing ones, e.g. with `--foo` in base, and then `--no-foo` appended (in case programs handle that).

blueyed:
Looked at this again.
Currently you can do this using the InitForJob callback, which is not as convenient as specifying `…_append_args` or similar.
And since I plan to have something like `:NeomakeRun flake --select X` it makes sense to have a generic setting/handling for this.
See https://github.com/neomake/neomake/pull/2161.
I am not sure if a setting is makes sense (mostly because it adds overhead, but that can always be optimized of course).
What do you think?

blueyed:
Closing in favor of https://github.com/neomake/neomake/pull/2161 as per previous comment.

PR #1006, opened 2017-02-08
Tests for PHP makers

codeinabox:


codeinabox:
Understood. What is your preference on checking the whole loclist dictionary as opposed to just some properties like this?

```viml

  AssertEqual len(getloclist(0)), 2
  AssertEqual getloclist(0)[0].lnum, 16
  AssertEqual getloclist(0)[0].text, "    Avoid unused parameters such as '$request'."
  AssertEqual getloclist(0)[1].lnum, 48
  AssertEqual getloclist(0)[1].text, "    Avoid variables with short names like $id. Configured minimumlength is 3."
```

I am trying the approach you suggested but the below code example doesn't work as it seems the bufnr value does not match up

```viml
  AssertEqual getloclist(0), [
  \ {'lnum': 28, 'bufnr': bufnr('%'), 'col': 6, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'e', 'pattern': '',
  \  'text': 'Missing @return tag in functioncomment'},
  \ {'lnum': 36, 'bufnr': bufnr('%'), 'col': 5, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'e', 'pattern': '',
  \  'text': 'Missing short description in doc comment'}]
```

>     (1/2) [EXECUTE] phpcs
>     (1/2) [EXECUTE] (X) [{'lnum': 28, 'bufnr': 33, 'col': 6, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'e', 'pattern': '', 'text': 'Missing @return tag in functioncomment'}, {'lnum': 36, 'bufnr': 33, 'col': 5, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'e', 'pattern': '', 'text': 'Missing short description in doc comment'}] should be equal to [{'lnum': 28, 'bufnr': 12, 'col': 6, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'e', 'pattern': '', 'text': 'Missing @return tag in functioncomment'}, {'lnum': 36, 'bufnr': 12, 'col': 5, 'valid': 1, 'vcol': 0, 'nr': -1, 'type': 'e', 'pattern': '', 'text': 'Missing short description in doc comment'}]
>     (2/2) [EXECUTE] phpmd
>   Success/Total: 1/2
>

codeinabox:
@blueyed just checking which assertion approach you prefer

blueyed:
@codeinabox 
The former is easier to see what is wrong, but the latter is more complete.
So it really depends..  when you look at the other `ft_*` tests it usually compares whole entries.

To make the `bufnr` work you usually have to use `file file1.php` and then use `file1.php` in the linter output (so that Vim matches it to the current buffer).

codeinabox:
Have updated my PR.

tests/ft_php.vader
```diff
@@ -0,0 +1,38 @@
...
+
+  Save &errorformat
+  let &errorformat = neomake#makers#ft#php#phpmd().errorformat
+  lgetexpr phpmd_output
+
+  AssertEqual len(getloclist(0)), 2
+  AssertEqual getloclist(0)[0].lnum, 16
+  AssertEqual getloclist(0)[0].text, "    Avoid unused parameters such as '$request'."
+  AssertEqual getloclist(0)[1].lnum, 48
+  AssertEqual getloclist(0)[1].text, "    Avoid variables with short names like $id. Configured minimumlength is 3."
```
blueyed: Shouldn't the whitespace at the beginning get removed here?

blueyed:
Thanks, pushed some fixups.

I've removed the tests for args, since I assumed they are not dynamic, but with `neomake#makers#ft#php#phpcs` they are, so it might make sense there.  What do you think?

Can you look at https://github.com/neomake/neomake/pull/1025 also, please?

codeinabox:
I am happy for you to remove the args tests. Would you like me to add some tests for `php` before picking up #1025?

blueyed:
> Would you like me to add some tests for php before picking up #1025?

Would be nice to have, yes!

tests/ft_php.vader
```diff
@@ -1,5 +1,18 @@
...
+  \"Fatal error: Call to undefined function hello() in /app/src/hello.php on line 1"
+
+  Save &errorformat
+  let &errorformat = neomake#makers#ft#php#php().errorformat
+  lgetexpr php_output
+
+  let loclist = getloclist(0)
+  AssertEqual len(loclist), 1
+  AssertEqual loclist[0].lnum, 1
+  AssertEqual loclist[0].text, 'Call to undefined function hello()'
```
codeinabox: @blueyed as promised here are the tests for php maker

blueyed:
Thanks, rebased.
Will merge it once it's green.

PR #1014, opened 2017-02-11
Support new compilation warnings format with mix

aerosol:
Hi, semi-related to https://github.com/elixir-lang/elixir/issues/5143

`mix compile --warnings-as-errors` returns a multiline string now:

```
/tmp/foo $ mix compile --warnings-as-errors
Compiling 1 file (.ex)
warning: variable world is unused
  lib/foo.ex:3
```

This PR fixes the `mix` maker, so both warnings and compilation errors are detected properly.


aerosol:
cc @geowa4

blueyed:
Thanks!

PR #1020, opened 2017-02-13
Fix incorrect lookup for default highlight color

davesque:
Not sure if this was intentional, but this was causing error signs and markers to be almost invisible in my color scheme.  I suspect this would also be the case for many other color schemes.

### DIFF
--- a/autoload/neomake/highlights.vim
+++ b/autoload/neomake/highlights.vim
@@ -96,7 +96,7 @@ endfunction
 
 function! neomake#highlights#DefineHighlights() abort
     for [group, fg_from] in items({
-                \ 'NeomakeError': ['Error', 'bg'],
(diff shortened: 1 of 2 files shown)


blueyed:
Does it not work for signs with you really?
There is some workaround/hack for it already: https://github.com/neomake/neomake/blob/54ac82257d3cad799c39b272d3e3fe926c201e0b/autoload/neomake/signs.vim#L231-L237.
(via https://github.com/neomake/neomake/pull/659)

Which theme are you using?

davesque:
I am using the hybrid color scheme.

https://github.com/w0ng/vim-hybrid

blueyed:
I am closing this given the existing tests / workaround.
I've not investigated by looking into the specific colorscheme.

Please re-open in case there's a way to handle the current state and this in a sane way.

PR #1023, opened 2017-02-14
New g:neomake_always_show_list option suggestion

rmolin88:
In my case I always want neomake to show what happen. Sometimes my makers will have errors that are not recognized by the 'errorformat' and I am just left hanging and having to open the location list manually.
Thanks for a great plugin

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -498,9 +498,17 @@ function! s:HandleLoclistQflistDisplay(file_mode) abort
         let height = get(g:, 'neomake_list_height', 10)
         let win_val = winnr()
         if a:file_mode
-            exe 'lwindow' height
+            if open_val == 3
+                exe 'lopen' height
+            else
+                exe 'lwindow' height
+            endif
         else
-            exe 'cwindow' height
+            if open_val == 3
+                exe 'copen' height
+            else
+                exe 'cwindow' height
+            endif
         endif
         if open_val == 2 && win_val != winnr()
             wincmd p
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -496,15 +496,16 @@ function! s:HandleLoclistQflistDisplay(file_mode) abort
     let open_val = get(g:, 'neomake_open_list')
     if open_val
         let height = get(g:, 'neomake_list_height', 10)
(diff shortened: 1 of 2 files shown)


rmolin88:
I guess it should say in the documentation that g:neomake_always_show_list depends on g:neomake_open_list being enabled

blueyed:
Thanks!
It could automatically enable `neomake_open_list = 2` then maybe?  (nobody uses 1 for this setting probably)

Personally I am using some custom method to only open it in case of errors, which I'd like to clean up and submit - and therefore an additional setting might get in the way then later.
See https://github.com/blueyed/dotfiles/blob/49cbd9c5f8c3bf6c255e7d61650a711a5bcde864/vim/plugin/statusline.vim#L269-L325.

Maybe you can add a hook yourself to open the list always?

```vim
let g:neomake_open_list = 0  " default
augroup my_neomake
  au!
  autocmd User NeomakeCountsChanged,NeomakeFinished call MyOpenWindow()
augroup END
```

You can use `g:neomake_hook_context.file_mode ? 'lopen' : 'copen'` there, but need to handle going back to the prev window then.

What do you think?

rmolin88:
Your suggestion worked very well.

Thanks!!!

blueyed:
Nice!
Yay for hooks!

What are you using now?

rmolin88:
Indeed. Great choice on implementing them 👍 . They give your plugin a lot of flexibility. I honestly hadnt notice.
This is my configuration:
```vim
augroup custom_neomake
        autocmd!
	autocmd User NeomakeFinished call utils#NeomakeOpenWindow()
augroup END

function! utils#NeomakeOpenWindow() abort
	let height = get(g:, 'neomake_list_height', 10)
	exe g:neomake_hook_context.file_mode ? 'lopen' . height : 'copen' . height
	wincmd p
endfunction
```

blueyed:
Great!
Let's close this issue then?

I could imagine adding this to the doc or README, e.g. in a new Recipes section or the already existing FAQs.  Could also be added to the hooks section (`:h neomake-autocmds`).
What do you think?
Do you think it makes sense

rmolin88:
I was about to suggest that the sample code or some similar example be added to the documentation.

Thanks again for a great plugin and listening to suggestions 👍

PR #1025, opened 2017-02-17
php: use %E in errorformat instead of postprocess

blueyed:


codeinabox:
Now that #1006 is merged in and the tests for this refactor are passing, are any other changes required?

tests/ft_php.vader
```diff
@@ -44,5 +47,7 @@ Execute (phpmd):
   AssertEqual len(loclist), 2
   AssertEqual loclist[0].lnum, 16
   AssertEqual loclist[0].text, "Avoid unused parameters such as '$request'."
+  AssertEqual loclist[0].type, 'E'
   AssertEqual loclist[1].lnum, 48
   AssertEqual loclist[1].text, "Avoid variables with short names like $id. Configured minimumlength is 3."
+  AssertEqual loclist[1].type, 'E'
```
blueyed: I think we should use type=warning for phpmd by default?!
Related: https://github.com/neomake/neomake/issues/584

codeinabox: I agree as they are code style warnings and not syntax errors.

codeinabox: Is this something we need to change in this PR or should I create another? Otherwise changes LGTM

blueyed: I am changing it here, since we have the test here.

PR #1026, opened 2017-02-17
PHPStan maker

codeinabox:
This adds support for [PHPStan](https://github.com/phpstan/phpstan)

### DIFF
--- a/tests/ft_php.vader
+++ b/tests/ft_php.vader
@@ -57,3 +57,19 @@ Execute (phpmd):
   AssertEqual loclist[1].type, 'W'
 
   bwipe /app/src/AppBundle/Controller/DefaultController.php
+
+Execute (phpstan):
+  let phpstan_output =
+  \"/app/src/hello.php:6:Function dsdsd not found"
+
+  Save &errorformat
+  let &errorformat = neomake#makers#ft#php#phpstan().errorformat
+  lgetexpr phpstan_output
+
+  let loclist = getloclist(0)
+  AssertEqual len(loclist), 1
+  AssertEqual loclist[0].lnum, 6
+  AssertEqual loclist[0].text, "Function dsdsd not found"
+  AssertEqual loclist[0].type, 'E'
+
+  bwipe /app/src/hello.php
--- a/autoload/neomake/makers/ft/php.vim
+++ b/autoload/neomake/makers/ft/php.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#php#EnabledMakers() abort
-    return ['php', 'phpmd', 'phpcs']
+    return ['php', 'phpmd', 'phpcs', 'phpstan']
 endfunction
 
 function! neomake#makers#ft#php#php() abort
@@ -38,5 +38,12 @@ function! neomake#makers#ft#php#phpmd() abort
     return {
         \ 'args': ['%:p', 'text', 'codesize,design,unusedcode,naming'],
         \ 'errorformat': '%W%f:%l%\s%\s%#%m'
+        \ }
+endfunction
+
+function! neomake#makers#ft#php#phpstan() abort
+    return {
+        \ 'args': ['analyse', '--errorFormat','raw'],
+        \ 'errorformat': '%E%f:%l:%m'
         \ }
 endfunction


codeinabox:
@blueyed I've given an example of the output in the tests. Should it be easy enough for me to write an error formatter for this?

tests/ft_php.vader
```diff
@@ -46,3 +46,25 @@ Execute (phpmd):
   AssertEqual loclist[0].text, "Avoid unused parameters such as '$request'."
   AssertEqual loclist[1].lnum, 48
   AssertEqual loclist[1].text, "Avoid variables with short names like $id. Configured minimumlength is 3."
+
+Execute (phpmd):
```
blueyed: Name is wrong.

tests/ft_php.vader
```diff
@@ -46,3 +46,25 @@ Execute (phpmd):
...
+  let phpstan_output =
+  \" ------ ----------------------------------------------------------------------------------------------\n
+  \  Line   TaskRepository.php\n
+  \ ------ ----------------------------------------------------------------------------------------------\n
+  \  25     Parameter $taskId of method TodoBundle\Repository\TaskRepository::findById() has invalid typehint type TodoBundle\Repository\Identifier.\n
+  \  27     Call to an undefined method TodoBundle\Repository\TaskRepository::findOneById().\n
+  \ ------ ----------------------------------------------------------------------------------------------\n
+  \\n
+  \ [ERROR] Found 2 errors"
```
blueyed: Is there an easier to parse output formatter maybe, where the file is displayed on each line?
Otherwise it might be possible still, see `:h errorformat-separate-filename`.

tests/ft_php.vader
```diff
@@ -46,3 +46,25 @@ Execute (phpmd):
...
+  Save &errorformat
+  let &errorformat = neomake#makers#ft#php#phpstan().errorformat
+  lgetexpr phpstan_output
+
+  let loclist = getloclist(0)
+  AssertEqual len(loclist), 2
+  AssertEqual loclist[0].lnum, 25
+  AssertEqual loclist[0].text, "Parameter $taskId of method TodoBundle\Repository\TaskRepository::findById() has invalid typehint type TodoBundle\Repository\Identifier."
+  AssertEqual loclist[1].lnum, 27
+  AssertEqual loclist[1].text, "Call to an undefined method TodoBundle\Repository\TaskRepository::findOneById()."
```
blueyed: Also check `type` here.

tests/ft_php.vader
codeinabox: Not yet but apparently there is talk of adding control of output

blueyed:
@codeinabox 
Ping.  Needs to be rebased by now also.

codeinabox:
Have rebased. Will try to find some time to implement an error formatter though probably worth waiting for https://github.com/phpstan/phpstan/pull/132

blueyed:
Link to CI failure(s): https://travis-ci.org/neomake/neomake/jobs/205362077#L1666

It makes sense to wait for some better output format indeed.

Is there JSON output already?  (since I am working to improve this currently)

codeinabox:
No JSON output is available. Agreed, I shall park this for now and pick it up when the error output change has been implemented.

Fuco1:
The PR was resolved and I'm working on the JSON formatter, you can expect it very soon!

codeinabox:
@Fuco1 great, is there a PR I can follow for the JSON formatter?

codeinabox:
I see there is now [custom error formatters](https://github.com/phpstan/phpstan#custom-error-formatters) so I will try to pick this up again

blueyed:
(retriggering a bulld)

codeinabox:
I've updated the tests and will try to finish the implementation this week

codeinabox:
@blueyed I've sorted out the error format but how do I get it work if phpstan is installed as part of a project's dependencies? I know for eslint there is https://github.com/benjie/neomake-local-eslint.vim

blueyed:
@codeinabox 
Why wouldn't you change your PATH dynamically then (to include e.g. `node_modules/.bin`)?  That's the case with Python's virtualenv's for example, and can be achieved with something like https://github.com/Tarrasch/zsh-autoenv in your shell.

Apart from that, we have a util function to look for some file/dir based on a glob.. would that help?

codeinabox:
Didn't consider dynamically updating the PATH. I guess the question is path resolution the responsibility of makers or is that a configuration issue?

blueyed:
Neomake or a maker should not update `$PATH` probably, but only use it.

As for `node_modules` and `eslint`, the following should do:
```vim
augroup my_neomake_setup
  au!
  au FileType javascript let nm = finddir('node_modules', ';') | if !empty(nm) | let exe = nm.'/.bin/eslint' | if executable(exe) | let b:neomake_javascript_eslint_exe = fnamemodify(exe, ':p') | endif | endif
augroup END
```

codeinabox:
I have now tested and this it's working as expected.

codeinabox:
This now code complete and tested against the latest version of phpstan which supports the raw error formatter.

blueyed:
Thanks, merged!

Do you think my last comment should go into the documentation/FAQ somehow?
Do you use a similar approach now?

codeinabox:
I opted just to add `./vendor/bin` to my shell PATH configuration but yes I think that would be a useful addition to the wiki. I've updated it with a list of the PHP makers.

PR #1030, opened 2017-02-21
Add generic postprocess to highlight named args

blueyed:
TODO:
 - [ ] doc

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -643,7 +643,11 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
     let maker_type = file_mode ? 'file' : 'project'
     let cleaned_signs = 0
     let ignored_signs = []
+    unlet! s:postprocess  " vim73
     let s:postprocess = get(maker, 'postprocess', function('neomake#utils#CompressWhitespace'))
+    if type(s:postprocess) == type({})
+        let s:postprocess = s:postprocess.fn
+    endif
     let debug = get(g:, 'neomake_verbose', 1) >= 3
 
     while index < len(list)
@@ -652,7 +656,21 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
         let index += 1
 
         let before = copy(entry)
-        call call(s:postprocess, [entry], maker)
+        if type(s:postprocess) == type([])
+            for s:f in s:postprocess
+                if type(s:f) == type({})
+                    let s:this = extend(copy(s:f), {'maker': maker})
+                    call call(s:f.fn, [entry], s:this)
+                else
+                    let s:this = {'maker': maker}
+                    call call(s:f, [entry], s:this)
+                endif
+                unlet s:f  " vim73
+            endfor
+        else
+            let s:this = {'maker': maker}
+            call call(s:postprocess, [entry], s:this)
+        endif
         if entry != before
             let list_modified = 1
             if debug
--- a/autoload/neomake/makers/ft/vim.vim
+++ b/autoload/neomake/makers/ft/vim.vim
@@ -14,8 +14,11 @@ function! neomake#makers#ft#vim#vint() abort
 
     return {
         \ 'args': l:args,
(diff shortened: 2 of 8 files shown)


autoload/neomake.vim
```diff
@@ -645,7 +648,22 @@ function! s:AddExprCallback(jobinfo, prev_index) abort
...
+            for s:f in s:postprocess
+                if type(s:f) == type({})
+                    let s:this = extend(copy(s:f), {'maker': maker})
+                    call call(s:f.fn, [entry], s:this)
+                else
+                    let s:this = {'maker': maker}
+                    call call(s:f, [entry], s:this)
+                endif
+            endfor
+            " TODO: abort on first invalid entry already?!
```
tweekmonster: What does this mean?

autoload/neomake/makers/ft/vim.vim
```diff
@@ -14,8 +14,11 @@ function! neomake#makers#ft#vim#vint() abort
 
     return {
         \ 'args': l:args,
-        \ 'errorformat': '%I%f:%l:%c:style_problem:%m,%f:%l:%c:%t%*[^:]:%m'
-        \ }
+        \ 'errorformat': '%I%f:%l:%c:style_problem:%m,%f:%l:%c:%t%*[^:]:%m',
+        \ 'postprocess': {
+        \   'fn': function('neomake#postprocess#GenericLengthPostprocess'),
+        \   'pattern': '\vUndefined variable: \zs([^ ]+)',
```
tweekmonster: I haven't explored this much, but would this be easier to maintain if it was also allowed to be a list of patterns?

blueyed: Might make sense.
Currently it is handled by the postprocess function itself, but we could make the default one accept a list, too.
OTOH, using a list for the postprocess handlers felt dirrty already; as for the pattern, it could be easily be a group/ORed pattern.

autoload/neomake.vim
blueyed: If one postprocess sets valid=--1 we might abort here already.
Bot OTOH another one might revert it?!

tweekmonster: Ah okay, I see.  Removing the entry here would require duplication, right?  I think it's fine to just let it fall through to the next set of branches since that's how it worked before and `remove_invalid_entries` might still apply without post-processors.

Since it's a list of processors now, passing through `entry != before` might help debug a processor that was affected by a prior processor.

autoload/neomake/makers/ft/vim.vim
tweekmonster: > OTOH, using a list for the postprocess handlers felt dirrty already; as for the pattern, it could be easily be a group/ORed pattern.

Yeah, adding a loop for it wouldn't be ideal either.  Complicated patterns could be a list and joined into a string for the maker config.

I was making the comment wrt maintenance since I've had to craft some pretty gnarly patterns and it's sometimes tough keeping track of all the parenthesis (especially if you're trying to actually match parenthesis).  Hopefully these patterns don't get that unwieldy.

autoload/neomake.vim
blueyed: > Since it's a list of processors now, passing through `entry != before` might help debug a processor that was affected by a prior processor.

You mean to set this on `s:this`/`self`, e.g. `s:this.changed = entry != before`?
We could also pass in `before` as `orig_entry` then instead.

blueyed:
@tweekmonster 
I think the main reason for pinging you was because you've mentioned to use some global postprocessing, which could happen on the list of all entries (outside of the current loop).

However, it's probably fine like this for now, and I'd skip documenting it for now to keep it experimental.

tweekmonster:
Ah, sorry.  I was just taking the changes at face value.  But, yes, this looks good at as it is.

Now that I'm seeing this fleshed out, it might be better if there was a global user-defined post-processor that runs *after* the maker processors (right before `entry != before`).  Since it's used-defined, the user would be trying to make a processor based on what they see in the messages.

PR #1035, opened 2017-02-23
Typescript maker, tsc, does not pickup the tsconfig.json if cwd is set

ebenoist:
As noted in issue #499 and #565 and solved in #731, this setting does not seem to work with newer versions of typescript. Removing cwd seems to do the trick.

@mhartington @blueyed any thoughts?

### DIFF
--- a/autoload/neomake/makers/ft/typescript.vim
+++ b/autoload/neomake/makers/ft/typescript.vim
@@ -8,9 +8,8 @@ function! neomake#makers#ft#typescript#tsc() abort
     " tsc should not be passed a single file.  Changing to the file's dir will
     " make it look upwards for a tsconfig.json file.
     return {
-        \ 'args': ['--noEmit'],
+        \ 'args': ['--project', neomake#utils#FindGlobFile(expand('%:p:h'), 'tsconfig.json'), '--noEmit'],
         \ 'append_file': 0,
-        \ 'cwd': '%:p:h',
         \ 'errorformat':
             \ '%E%f %#(%l\,%c): error %m,' .
             \ '%E%f %#(%l\,%c): %m,' .


mhartington:
My 2cents...

I've never used the default maker, and have been using something like this.

```
let g:neomake_typescript_tsc_maker = {
          \ 'args': ['--project', getcwd() . '/tsconfig.json', '--noEmit'],
          \ 'append_file': 0,
          \ 'errorformat':
          \   '%E%f %#(%l\,%c): error %m,' .
          \   '%E%f %#(%l\,%c): %m,' .
          \   '%Eerror %m,' .
          \   '%C%\s%\+%m'
        \ }
```

`tsc` works pretty differently than say something like tslint or eslint. It needs to have the whole scope of the project in order to work correctly.

blueyed:
Fine by me I guess.

It depends on your `cwd` then I guess still?
I.e. it will use the cwd where Neomake was started from then.

I think a helper function to find the tsconfig might be useful in this regard (pending in https://github.com/neomake/neomake/pull/993/files#diff-58e65732fcc7ab450ae138122e6ee651R424).

ebenoist:
@mhartington Updated to follow that pattern, works well for me locally as well. @blueyed you're right this would still depend on where you open your project, but I think this will provide a good out of the box experience for most folks.

mhartington:
So @blueyed, how would this work? 

Something like 

```
'args': ['--project',  neomake#utils#FindGlobFile(expand('%:p:h'), 'tsconfig.json'), '--noEmit'],
```

autoload/neomake/makers/ft/typescript.vim
```diff
@@ -8,9 +8,8 @@ function! neomake#makers#ft#typescript#tsc() abort
     " tsc should not be passed a single file.  Changing to the file's dir will
     " make it look upwards for a tsconfig.json file.
     return {
-        \ 'args': ['--noEmit'],
+        \ 'args': ['--project', getcwd() . '/tsconfig.json', '--noEmit'],
```
blueyed: Wouldn't just `./tsconfig.json` also work then?  (and allow to configure `neomake_typescript_tsc_cwd` if you want to)

blueyed:
@mhartington 
That would be the idea, yes.

mhartington:
In that case, I think using the `neomake#utils#FindGlobFile` would be a better solution, as not every project uses a root level tsconfig.

blueyed:
Then let's get https://github.com/neomake/neomake/pull/993 in first.

blueyed:
@ebenoist 
#993 is merged, so this could be rebased/changed, if you also agree.

ebenoist:
@blueyed totally onboard, is the final verdict

```
    return {
        \ 'args': ['--project', '/tsconfig.json', '--noEmit'],
        \ 'append_file': 0,
        \ 'errorformat':
            \ '%E%f %#(%l\,%c): error %m,' .
            \ '%E%f %#(%l\,%c): %m,' .
            \ '%Eerror %m,' .
            \ '%C%\s%\+%m'
        \ }

```

blueyed:
?
See https://github.com/neomake/neomake/pull/1035#issuecomment-282045875.

ebenoist:
@blueyed got it-- updated.

blueyed:
Thanks, merged (I've also changed the comment).

silviogutierrez:
So this improves the tsc checker, but in my experience coming over from syntastic, it's much better to use tsuquyomi for actual typescript checking. It's very fast and uses TSServer instead of the tsc compiler directly.

For syntastic, they provide a checker here:

https://github.com/Quramy/tsuquyomi/blob/master/syntax_checkers/typescript/tsuquyomi.vim

However, I haven't been able to find a way to use Neomake to invoke *just* a vim function instead of an external executable. Any ideas how we could create a similar Neomake maker off the file from above?

Thanks a lot for all the work on this plugin. Typescript checking is the last thing stopping my full migration over from syntasttic.

-
Silvio

mhartington:
@silviogutierrez I've authored [nvim-typescript](https://github.com/mhartington/nvim-typescript) for neovim support. I can try to add this functionality to nvim-typescript and integrate with neomake

blueyed:
> However, I haven't been able to find a way to use Neomake to invoke just a vim function instead of an external executable. Any ideas how we could create a similar Neomake maker off the file from above?

It's currently not possible AFAIK, but there are several issues / is some discussion about this already.

It would not be hard, but it's not clear currently how to approach it best.

I could imagine for `maker.exe` just being a function, or something similar.
But then again it's not async in this case anymore: jobs only work for external processes.
This means that we could have a Vim function for `maker.exe`, that would then spawn a job/process itself?

silviogutierrez:
Yes, I believe tsuquyomi.vim itself dispatches to a job manager (which is why it depends on vimproc). It's async using syntastic and quite fast.

blueyed:
@silviogutierrez 
Check out https://github.com/neomake/neomake/pull/1050: try to create a maker that is just:
```vim
let g:neomake_FT_NAME_maker = { 'get_list_entries': function('SyntaxCheckers_typescript_tsuquyomi_GetLocList') }
```

You would need to copy the function from Syntastic, and it's probably better to add it to `./autoload/neomake/makers/ft/typescript.vim` directly for hacking on it.

PR #1048, opened 2017-02-25
Add neomake#utils#JSONdecode

blueyed:
This is useful in general.
I plan to use it for a new redpen linter: https://github.com/redpen-cc/redpen

### DIFF
--- a/autoload/neomake/compat.vim
+++ b/autoload/neomake/compat.vim
@@ -21,7 +21,9 @@ if exists('*json_decode')
 else
   let neomake#compat#json_true = 1
   let neomake#compat#json_false = 0
-  let neomake#compat#json_null = {}
+  function! s:json_null() abort
+  endfunction
+  let neomake#compat#json_null = [function('s:json_null')]
 
   " Via Syntastic (https://github.com/vim-syntastic/syntastic/blob/6fb14d624b6081459360fdbba743f82cf84c8f92/autoload/syntastic/preprocess.vim#L576-L607),
   " based on https://github.com/MarcWeber/vim-addon-json-encoding/blob/master/autoload/json_encoding.vim.
--- a/autoload/neomake/compat.vim
+++ b/autoload/neomake/compat.vim
@@ -9,3 +9,56 @@ else
     return get(getbufvar(a:buf, ''), a:key, a:def)
   endfunction
 endif
+
+if exists('*json_decode')
+  let neomake#compat#json_true = v:true
+  let neomake#compat#json_false = v:false
+  let neomake#compat#json_null = v:null
+
+  function! neomake#compat#json_decode(json) abort
+    return json_decode(a:json)
+  endfunction
+else
+  let neomake#compat#json_true = 1
+  let neomake#compat#json_false = 0
+  let neomake#compat#json_null = {}
+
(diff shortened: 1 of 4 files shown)


autoload/neomake/compat.vim
```diff
@@ -9,3 +9,56 @@ else
...
+  let neomake#compat#json_false = v:false
+  let neomake#compat#json_null = v:null
+
+  function! neomake#compat#json_decode(json) abort
+    return json_decode(a:json)
+  endfunction
+else
+  let neomake#compat#json_true = 1
+  let neomake#compat#json_false = 0
+  let neomake#compat#json_null = {}
```
tweekmonster: Though `json_true` and `json_false` is good enough in terms of compatibility, I think `json_null` could lead to false positives.  By using `{}`, a distinction between an empty JSON object and `null` would be impossible.

It might be better to set its value to a script-local function can't be duplicated by accident:

```vim
function! s:json_null()
endfunction

let neomake#compat#json_null = [function('s:json_null')]
```

This is along the lines of using `null = object()` in Python.

With this, there'd only be one "true" `null` value for the compat `json_decode()`.  The same would need to be done for `json_true` and `json_false` if there's ever a need to create a compat `json_encode()`.

tests/json.vader
```diff
@@ -0,0 +1,27 @@
...
+  \   'true': neomake#compat#json_true,
+  \   'children': [],
+  \   'null': neomake#compat#json_null}}
+
+  if exists('v:null')
+    AssertEqual neomake#compat#json_null, v:null
+    AssertEqual neomake#compat#json_true, v:true
+    AssertEqual neomake#compat#json_false, v:false
+  else
+    AssertEqual neomake#compat#json_null, {}
```
tweekmonster: As mentioned above, this doesn't seem like it would be a meaningful test.

autoload/neomake/compat.vim
blueyed: Good point!
Why did you chose to use a list, and not just a function?

tweekmonster: To keep Vim from complaining about name of the variable:

```vim
let neomake#compat#json_null = function('s:json_null')
```

```
E704: Funcref variable name must start with a capital: neomake#compat#json_null
```

tweekmonster: Also, because it's a shorter way to keep an object ref vs `{'null': function('s:json_null')}`

tests/json.vader
```diff
@@ -0,0 +1,27 @@
...
+  \   'true': g:neomake#compat#json_true,
+  \   'children': [],
+  \   'null': g:neomake#compat#json_null}}
+
+  if exists('v:null')
+    AssertEqual g:neomake#compat#json_null, v:null
+    AssertEqual g:neomake#compat#json_true, v:true
+    AssertEqual g:neomake#compat#json_false, v:false
+  else
+    AssertEqual type(g:neomake#compat#json_null), type([])
```
tweekmonster: Should there be one more test?

```vim
AssertEqual type(g:neomake#compat#json_null[0]), type(function('tr'))
```

Or would that be going overboard?

blueyed: Why not?  :)
Thanks!

Issue #190, opened 2015-09-28
Allow linting on-the-fly (by piping buffer content to maker)


dan-santana:
It would be nice to have an option to pipe the buffer content to the maker instead of relying only on the file saved to disk.

With this feature, it would be possible to make on-the-fly linting with some tools (e.g. `eslint` with `--stdin` and `--stdin-filename` options) by `:Neomake`ing on `CursorHold`.

For example, the following would suffice to make the magic happen:

```
let g:neomake_javascript_eslint_maker = {
        \ 'pipe': 1,
        \ 'args': ['-f', 'compact', '--stdin', '--stdin-filename', '%:p'],
        \ 'errorformat': '%E%f: line %l\, col %c\, Error - %m,' .
        \ '%W%f: line %l\, col %c\, Warning - %m'
        \ }

autocmd! CursorHold *.js,*.jsx Neomake
```

Bonus: It would allow testing the code on-the-fly too, if the test setup is fast enough.

benekastah:
This is a nice idea. I would be happy to review a pull request for something like this, but it's unlikely I'll implement features I don't need myself anytime soon.

dan-silva:
Good! I'll try to work on this as soon as possible (I'm still in a transition from Vim to Neovim, and this would be a killer feature to help me change completely).

gunar:
I use [vim-auto-save](https://github.com/vim-scripts/vim-auto-save) and I'm quite happy with it. Bonus: it solves your problem.

johannestroeger:
no, it just introduce side-effects.

dan-santana:
Actually using [vim-auto-save](https://github.com/vim-scripts/vim-auto-save) would create another problem in my setup: rebuild will be triggered when using [gulp](http://gulpjs.com/) or [webpack](https://webpack.github.io/) watchers.

Linting a single file is already costly in some projects, and if I also rebuild the entire project every time... probably my computer will explode (although it may work well for small projects).

gunar:
So I shouldn't be doing this? I've just transitioned from vim so I'll use it like this for a while and give feedback later on.

@dan-santana what kind of whatches are you talking about?

dan-santana:
Not that you shouldn't use it – if it fits to your workflow, there's no problem at all.

[...]

PR #1055, opened 2017-02-26
[RFC] autolinting

blueyed:
This is a step into the direction of better linting on-the-fly etc.

Use `call neomake#configure#autolint('niw', 100)` to configure it in
normal and insert mode with a delay of 100ms, and to setup a `BufWritePost` autocommand.

Use `i` to configure it for insert mode only.

The default delay is 500ms.

TODO:
 - [ ] better config (see https://github.com/dojoteef/neomake-autolint/commit/c0e1d874aac953bd351cf140a9b4be39f5902174 / https://github.com/dojoteef/neomake-autolint/issues/8).
 - [x] split out handling of temp files: https://github.com/neomake/neomake/pull/1057
 - [ ] allow config overrides based on autolint mode (by refactoring settings in general (dict based))
 - [ ] short-circuit maker init/setup e.g. while in the same mode / settings were not changed (performance); basically try to just run job_start again.
 - [ ] explicitly cancel jobs when restarting, instead of letting Neomake handle it?!

/cc @dojoteef

### DIFF
--- a/autoload/neomake/configure.vim
+++ b/autoload/neomake/configure.vim
@@ -31,6 +31,7 @@ function! neomake#configure#autolint(modes, ...) abort
     au!
     if len(a:modes)
       if !has('timers')
+        call neomake#utils#ErrorMessage('Timer support is required for autolinting.')
       endif
       if a:modes =~# 'n'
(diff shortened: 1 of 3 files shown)


blueyed:
Moving this over to https://github.com/neomake/neomake/pull/1167.

Issue #552, opened 2016-08-12
Handle fugitive buffers


blueyed:
When invoked on a [fugitive](https://github.com/tpope/vim-fugitive/) buffer, we should write the contents to an actual file and then run the makers on that.
This should include re-writing the makers output to apply to the fugitive buffer correctly.

blueyed:
Works nicely via #1055.

Issue #73, opened 2015-03-29
g:neomake_make_modified is not compatible with all checkers


Neki:
In some languages, the file name (and directory structure) is significant. For instance, Java requires the class "org.company.MyClass" to be defined in a file at `org/company/MyClass.java`. The same goes for Haskell (ghc-mod will refuse to check a file whose name does not end in `.hs` or `.lhs`, and I am pretty sure that messing with the Haskell naming conventions will break some tools), Python, and probably a lot of other languages.

When `g:neomake_make_modified` is set to 1, the name of the file that is actually checked does not (and can not) follow these rules, and thus the "run make on modified buffers" feature will simply not work. I verified this with the ghc-mod checker for Haskell, that simply outputs to stderr `ghc-mod: target src/Test.hs.1605.neomake.tmp is not a module name or a source file`.

I can foresee other sources of issues: checkers that can only check a whole project at once, issues if `:Neomake` and `:Neomake!` were to be run concurrently...

I see two options:
* dropping this feature altogether (the above is the reason why Syntastic does not support it, see scrooloose/syntastic#1034 for instance). This is my personal preference, I think this feature is too much trouble for little benefit.
* have checkers explicitly declare that they support this feature, which is still feasible at this point since there are not a lot of checkers.

Thoughts ?

benekastah:
I think dropping the feature is probably the best course, too. @dpavlic asked for this feature in #28. Are you still using this feature?

blueyed:
See https://github.com/neomake/neomake/pull/1057 for bringing it back.

PR #1057, opened 2017-02-27
Handle unreadable/modified buffers

blueyed:
This includes unnamed and modified buffers, and helps with doing linting
on-the-fly (autolint).

It also enables support for linting of unreadable files (e.g. fugitive).

Fixes https://github.com/neomake/neomake/issues/552.

Extracted out of https://github.com/neomake/neomake/pull/1055, to get this in first.

There was `neomake_make_modified` already (removed in #73 again (143725dc)).

I think we need a way to whitelist makers and/or configure how the temporary file gets created.
Some linters support an option like `--stdin`, which could be used in case of modified files.

There is a new setting `tempfile_enabled`, which defaults to 1 currently.
Makers that cannot handle temporary files should disable it.
`tempfile_name` can be used to specify a filename, which would allow to e.g. copy stuff somewhere to make the linter use it properly.

TODO:
 - [ ] change default to disabled, to get this merged, without causing too much trouble (#1092 is based on this).  (done in https://github.com/neomake/neomake/pull/1110).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,5 +64,6 @@ script:
       if [[ "$ENV" = "testvim" ]]; then
         vim --version
       fi
+      umask 022
       make $ENV $MAKE_ARGS
     fi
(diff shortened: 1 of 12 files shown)


blueyed:
Merged via https://github.com/neomake/neomake/pull/1092, which was based on this.

Issue #1043, opened 2017-02-24
Make some variables maker-dependant rather than global


teto:
I wonder if it were possible to let the makers override more of the neomake variables (as is done with neomake_remove_invalid_entries); such as:
- g:neomake_info_sign / warn/error
- g:neomake_place_signs
- g:neomake_open_list

### Expected behavior

When I use linters, I don't want to autoopen the location list for instance, while when running Neomake! make, I do want the quickfix to open.

blueyed:
Good idea!

PR #1064, opened 2017-03-02
Make open_list and place_signs local-global

teto:
Addresses https://github.com/neomake/neomake/issues/1043#issuecomment-282279578 ?

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -17,10 +17,10 @@ let s:need_errors_cleaning = {
     \ 'file': {}
     \ }
 let s:maker_defaults = {
-            \ 'buffer_output': 1,
-            \ 'remove_invalid_entries': get(g:,'neomake_remove_invalid_entries', 0),
-            \ 'place_signs': get(g:, 'neomake_place_signs', 1),
-            \ 'open_list': get(g:, 'neomake_open_list', 0)}
+            \ 'buffer_output': 1}
+            " \ 'remove_invalid_entries': get(g:,'neomake_remove_invalid_entries', 0),
+            " \ 'place_signs': get(g:, 'neomake_place_signs', 1),
+            " \ 'open_list': get(g:, 'neomake_open_list', 0)}
 let s:project_job_output = {}
 " List of job ids with pending output per buffer.
 let s:buffer_job_output = {}
@@ -408,7 +408,10 @@ function! neomake#GetMaker(name_or_maker, ...) abort
         \ 'exe': maker.name,
         \ 'args': [],
         \ 'errorformat': &errorformat,
-        \ })
+        \ 'remove_invalid_entries': get(g:,'neomake_remove_invalid_entries', 0),
+        \ 'place_signs': get(g:, 'neomake_place_signs', 1),
+        \ 'open_list': get(g:, 'neomake_open_list', 0)
+        \ }, "keep")
     let bufnr = bufnr('%')
     for [key, default] in items(defaults)
         let maker[key] = neomake#utils#GetSetting(key, maker, default, fts, bufnr)
@@ -530,16 +533,16 @@ function! neomake#GetEnabledMakers(...) abort
     return enabled_makers
 endfunction
 
-function! s:HandleLoclistQflistDisplay(open_val, file_mode) abort
-    if a:open_val
+function! s:HandleLoclistQflistDisplay(jobinfo) abort
+    if a:jobinfo.open_list
         let height = get(g:, 'neomake_list_height', 10)
         let win_val = winnr()
-        if a:file_mode
+        if a:jobinfo.file_mode
             exe 'lwindow' height
         else
             exe 'cwindow' height
         endif
-        if a:open_val == 2 && win_val != winnr()
+        if a:jobinfo.open_list == 2 && win_val != winnr()
             wincmd p
         endif
     endif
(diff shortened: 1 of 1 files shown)


autoload/neomake.vim
```diff
@@ -18,7 +18,9 @@ let s:need_errors_cleaning = {
     \ }
 let s:maker_defaults = {
             \ 'buffer_output': 1,
-            \ 'remove_invalid_entries': 0}
+            \ 'remove_invalid_entries': get(g:,'neomake_remove_invalid_entries', 0),
+            \ 'place_signs': get(g:, 'neomake_place_signs', 1),
+            \ 'open_list': get(g:, 'neomake_open_list', 0)}
```
blueyed: This would not pick up changed settings (unless you resource the file).
I think it should be moved to https://github.com/neomake/neomake/blob/3c52f9f27a9ecbf32ed5f79b5faf270ce4cf3ab8/autoload/neomake.vim#L439-L443.

autoload/neomake.vim
```diff
@@ -528,17 +530,16 @@ function! neomake#GetEnabledMakers(...) abort
     return enabled_makers
 endfunction
 
-function! s:HandleLoclistQflistDisplay(file_mode) abort
-    let open_val = get(g:, 'neomake_open_list')
-    if open_val
+function! s:HandleLoclistQflistDisplay(open_val, file_mode) abort
```
blueyed: Pass in just `jobinfo` here? (which contains `maker.open_list` and `file_mode` then)

autoload/neomake.vim
teto: Also it should be :
```
call extend(defaults, {
        \ 'exe': maker.name,
        \ 'args': [],
        \ 'errorformat': &errorformat,
        \ 'remove_invalid_entries': get(g:,'neomake_remove_invalid_entries', 0),
        \ 'place_signs': get(g:, 'neomake_place_signs', 1),
        \ 'open_list': get(g:, 'neomake_open_list', 0)}
        \ }, "keep")
```
to keep user configuration

autoload/neomake.vim
teto: I do want to to have the settings maker-dependant: i.e., some makers autoopen the Quickfix, others not. I can transfer those properties to jobinfo afterwards ?

blueyed: The `defaults` from above are just the defaults for the makers - so it should work as before regarding this.

Speaking about it: having tests for this would prove that.. :)

teto: I hate vimscript and not very excited about writing the tests :p 
Which test would you recommand to use as a base ? or should I extend an existing one ?

blueyed: Feel free to skip it then: I will pick this up after the open PRs about tempfiles, autolinting etc are finished/merged.

blueyed:
@teto 
Any update on this?

teto:
I recently gave it a look but I gave up because of the conflicts and I knew i wouldn't write the tests. I would really love to see it as for linters, signs are enough IMO.

blueyed:
Thanks for reporting back.

I will look into this myself, likely based on the new-style config (`neomake#config#set()`).

teto:
I've fixed some linters that were not available on my system and this is driving me crazy, I have the linters location list + quickfix compilation autoopening. Along with vim-grepper window, I can't see my code anymore.
I believe the auto_open should be a maker property rather than job's. I would rather have linters location list stay hidden (and show marks) while having the quickfix autoopen makes me realize when the compiler finished his job.
I have no opinion for the place_signs variable as it's usually not a problem.

blueyed:
You might want to also try https://github.com/blueyed/vim-qf_resize.

autoload/neomake.vim
```diff
@@ -528,17 +534,16 @@ function! neomake#GetEnabledMakers(...) abort
     return enabled_makers
 endfunction
 
-function! s:HandleLoclistQflistDisplay(file_mode) abort
-    let open_val = get(g:, 'neomake_open_list')
```
blueyed: Might be as easy as just using `neomake#utils#GetSetting` here..

teto:
@blueyed any hope you can take over this :> ? I am experimenting with a nix-shell maker and would like to have the list autoopen just for this maker.

blueyed:
Sure.  Hope never dies.. :)
Shouldn't be that hard.

blueyed:
btw: you can use `neomake_open_list = 2` and then handle opening the list yourself in a `NeomakeJobFinished` / `NeomakeFinished` User hook.
That's what I am doing myself, to open it only in case of errors.
But the/this issue still makes sense.

blueyed:
Ok, I can see that it makes sense to control signs better, and the same with highlights.
Will keep this open as a reminder.
Thanks for your feedback!

blueyed:
@teto 
JFI: I often just use `:NeomakeDisable(Buffer/Tab)` when working on projects that are not compliant with flake8 or whatever default makers I am using for a language.

teto:
Thanks I doubt I'll ever complete this PR so let's cleanup your queue. I'll use your suggestion instead.

blueyed:
No, it's fine to keep this open as a reminder (at least for me ;))

teto:
I use neomake just for nvim-dev and cleaninbg up my issues hence closing.

PR #1065, opened 2017-03-02
Move restart check into s:MakeJob

blueyed:
This also stores the maker from after the `maker.fn` on `jobinfo`.

Not sure about using an exception here (instead of a special return
value).

/cc @tweekmonster ^^

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -97,7 +97,9 @@ function! neomake#CancelJob(job_id, ...) abort
             endtry
         else
             let vim_job = jobinfo.vim_job
-            if job_status(vim_job) !=# 'run'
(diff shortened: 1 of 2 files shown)


autoload/neomake.vim
```diff
@@ -161,6 +159,33 @@ function! s:MakeJob(make_id, options) abort
             let maker = returned_maker
         endif
     endif
+    lockvar maker
+    let jobinfo.maker = maker
+
+    " Check for already running job for the same maker (from other runs).
+    " This used to use this key: maker.name.',ft='.maker.ft.',buf='.maker.bufnr
```
blueyed: When going back to using a key, this could be moved back again (to restart earlier).
Just using the maker name per buffer might be sane anyway?!

Issue #441, opened 2016-05-25
autocmd BufEnter stuck in loclist


cycomanic:
Hi, 

I'm having issues using autocmd BufEnter with `let g:neomake_open_list = 2` The loclist never shows up and I stop seeing my commandline. If I set open_list=1 I am stuck in the locationlist, i.e. as soon as I switch into the buffer which contains the code I'm send back to the loclist. This was mentioned by @ @edwardsmit in #408 . Tested with neovim 0.1.3 and pylint

edwardsmit:
@cycomanic I think it's the *exact* same problem I'm seeing.

tamorim:
@cycomanic @edwardsmit 
I had the same problem.
Switching to `BufWinEnter` solved for me!

blueyed:
I assume this to be fixed by beb34c9.

PR #1076, opened 2017-03-05
Lint on beginning

raviqqe:


blueyed:
Thanks for the PR, but using `BufEnter` is problematic: https://github.com/neomake/neomake/issues/441#issuecomment-226920597.
(It also depends on if you use `neomake_open_list`)

I would rather have some better default configuration, where the autocommands would be handled then (see https://github.com/neomake/neomake/projects/1).

In https://github.com/neomake/neomake/pull/1055 there is some attempt at that.

I am closing this PR therefore.

PR #1085, opened 2017-03-06
Fix flow in jsx

stevenyap:
This commit adds flow makers to JSX.

Coders who use flow generally uses JSX too so we need it for both.

### DIFF
--- a/autoload/neomake/makers/ft/jsx.vim
+++ b/autoload/neomake/makers/ft/jsx.vim
@@ -29,3 +29,7 @@ endfunction
 function! neomake#makers#ft#jsx#rjsx()
     return neomake#makers#ft#javascript#rjsx()
 endfunction
+
+function! neomake#makers#ft#jsx#flow()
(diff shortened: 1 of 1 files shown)


matthewlehner:
I've just done the exact same thing on matthewlehner/neomake@9804271 – looks like that test failure might be an intermittent failure as the new JSX flow function follows the same style as the other functions in this file.

Can someone trigger a rerun, or have a look manually?

blueyed:
Thanks.

`vint-diff` is right, `abort` should be used for all functions in the file:

> autoload/neomake/makers/ft/jsx.vim:3:1: Use the abort attribute for functions in autoload (see Google VimScript Style Guide (Functions))

I can handle that later.

matthewlehner:
@blueyed I've updated my fork with and submitted #1122 with the abort attributes added to the functions.

blueyed:
Closing in favor of https://github.com/neomake/neomake/pull/1122.

Issue #1089, opened 2017-03-06
Java Linting Not Working Correctly


IQubic:
I'm trying to lint a multi-file Java project. When I ask Neomake to lint one of my source files it tells me that there is a "symbol not found".
Let's say the file is something like this:

    package textExcel;
    
    public class RealCell implements Cell {
    }

Neomake tells me that the symbol Cell can't be found. That's bogus. Cell is another file in the same directory, also in the same package.

I even added this to my .vimrc:

    let g:neomake_java_javac_args = ['-cp', '~/textExcel']

And that's still not working.

### Expected behavior

I want Neomake to feed my javac args to javac when linting.

### Steps to reproduce

If you want to take a look at the project that isn't linting correctly, it can be found here:
https://github.com/IQubic/Text-Excel

But in general just create a java project with at least two classes, in separate files. Then have one depend on the other. Linting will most likely fail.

### Output from (verbose) NeomakeInfo
Neomake [155.702]: [6] Starting async job: ['javac', '-cp', '~/textExcel/src/textExcel', '/home/avi/textExcel/src/textExcel/RealCell.java']
 
#### Neomake debug information
Async support: 1
Current filetype: java

##### Enabled makers
For the current filetype (with :Neomake):
 -java

IQubic:
Turns out I misunderstood how to use the -cp argument. Closing this thread as Neomake works just fine, I was just being stupid.

IQubic:
Reopening this because I still can't get this to work.
This works from command line:

    javac -cp ~/textExcel/src RealCell.java

But adding this to my vimrc:

   let g:neomake_java_javac_args = ['-cp', '~/textExcel/src/']

Still gives me "Cannot find symbol"

Is there something I'm missing here?

blueyed:
Try an absolute path.. `~` might not get resolved.
We're using `expand()` but it's not triggered here.

Try `let g:neomake_java_javac_args = ['-cp', expand('~').'/textExcel/src/']`.

/cc @tweekmonster

tweekmonster:
I always just assumed that Vim sorted out the tilde.  I guess this should be covered by `ExpandArgs()`

blueyed:
It might depend on how jobs are run, i.e. if the shell resolves it.

@IQubic 
Are you using Vim/Neovim?  Windows/Linux?

blueyed:
Closing this issue itself, but I am still curious about @IQubic's setup - that's also why I've created an issue template in the first place.. ;)

[...]

PR #1091, opened 2017-03-06
Expand tilde in neomake#utils#ExpandArgs()

blueyed:
Ref: https://github.com/neomake/neomake/issues/1089

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -384,6 +384,11 @@ function! neomake#utils#ExpandArgs(args) abort
                     \ . '''\(\%(\\\@<!\\\)\@<!%\%(%\|\%(:[phtre]\+\)*\)\ze\)\w\@!'', '
                     \ . '''\=(submatch(1) == "%%" ? "%" : expand(submatch(1)))'', '
                     \ . '''g'')')
+        let ret = map(ret,
+                    \ 'substitute(v:val, '
+                    \ . '''\(\%(\\\@<!\\\)\@<!\~\)'', '
+                    \ . 'expand(''~''), '
+                    \ . '''g'')')
     finally
         let &iskeyword = isk
     endtry
--- a/tests/utils.vader
+++ b/tests/utils.vader
@@ -156,6 +156,9 @@ Execute(neomake#utils#redir):
   AssertThrows neomake#utils#redir(['NeomakeTestCommand', 'echoerr 3'])
 
 Execute (neomake#utils#ExpandArgs):
+  Save $HOME
+  let $HOME='/home/sweet'
+
   file file1.ext
   let isk = &iskeyword
   let args = [
@@ -176,6 +179,8 @@ Execute (neomake#utils#ExpandArgs):
   \ '^%$',
   \ 'C:\\%\\',
   \ 'C:\\%:r\\',
+  \ '~',
+  \ '\~',
   \ ]
   let expected_args = [
   \ 'file1.ext',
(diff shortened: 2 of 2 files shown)


autoload/neomake/utils.vim
```diff
@@ -379,8 +379,14 @@ function! neomake#utils#ExpandArgs(args) abort
     let isk = &iskeyword
     set iskeyword=p,h,t,r,e,%,:
     try
+        let home = expand('~')
         let ret = map(a:args,
                     \ 'substitute(v:val, '
+                    \ . '''\(\%(\\\@<!\\\)\@<!\~\)'', '
```
tweekmonster: I tried to include this in the pattern below, but I was having a hard time getting it to work reliably.

tweekmonster:
There appears to be a problem with the `Stringify()` test.  I thought that my changes were causing it to fail, but it seems to fail on master with `make docker_test`.

blueyed:
> There appears to be a problem with the `Stringify()` test.  I thought that my changes were causing it to fail, but it seems to fail on master with `make docker_test`.

Works for me.  How does it fail?
I've noticed that it fails for nvim 0.1.7 though:

>    (108/147) [EXECUTE] (X) '{f2: function(''8'', {…})}' should be equal to '{f2: function(''8'')}'
> tests/utils.vader:311: AssertEqual neomake#utils#Stringify(obj), "{f2: function('".(fnr_base + 1)."', {…})}"

Any idea what patch (that was added later in Neovim) is that?
Otherwise I would just assume the new behavior with neovim 0.2.0?!

tweekmonster:
Sorry, I should've posted the error.  I get this from the Vim 8.0.0238 test:

```
(108/149) [EXECUTE] (X) Vim(let):E704: Funcref variable name must start with a capital: exp
```

But, I don't see `exp` anywhere.

autoload/neomake/utils.vim
```diff
@@ -379,16 +379,15 @@ function! neomake#utils#ExpandArgs(args) abort
...
-                    \ . 'home, '
+                    \ . '''\(\%(\\\@<!\\\)\@<!%\%(%\|\%(:[phtre]\+\)*\)\ze\)\w\@!'', '
+                    \ . '''\=(submatch(1) == "%%" ? "%" : expand(submatch(1)))'', '
                     \ . '''g'')')
         let ret = map(ret,
                     \ 'substitute(v:val, '
-                    \ . '''\(\%(\\\@<!\\\)\@<!%\%(%\|\%(:[phtre]\+\)*\)\ze\)\w\@!'', '
-                    \ . '''\=(submatch(1) == "%%" ? "%" : expand(submatch(1)))'', '
+                    \ . '''\(\%(\\\@<!\\\)\@<!\~\)'', '
+                    \ . 'expand(''~''), '
```
tweekmonster: Why did you switch the order?  This doesn't bother me.  I'm just curious.

blueyed: I thought it would be better, in case $HOME expands to something that would get picked up later then.

blueyed:
You need to update Vader (https://github.com/junegunn/vader.vim/pull/121).

tweekmonster:
@blueyed Forgot to mention it last night.  I cloned your fork of vader in the test directory and the test passed.  A symlink outside of the repo was causing the docker test to hang.

blueyed:
FWIW: I am using a symlink to use my Vader always: ./tests/vim/plugins/vader.override -> /home/user/.dotfiles/vim/plugged/vader.vim/
The issue itself should be fixed in Vader itself, but the once checked out version is not updated automatically.
Are you using https://github.com/junegunn/vader.vim/pull/107 then now?

Issue #572, opened 2016-09-01
Fix auto-selection of enabled makers for multi-filetype


blueyed:
With a filetype like `vader.vim` or `javascript.jsx` the makers that are returned by [neomake#GetEnabledMakers](https://github.com/neomake/neomake/blob/914609831b904f84f296826c2ad8f99d4bc7ce23/autoload/neomake.vim#L295) are only those that are present in all filetypes..

This causes issues like #272 and #492.

(I've just seen that there is a undocumented `SupersetOf` hook in makers, where e.g. `jsx` says "javascript" (https://github.com/neomake/neomake/blob/914609831b904f84f296826c2ad8f99d4bc7ce23/autoload/neomake/makers/ft/jsx.vim#L3-L5).
It seems to only be used for sorting the filetypes though (https://github.com/neomake/neomake/blob/914609831b904f84f296826c2ad8f99d4bc7ce23/autoload/neomake/utils.vim#L163-L175), and I think that is over-engineered: the order should be defined by the user (given the order in the `&ft`) - another issue.)

I could imagine the following fixes: 
1. just use all makers from all mentioned filetypes, and not the union of it.  Where would that fail?
1. ignore empty lists of (auto-detected only?) makers.  For `vader.vim` that would just use the ones from `vim`.  For `javascript.jsx` it would still use the union - but e.g. in the case of `eslint` that _should_ have worked before already.  Maybe we can make use of the `SupersetOf` here?

btw: there is no way to configure enabled makers for `foo.bar` explicitly currently, since you cannot specify `g:neomake_foo.bar_enabled_makers` (because of the dot in there).

jamesoff:
From the brief look I've had at this, removing the filter() call on the list of enabled makers seems to give me the result I need, but I agree that there may be cases where we don't want all of the makers to run - not that I can think of any right now.

I believe that's how Syntactic does it.

jeromedalbert:
I work on `.js.es6.erb` files that I have set up to have the filetype `eruby.javascript.es6`.

I want to ESLint those files, but as @blueyed explained, I cannot do this unless `eruby`, `javascript` and `es6` filetypes have the `eslint` maker. It does not make any sense to have an `eslint` maker for `eruby` though...

So here is the big hack I did for now:
```
function! Lint()
  if &filetype =~ 'javascript'
    Neomake eslint
  else
    Neomake
  end
endfunction

augroup lint_events
  autocmd!
  autocmd BufWritePost * call Lint()
augroup end
```

[...]

PR #1114, opened 2017-03-15
Add add standard and semistandard makers to JSX filetypes

indiesquidge:
I'm using the [`vim-jsx`](https://github.com/mxw/vim-jsx) plugin, which has a setting to use JSX in regular `.js` files by setting the filetype to `javascript.jsx`, which breaks neomake unless both the JSX _and_ JavaScript makers are enabled. ([This is a known issue with neomake](https://github.com/neomake/neomake/issues/572).)

However, I've recently upgraded this plugin and noticed that the standard maker was working fine in `javascript` filetypes, but not in `javascript.jsx` filetypes, even though I had the maker enabled for both.

It looks as though the maker lists for `jsx.vim` and `javascript.vim` are getting out of sync. This is a patch to help standard and semistandard work again in JSX files.

### DIFF
--- a/autoload/neomake/makers/ft/jsx.vim
+++ b/autoload/neomake/makers/ft/jsx.vim
@@ -22,6 +22,14 @@ function! neomake#makers#ft#jsx#eslint()
     return neomake#makers#ft#javascript#eslint()
 endfunction
 
+function! neomake#makers#ft#jsx#standard()
+    return neomake#makers#ft#javascript#standard()
+endfunction
+
+function! neomake#makers#ft#jsx#semistandard()
+    return neomake#makers#ft#javascript#semistandard()
+endfunction
+
 function! neomake#makers#ft#jsx#eslint_d()
     return neomake#makers#ft#javascript#eslint_d()
 endfunction


indiesquidge:
I'd be inclined to sync up more of the `jsx.vim` file with `javascript.vim`. I realize that may be band-aid to an otherwise much larger problem (I would assume the two filetypes won't always want to be maintained separately, as any maker in in a JS file should also work in JSX files).

Things like this could probably be cleaned up

```vim
function! neomake#makers#ft#jsx#jshint()
    let maker = neomake#makers#ft#javascript#jshint()
    let maker.exe = 'jsxhint'
    return maker
endfunction

function! neomake#makers#ft#jsx#jsxhint()
    return neomake#makers#ft#jsx#jshint()
endfunction
```

Couldn't we just say this?

```vim
function! neomake#makers#ft#jsx#jsxhint()
    return neomake#makers#ft#javascript#jshint()
endfunction
```

AFAICT the only point of the `jsx.vim` file is to alias all existing functions in `javascript.vim`, but I could be wrong.

seeARMS:
+1, recently upgraded this plugin as well and didn't realize why my JSX files weren't getting linted by standard until I came across this issue

matthewlehner:
@indiesquidge Vint is failing because the functions you've added are missing the `abort` attribute. I opened #1122 to add the flow maker to jsx.vim, then saw your PR and rolled in these makers as well.

blueyed:
So standard was working before in jsx file already, until some (recent) change in Neomake?

I've found a severe bug that might have caused this - fix:

```
diff --git i/autoload/neomake.vim w/autoload/neomake.vim
index 25e873bf..b57ce4e9 100644
--- i/autoload/neomake.vim
+++ w/autoload/neomake.vim
@@ -680,7 +680,7 @@ function! s:Make(options) abort
         let makers = options.enabled_makers
         unlet options.enabled_makers
     else
-        let makers = call('neomake#GetEnabledMakers', file_mode ? options.fts : [])
+        let makers = call('neomake#GetEnabledMakers', file_mode ? [options.fts] : [])
         if !len(makers)
             if file_mode
                 call neomake#utils#DebugMessage('Nothing to make: no enabled file mode makers.', {'make_id': make_id})
```

Will create a PR with a test for this now.

indiesquidge:
Yup, that fixed it for me, @blueyed. Thanks a bunch :)

PR #942, opened 2017-01-17
Cargo: Use json output to get better error messages

JelteF:
This modifies the cargo maker to use its json output. This makes it much easier
to parse than the human friendly version. 

It currently uses python to parse the JSON in a safe way. This obviously
requires +python to be available, which seems like a reasonable tradeoff.

I'm planning to add support for the length field in a future pull request/commit.

PR #1117, opened 2017-03-15
Add Elm maker using elm-make

keiththomps:
This adds a maker for [Elm](http://elm-lang.org/). The compiler
errors provided by elm are quite detailed, but this only adds
handling for the base error/warning types and short messages.

This maker relies on the JSON output of elm-make and is *heavily*
based on #942 (it's basically a carbon copy).

### DIFF
--- a/autoload/neomake/makers/ft/elm.vim
+++ b/autoload/neomake/makers/ft/elm.vim
@@ -8,48 +8,44 @@ function! neomake#makers#ft#elm#elmMake() abort
     return {
         \ 'exe': 'elm-make',
         \ 'args': ['--report=json', '--output=' . neomake#utils#DevNull()],
-        \ 'mapexpr': 'neomake#makers#ft#elm#ElmMakeMapexpr(v:val)',
-        \ 'errorformat':
-            \ '[%t%n] "%f" %l:%v %m,'.
-            \ '[%t] "%f" %l:%v %m',
-        \ 'postprocess': function('neomake#makers#ft#elm#ElmMakePostProcess')
+        \ 'process_output': function('neomake#makers#ft#elm#ElmMakeProcessOutput')
         \ }
 endfunction
 
-function! neomake#makers#ft#elm#ElmMakeMapexpr(val) abort
-    if a:val[0] !=# '['
-        return
+function! neomake#makers#ft#elm#ElmMakeProcessOutput(context) abort
+    " output will be a List, containing a JSON string of an array of objects
+    if a:context.output[0][0] !=# '['
+        return []
     endif
-    let l:decoded = neomake#utils#JSONdecode(a:val)
+    let l:decoded = neomake#utils#JSONdecode(a:context.output[0])
     if type(l:decoded) == type([])
-      for item in l:decoded
-        if get(item, 'type', '') ==# 'warning'
-          let l:code = 'W'
-        else
-          let l:code = 'E'
-        endif
+        let l:errors = []
+        for item in l:decoded
+            if get(item, 'type', '') ==# 'warning'
+                let l:code = 'W'
+            else
+                let l:code = 'E'
+            endif
 
-        let l:type = item['tag']
-        let l:message = item['overview']
-        let l:region_start = item['region']['start']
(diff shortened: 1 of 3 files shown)


keiththomps:
If there are any suggests as to how to utilize the `details` portion of the output I would love to hear them. The details are pretty useful, but very wordy and I couldn't think of a way to utilize them that would make sense. Here are the examples (used in the tests actually) printed out by the compiler:

autoload/neomake/makers/ft/elm.vim
```diff
@@ -0,0 +1,56 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#elm#EnabledMakers() abort
+    return ['elmMake']
+endfunction
+
+function! neomake#makers#ft#elm#elmMake() abort
+    return {
+        \ 'exe': 'elm-make',
+        \ 'args': ['--report=json', '%:p', '--output=' . neomake#utils#DevNull()],
```
blueyed: You do not need the `%:p`, do you?
`append_file` should append the current file.

tests/ft_elm.vader
```diff
@@ -0,0 +1,65 @@
+Include: include/setup.vader
+
+Execute (elmMake parse success message):
```
blueyed: s/parse/parses/

tests/ft_elm.vader
```diff
@@ -0,0 +1,65 @@
...
+  let &errorformat = neomake#makers#ft#elm#elmMake().errorformat
+
+  let error = neomake#makers#ft#elm#ElmMakeMapexpr(json)
+
+  AssertEqual '[E] "Bingo.elm" 6:1 4 BAD MAIN TYPE : The `main` value has an unsupported type.', error
+
+  cgetexpr error
+
+  let result = getqflist()[0]
+  unlet result.bufnr
```
blueyed: Is this a trick to not require the actua bufnr in the tests?

autoload/neomake/makers/ft/elm.vim
```diff
@@ -0,0 +1,56 @@
...
+      for item in l:decoded
+        if get(item, 'type', '') ==# 'warning'
+          let l:code = 'W'
+        else
+          let l:code = 'E'
+        endif
+
+        let l:type = item['tag']
+        let l:message = item['overview']
+        let l:detail = item['details']
```
blueyed: vimlint complains: `./autoload/neomake/makers/ft/elm.vim:34:13:Error: EVL102: unused variable2 `l:detail``

blueyed:
Given that I think #942 would benefit from https://github.com/neomake/neomake/pull/1050, this would then probably too.
I'll try to get that one ready/merged now.

tests/ft_elm.vader
keiththomps: Yeah, an alternative appears to be checking that `bufnr` equals `bufnr('%') + 1`, but that feels brittle. I'm not that familiar with vimscripting so any advice would be appreciated.

blueyed: Seems to be a nice shortcut after all in this case.

I tend to use `:file` to the the buffer name accordingly.

blueyed:
https://github.com/neomake/neomake/pull/1050 is ready, but missing docs.

keiththomps:
After reading #1050 I've got a rough idea of how I would use it (must easier to grok than setting up an `errorformat`). Once that's merged I'll adjust this pull request accordingly.

blueyed:
#1050 is merged now.  It has docs even, so you want to check them out.. :)

keiththomps:
I've updated this pull request to use `process_output` and it works great. For some reason warnings and errors are now being highlighted with the same icon compared to the last time I took a screen shot.

blueyed:
Thanks, please add it to the list of makers in the wiki!

keiththomps:
Will do :+1:. There's actually a slight change that I need to make to the parsing so that errors are displayed instead of warnings if one of each exists on the same line.

ahstro:
Awesome :clap: 🎉 😄

PR #1121, opened 2017-03-16
add flow maker for jsx filetypes

trevorwhitney:
Copying the flow maker over so it's available for jsx filetypes. Didn't think this needed a test, but please advise.

PR #1122, opened 2017-03-16
Adds missing js.vim makers to jsx.vim

matthewlehner:
This adds the Flow, Standard, and Semistandard makers from javascript.vim to jsx.vim, and updates the functions to have the abort attribute.

Rolls in changes from #1085, #1114, and #1121 with vint passing on Travis.

### DIFF
--- a/autoload/neomake/makers/ft/jsx.vim
+++ b/autoload/neomake/makers/ft/jsx.vim
@@ -1,32 +1,32 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#ft#jsx#SupersetOf()
+function! neomake#makers#ft#jsx#SupersetOf() abort
     return 'javascript'
 endfunction
 
-function! neomake#makers#ft#jsx#EnabledMakers()
+function! neomake#makers#ft#jsx#EnabledMakers() abort
     return ['jshint', 'eslint']
 endfunction
 
-function! neomake#makers#ft#jsx#jshint()
+function! neomake#makers#ft#jsx#jshint() abort
     let maker = neomake#makers#ft#javascript#jshint()
     let maker.exe = 'jsxhint'
     return maker
 endfunction
 
-function! neomake#makers#ft#jsx#jsxhint()
+function! neomake#makers#ft#jsx#jsxhint() abort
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks!

I wonder if this is still required after https://github.com/neomake/neomake/pull/1124?

blueyed:
@matthewlehner 
Can you shed some light / try with current master?

Some of the changes could be kept for sure, but probably not all are needed.

matthewlehner:
@blueyed It looks like everything is working as it should! Javascript makers are now inherited in JSX files. Thanks!

I'll send a PR with the overlapping makers removed.

This does bring up a few questions for me about the default JSX makers. If I remove the JSX makers variable, the default makers are turned on, which, by default overlap with some of the default javascript makers. It's not a big deal, but may be a bit surprising. I'd suggest that because of the overlap, the makers should be turned off by default.

matthewlehner:
@blueyed I've just tested this a bit more thoroughly and uncovered a few edge cases that _may_ not be worth exploring…

Assuming the following config:
```viml
let g:neomake_javascript_enabled_makers = ['eslint']
let g:neomake_jsx_enabled_makers = ['flow']
```

Without flow as a registered jsx maker, this doesn't work. That said, it works fine when added to the javascript enabled makers. Is this the expected behaviour?

blueyed:
> Is this the expected behaviour?

Hard to say for me.
`flow` is not available as `neomake#makers#ft#jsx#flow()`, but could be looked up through `neomake#makers#ft#jsx#SupersetOf()` (which returns `javascript`).
So it could be made working.

blueyed:
Check out https://github.com/neomake/neomake/pull/1131.

blueyed:
Especially the changes to the tests: https://github.com/neomake/neomake/pull/1131/files#diff-039ea425629da9b3a95cdda80639f90e et al.

blueyed:
Merged now, so please revisit.

matthewlehner:
@blueyed Cool, this is great! Thank you so much!

Solves all my problems, and looks like it's cleaned up a lot of the JS makers as well.

blueyed:
I am glad it worked out for you!

PR #1123, opened 2017-03-16
Version of cargo maker update in #942 that uses #1050

JelteF:
I tried to use the newly added function in #1050. But I'm running into issues
with the bufnr. Right now create a new buffer for each file that is found if a
buffer does not exist already. 

This works fine, but it will also be shown when calling buffers. Which is
different from the previous behaviour, there it was hidden.
I found that I can set the buffer to hidden with `:setlocal nobuflisted`. What
do you think is the best way to set this for the newly created buffer, e.g.
switching between current and new buffer very shortly.


JelteF:
The tests have not been updated yet and will fail.

blueyed:
@JelteF 
#1050 is ready, a review would be nice - especially with regard to the doc: https://github.com/neomake/neomake/pull/1050/files#diff-0209e2cd5670892a1f78ee3c87ffcf37.

JelteF:
HI @blueyed, the docs look good, they would have helped with my first version of this branch. 

I just rebased onto the current master, but I'm running into a bug that the errors that are generated are always shown in the current buffer. This was not the case with the version based on the earlier version of #1050, so there's probably a regression here. 

As an example please look at this branch of a library of mine: https://github.com/JelteF/derive_more/tree/neomake-issue 
I added an error on purpose in `src/add_assign_like.rs`. If you open any other rust file it will show the error as well (.e.g. `src/mul_like.rs`), even though I supply both the bufnr and the filename in my entries.

blueyed:
@JelteF 
How do I make the error show up?
`cargo` does not seem to report it:
```
22:44:44 [D +12.6] [5.-:1] Running makers: cargo
22:44:44 [V      ] [5.-:1] Starting async job: ['cargo', 'test', '--no-run', '--message-format=json', '--quiet']
22:44:44 [D      ] [5.3:1] exit: cargo: 0
22:44:44 [D      ] [5.3:1] cargo: completed with exit code 0.
22:44:44 [D      ] [5.3:1] Cleaning jobinfo
22:44:44 [D      ] Skipping User autocmd NeomakeJobFinished: no hooks.
22:44:44 [D      ] File-level errors cleaned in buffer 1
22:44:44 [V      ] [5.3:1] Calling User autocmd NeomakeFinished with context: {'jobinfo': '…', 'file_mode': 1, 'make_id': 5}
```

I assume I need to configure/run something first?

JelteF:
@blueyed can you reopen this? it seems like you closed it by accedinet because of github syntax.

blueyed:
Sure, sorry - it only fixed the comment, not the PR.. ;)

blueyed:
@JelteF 
Do you think this is ready apart from the tests?
The tests should be quite simple: get the cargo maker and change it to echo the output you want to test, i.e. use `printf` for `exe` and the output with `args`.

blueyed:
You might want to leave out the highlight assertions, which are rather ugly and different for neovim/vim.
(this would be easier after https://github.com/neomake/neomake/issues/1146).
Just do it in one place maybe - I found the best to monkeypatch it:
```
function! neomake#highlights#AddHighlight(entry, type) abort
    call add(g:neomake_test_called, [a:entry, a:type])
endfunction
```

tests/ft_rust.vader
```diff
@@ -53,7 +53,7 @@ Execute (cargo error message):
...
...,"text":"            let ast = syn::parse_macro_input(&s).unwrap();"},{"highlight_end":70,"highlight_start":1,"text":"            $mod_::expand(&ast, stringify!($trait_)).parse().unwrap()"},{"highlight_end":10,"highlight_start":1,"text":"        }"},{"highlight_end":6,"highlight_start":1,"text":"    }"},{"highlight_end":3,"highlight_start":1,"text":");"}]},"macro_decl_name":"create_derive!","span":{"byte_end":579,"byte_start":539,"column_end":41,"column_start":1,"expansion":null,"file_name":"src/lib.rs","is_primary":false,"label":null,"line_end":25,"line_start":25,"suggested_replacement":null,"text":[{"highlight_end":41,"highlight_start":1,"text":"create_derive!(from, From, from_derive);"}]}},"file_name":"src/lib.rs","is_primary":true,"label":"expected str, found struct `proc_macro::TokenStream`","line_end":19,"line_start":19,"suggested_replacement":null,"text":[{"highlight_end":48,"highlight_start":46,"text":"            let ast = syn::parse_macro_input(&s).unwrap();"}]}]},"package_id":"derive_more 0.4.0 (path+file:///home/jelte/fun/derive_more)","reason":"compiler-message","target":{"kind":["proc-macro"],"name":"derive_more","src_path":"/home/jelte/fun/derive_more/src/lib.rs"}}']
```
blueyed: Feel free to put it into a separate file in `tests/fixtures/`.

JelteF: Good one, I'll  do that and then the pull request should be finished.

JelteF:
Added the files inside fixtures. The tests now pass locally for me. Once travis is working again it should be possible to merge this.

JelteF:
@blueyed If you're online can you cancel the builds except for the last one? Then it should be green sooner.

autoload/neomake/makers/ft/rust.vim
```diff
@@ -22,3 +22,70 @@ function! neomake#makers#ft#rust#rustc() abort
...
+        \ 'process_output': function('neomake#makers#ft#rust#CargoProcessOutput'),
+        \ }
+endfunction
+
+function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
+    let l:errors = []
+    for l:line in a:context['output']
+        if l:line[0] !=# '{'
+            continue
+        endif
```
blueyed: This does not work in case cargo would not provide a json dump per line, right?
Shouldn't you join all lines/output, and then decode those?

autoload/neomake/makers/ft/rust.vim
```diff
@@ -22,3 +22,70 @@ function! neomake#makers#ft#rust#rustc() abort
...
+
+        let l:decoded = neomake#utils#JSONdecode(l:line)
+        let l:data = get(l:decoded, 'message', -1)
+
+        if type(l:data) != type({}) ||  !len(l:data['spans'])
```
blueyed: Extra space before `!len` here.

autoload/neomake/makers/ft/rust.vim
```diff
@@ -22,3 +22,70 @@ function! neomake#makers#ft#rust#rustc() abort
...
+            let l:error.nr = l:code_dict['code'][1:]
+        endif
+
+        let l:error.text = l:data.message
+        let l:span = l:data.spans[0]
+        let l:error.filename = l:span.file_name
+
+        let l:error.bufnr = bufnr(l:error.filename)
+        if l:error.bufnr == -1
+            execute 'badd ' . substitute(l:error.filename, ' ', '\\ ', 'g')
```
blueyed: Missing the `TODO` we talked about here.

autoload/neomake/makers/ft/rust.vim
```diff
@@ -22,3 +22,70 @@ function! neomake#makers#ft#rust#rustc() abort
...
+            let l:error.text = l:error.text . ': ' . l:detail
+        elseif len(l:children) && has_key(l:children[0], 'message')
+            let l:error.text = l:error.text . '. ' . l:children[0].message
+        endif
+
+        call add(l:errors, l:error)
+
+    endfor
+    return l:errors
+endfunction
```
blueyed: I find all those `l:` prefixes hard to read.

tests/fixtures/rust/cargo_error.json
```diff
@@ -0,0 +1 @@
...:1,"text":"            let ast = syn::parse_macro_input(&s).unwrap();"},{"highlight_end":70,"highlight_start":1,"text":"            $mod_::expand(&ast, stringify!($trait_)).parse().unwrap()"},{"highlight_end":10,"highlight_start":1,"text":"        }"},{"highlight_end":6,"highlight_start":1,"text":"    }"},{"highlight_end":3,"highlight_start":1,"text":");"}]},"macro_decl_name":"create_derive!","span":{"byte_end":579,"byte_start":539,"column_end":41,"column_start":1,"expansion":null,"file_name":"src/lib.rs","is_primary":false,"label":null,"line_end":25,"line_start":25,"suggested_replacement":null,"text":[{"highlight_end":41,"highlight_start":1,"text":"create_derive!(from, From, from_derive);"}]}},"file_name":"src/lib.rs","is_primary":true,"label":"expected str, found struct `proc_macro::TokenStream`","line_end":19,"line_start":19,"suggested_replacement":null,"text":[{"highlight_end":48,"highlight_start":46,"text":"            let ast = syn::parse_macro_input(&s).unwrap();"}]}]},"package_id":"derive_more 0.4.0 (path+file:///home/jelte/fun/derive_more)","reason":"compiler-message","target":{"kind":["proc-macro"],"name":"derive_more","src_path":"/home/jelte/fun/derive_more/src/lib.rs"}}
```
blueyed: Good to have it in a file, but my intention was to have it more readable there then.. :D
This would then also test the case where cargo might not put them on a single line I assume?!

tests/ft_rust.vader
```diff
@@ -0,0 +1,153 @@
+Include: include/setup.vader
+
+Execute (cargo skip non json line):
+  runtime autoload/neomake/highlights.vim
+
+  file src/lib.rs
```
blueyed: better to use `new` before, and `bwipe` after - to not mess with buffers having a name etc.

tests/ft_rust.vader
```diff
@@ -0,0 +1,153 @@
...
+
+Execute (cargo error message):
+  runtime autoload/neomake/highlights.vim
+
+  "Monkey patch to check setting of length
+  Save g:lengths
+  let g:lengths = []
+  function! neomake#highlights#AddHighlight(entry, type) abort
+      call add(g:lengths, a:entry.length)
+  endfunction
```
blueyed: This needs to be undone:
```vim
runtime autoload/neomake/highlights.vim
```

blueyed:
Only seems to fail on `neovim` for a few builds?!

> Vader error: Vim(buffer):E86: Buffer 3 does not exist (in function <SNR>24_vader[5]..vader#run[51]..<SNR>29_run[32]..vader#window#prepare[1]..<SNR>32_switch_to_workbench, line 2)

I can investigate in that (tomorrow or the day after probably).

So for now, you might just want to address my comments from above.

autoload/neomake/makers/ft/rust.vim
JelteF: Cargo does provide a json dump per line.

autoload/neomake/makers/ft/rust.vim
JelteF: They are required by vint right.

autoload/neomake/makers/ft/rust.vim
blueyed: I see, it's more too be more robust.
Feel free to ignore it for now though.

autoload/neomake/makers/ft/rust.vim
blueyed: No..

We have a `.vintrc.yaml` to ignore those:
```
policies:
  ProhibitImplicitScopeVariable:
    enabled: false
```

blueyed: Isn't that use for you?

JelteF:
I just fixed all the comments as far as I can tell. If you can take a look at why the nvim test fails that would be great. Locally I'm running neovim as well and the test works there, so I'm not sure what's going on.

tests/ft_rust.vader
```diff
@@ -119,35 +119,35 @@ Execute (cargo warning message):
...
-    \ 'pattern': '',
-    \ 'text': 'custom derive attribute panicked. message: Only structs and enums can derive From'}], getloclist(0)
-
-  AssertEqual [4], g:lengths
-  bwipe
+" Execute (cargo error message children):
+"   "Monkey patch to check setting of length
+"   Save g:lengths
+"   let g:lengths = []
+"   function! neomake#highlights#AddHighlight(entry, type) abort
```
blueyed: I guess the monkeypatch needs to be undone.
I can look at it tomorrow.

JelteF: I commented the monkey patch and that indeed fixes the issue. I just change it to not monkey patch in just nvim. If that works it's fine by me if you merge it like that. Although if you can think of a nicer solution that's obviously better.

blueyed:
Working on it now.

blueyed:
Let's see if https://github.com/neomake/neomake/pull/974 helps.

blueyed:
Squashed.
Will review it again, but then it seems to can finally land soonish.

blueyed:
Bam! 🎆

jonhoo:
Nice! Out of curiosity though, why is this a file maker, not a project maker? When I run `:Neomake`, I get no errors, even though I intentionally introduce them into the current file...

jonhoo:
Never mind — `cargo test` was taking forever to run, but didn't quit when I quit neovim. When I then started neovim again, running `cargo test` would block on the previous invocation. Adding back my `timeout 5s` fix resolves the issue.

PR #1153, opened 2017-03-22
s:ProcessEntries: use bufnr from entries if present

blueyed:
Fixes https://github.com/neomake/neomake/pull/1123#issuecomment-288536452.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -980,7 +980,7 @@ function! s:ProcessEntries(jobinfo, entries, ...) abort
     endif
 
     let counts_changed = 0
-    let cleaned_signs = 0
+    let cleaned_signs = []
     let ignored_signs = []
     let maker_type = file_mode ? 'file' : 'project'
     let do_highlight = get(g:, 'neomake_highlight_columns', 1)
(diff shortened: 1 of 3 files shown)


JelteF:
Seems like a fix that makes sense.

JelteF:
It's probably nice to move this code from #1123 to the mapping code:
```vim
        let l:error.bufnr = bufnr(l:error.filename)
        if l:error.bufnr == -1
            execute 'badd ' . l:error.filename
            let l:error.bufnr = bufnr(l:error.filename)
        endif
```

blueyed:
@JelteF 
maybe, yes - but is it required after all?  Wouldn't the filename be there still, and the buffer opened/added when selecting this list entry?

JelteF:
I tried that first. But the difference is that then you have to run the the maker again after opening the file, otherwise the signs are not added. The other version opens a buffer automatically for each found file it seemed.

blueyed:
> otherwise the signs are not added

Ah, that's another issue with signs then I would say: they should be queued and added once the buffer gets opened.

blueyed:
@JelteF 
I think it's fine to keep in #1123 for now, but should be marked as TODO, with a reference to here, or explaining it there.

JelteF:
Sounds like a good approach

PR #1156, opened 2017-03-23
dict based config

blueyed:
This will lookup config settings in b:config, g:config, and has support
for `context` (for filetype and bufnr) (where `context.maker` is also
looked at in the end).

The idea is to prefer the new format, and fallback to looking up the current scheme.
I am not sure if it should be deprecated really, since it allows for easier one-shot configs (`let g:neomake_someft_maker_args` vs `g:neomake['ft:someft']['maker']['args']`, where you have to initialize the dict(s).
`g:neomake['ft:someft']['maker']['args']` is not really supported in this PR as of now.

My main intention for this is with the automaking, where I would like to have some config namespace, i.e. `g:neomake.automake` would allow to override other settings in the `automake` context.

TODO:
 - [x] ensure that it behaves like neomake#utils#GetSetting
 - [ ] doc

### DIFF
--- a/autoload/neomake/config.vim
+++ b/autoload/neomake/config.vim
@@ -1,0 +1,147 @@
+let g:neomake#config#undefined = {}
+
+" Resolve a:name (split on dots) and init a:dict accordingly.
+function! s:resolve_name(dict, name) abort
+    let c = a:dict
+    let parts = split(a:name, '\.')
+    for p in parts[0:-2]
+        if !has_key(c, p)
+            let c[p] = {}
+        endif
+        let c = c[p]
+    endfor
+    return [c, parts[-1]]
+endfunction
+
+" Get a:name (resolved; split on dots) from a:dict, using a:context.
+function! s:get(dict, name, context) abort
+    let ft = has_key(a:context, 'ft') ? a:context.ft : &filetype
+    let parts = split(a:name, '\.')
+    let prefixes = ['']
+    if !empty(ft)
+        call insert(prefixes, 'ft.'.ft.'.', 0)
+    endif
+    for prefix in prefixes
+        let [c, k] = s:resolve_name(a:dict, prefix.join(parts[0:-1], '.'))
+        if has_key(c, k)
+            return c[k]
+        endif
+    endfor
+    return g:neomake#config#undefined
+endfunction
+
+" Get a:name from config.
+" Optional args:
+"  - a:1: default
+"  - a:2: context
+function! neomake#config#get(name, ...) abort
+    let Default = a:0 ? a:1 : g:neomake#config#undefined
+    let context = a:0 > 1 ? a:2 : {}
+    if a:name =~# '^b:'
+        if !has_key(context, 'bufnr')
+            let context.bufnr = bufnr('%')
+        endif
+        let name = a:name[2:-1]
+    else
+        let name = a:name
+    endif
+    let bufnr = has_key(context, 'bufnr') ? context.bufnr : bufnr('%')
+
+    for lookup in [
+                \ getbufvar(bufnr, 'neomake'),
+                \ get(t:, 'neomake', {}),
+                \ get(g:, 'neomake', {}),
+                \ get(context, 'maker', {})]
+        if !empty(lookup)
+            let R = s:get(lookup, name, context)
+            if R isnot# g:neomake#config#undefined
+                return R
+            endif
+        endif
+        unlet! lookup R  " old vim
+    endfor
+    return Default
+endfunction
+
+" Get a:name from config with information about the setting's source.
+" This is mostly the same as neomake#config#get, but kept seperate since it
+" is not used as much (?!).
+" Optional args:
+"  - a:1: default
+"  - a:2: context
+function! neomake#config#get_with_source(name, ...) abort
+    let Default = a:0 ? a:1 : g:neomake#config#undefined
+    let context = a:0 > 1 ? a:2 : {}
+    if a:name =~# '^b:'
+        if !has_key(context, 'bufnr')
+            let context.bufnr = bufnr('%')
+        endif
+        let name = a:name[2:-1]
+    else
+        let name = a:name
+    endif
+    let bufnr = has_key(context, 'bufnr') ? context.bufnr : bufnr('%')
+
+    for [source, lookup] in [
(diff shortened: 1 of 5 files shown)


autoload/neomake/config.vim
```diff
@@ -0,0 +1,27 @@
+let s:undef = {}
+
+function! neomake#config#get_setting(name, default, ...) abort
+  let context = a:0 ? a:1 : {}
+  let ft = has_key(context, 'ft') ? context.ft : &filetype
+  let bufnr = has_key(context, 'bufnr') ? +context.bufnr : bufnr('%')
```
tweekmonster: The `+var` thing has never occurred to me before, but it's genius 😀

autoload/neomake/config.vim
```diff
@@ -0,0 +1,27 @@
+let s:undef = {}
+
+function! neomake#config#get_setting(name, default, ...) abort
```
tweekmonster: I think this would be better as `neomake#config#get()`

tweekmonster:
I'm not sure what I think of this yet.  It does deal with the issue of long and numerous config variables, but in exchange for complexity.  It's not bad as long as the complexity isn't passed to the user.

The passing of a context is a little weird, too.  Couldn't these dicts be wrapped to have some helper functions?  Like `maker.get()`?  It could abstract the problem with nested dicts with something like `maker.get('foo.bar.baz', 'default')` or `maker.set('foo.bar.baz', 'qux')`.  Or, `neomake#config#get('foo.bar.baz', 'default')`.

Users could still use:

```vim
let g:neomake = {
\   'foo': {'bar': {'baz': 'qux'}},
\   'python': {
\     'enabled_makers': ['flake8'],
\     'flake8': {'ignore': ['E501']},
\   },
\ }

call neomake#config#set('foo', {'bar': {'baz': 'qux'}})
let g:bar = neomake#config#get('foo.bar')
```

The key lookup would be backward compatible so that `foo.bar.baz` -> `g:neomake_foo_bar_baz`.  Then deprecation could just be done by documenting the new config style and not mentioning the old style.

autoload/neomake/config.vim
blueyed: It's from tpope.. :)

blueyed:
Thanks for your feedback again!

> in exchange for complexity

What I want to have here is to also handle extra contexts, like e.g. when automaking (#1167) is used vs. manual invocation: the context will be there in general (because it should also have an influence on error level reporting - some things should be displayed as errors when manually invoked, but not when automatically linting  - ref: https://github.com/neomake/neomake/issues/1165).

`neomake#config#get('foo.bar.baz', 'default')` is a good idea indeed (especially for setting).

But then it's still a matter of context(s): currently this is just assumed as a prefix (and even `b:` settings expect a `ft` prefix, i.e. `b:neomake_ft_foo_maker` - although the `ft` part should be assumed to be the one of the buffer after all).

Therefore my idea was to have certain prefixes like `automake` and `ft:&ft`, which would be looked up in the dict before.
The `ft:` prefix is meant to distinguish it from other global settings.

tweekmonster:
Maybe I'm not yet seeing the idea like you are.  From my perspective, `neomake#config#get('python.enabled_makers')` is just as isolated as `ft:python.enabled_makers` would be.  Are you saying that `ft:` would be for built-in Neomake settings?

`neomake#config#get('automake.python.enabled_makers')` would naturally fit into the dict:

```vim
let g:neomake = {
\   'automake': {
\     'python': {
\       'enabled_makers': ['flake8'],
\     },
\   },
\ }
```

Thinking out loud: `automake:python` as a key is a weird detail.  The advantage I see for using it would be that someone couldn't accidentally reset everything with `let g:neomake.automake = {}`, but that would be user error (if unintentional).  The disadvantage is that you couldn't use `let g:neomake.automake:python.enabled_makers = ['flake8']`, which might be an annoyance.

blueyed:
> Maybe I'm not yet seeing the idea like you are. From my perspective, `neomake#config#get('python.enabled_makers')` is just as isolated as `ft:python.enabled_makers` would be.

It has the potential to conflict if a filetype has the same name as a setting.

This is the case already when using `automake` as the namespace for automake related settings (#1167).
`g:neomake.automake.enabled_makers` should override `g:neomake.enabled_makers` there in general, and `g:neomake.ft.automake.enabled_makers` would configure the makers for the `automake` filetype.
`g:neomake.automake.ft.automake.enabled_makers` would configure the makers for automaking in an`automake` file.
(I am using a sub-dict `ft` here for what was `ft:automake` before).

blueyed:
@tweekmonster 
Pushed some changes, how do you like them?

I would like to get this straight and then use it with #1167.

autoload/neomake/utils.vim
```diff
@@ -320,6 +320,16 @@ let s:unset = {}  " Sentinel.
 " Get a setting by key, based on filetypes, from the buffer or global
 " namespace, defaulting to default.
 function! neomake#utils#GetSetting(key, maker, default, ft, bufnr) abort
+  " Check new-style config.
+  " Add maker and bufnr to context only if g:neomake or b:neomake exist.
+  let context = {'ft': a:ft}
+  if exists('g:neomake') || !empty(getbufvar(a:bufnr, 'neomake'))
+    call extend(context, {'maker': a:maker, 'bufnr': a:bufnr})
+  endif
+  let Ret = neomake#config#get(a:key, g:neomake#config#undefined, context)
```
tweekmonster: `Ret` seems out of place.  Is it supposed to be capitalized?

blueyed: Yeah.. it can be a function reference.. :/

tweekmonster: Wouldn't that be an issue for the caller?  I thought that's only a problem if you attempt to call the function in the variable?

tweekmonster: I think this is a non-blocking issue either way.

blueyed: No, it happens when assigning the variable/funcref already.

blueyed: Thanks for your review, I plan to add some doc probably, and have #1167 rebased on it, and will use it there.

blueyed:
Squashed, and will merge it once it is green.

PR #1160, opened 2017-03-24
Improve error displaying for elm-make maker

keiththomps:
`context.output` returns each line from the output and the
`elm-make` output groups warnings in a JSON array on one line and
errors on the next line in a separate JSON array. This commit
iterates over the `context.output` so that we're actually showing
the errors instead of just warnings if both exist.


blueyed:
Thanks, merging although the build failed: I think the neovim nightly is broken.

Issue #745, opened 2016-10-28
Alternatives to autocmd for specifying when running makers


antoyo:
Hello.
It would be nice to have a smarter way to specify when to run `:Neomake` than using `autocmd`.

The issue with using `autocmd` is the duplicate work, specially when using a directory maker:
you don't want to run the directory maker on every file save because if you save many files at once (with e.g. `:wa`), neovim will freezes for minutes.

So I think we need a smarter way to handle the autorun of a maker.
I don't know what would be the best, thought.

This would also fixes issues like [this one](https://github.com/neomake/neomake/issues/309).

Thanks.

blueyed:
I think it's related to https://github.com/neomake/neomake/projects/1.

blueyed:
#1167 has been merged, this should address this.

Issue #405, opened 2016-05-11
Provide a wrapper / helper to run Neomake automatically


blueyed:
Currently the README suggests the following to enable automatic checking:

```
autocmd! BufWritePost * Neomake
```

This is a bit brittle though, e.g. it will also run on `:w /some/other/path`, but on the current file.

I've created https://github.com/neomake/neomake/issues/404 as a first step into this direction.

wsdjeg:
I also get this issue,should we use `<afile>` instead? also `:wa` command should not only check current buffer.
BTW we should provid a g:var for user do not run maker when use  `:wq` command!

what do you think?

blueyed:
Yes, we should provide some function for this.

Here is what I am currently using: https://github.com/blueyed/dotfiles/commit/0c69796bb44674403ddb240be3475f04e0be9934.

`expand("<afile>")` seems to refer to the relative file, while `expand("<amatch>")` refers to the full path always.

blueyed:
> BTW we should provid a g:var for user do not run maker when use :wq command!

That is not so easy I think, but we might be able to use `QuitPre` here and mark the window/buffer as not going to be processed.

wsdjeg:
I am reading syntasc source code,and try to find a way to implement it.

blueyed:
Something like this maybe? https://github.com/blueyed/dotfiles/commit/4c59098cc42766f61f7f67169fdf8d90bb8437cd

blueyed:
#1167

blueyed:
Closing in favor of #1167.

PR #1167, opened 2017-03-29
[WIP] Automaking

blueyed:
This is meant to configure Neomake in a sane way, and also provides support for "autolinting", i.e. on events like `TextChanged` etc.  This would replace the previous instructions to configure Neomake for `BufWritePost` etc manually.
(Followup of #1055)

You can use `call neomake#configure#autolint('niw', 500)` to configure it in
normal and insert mode with a delay of 500ms, and to setup a `BufWritePost` autocommand.

TODO:
 - [ ] allow config overrides based on automake mode (by refactoring settings in general (dict based)) (https://github.com/neomake/neomake/pull/1156)
 - [ ] short-circuit maker init/setup e.g. while in the same mode / settings were not changed (performance); basically try to just run job_start again. - XXX: might be done?!
 - [ ] explicitly cancel jobs when restarting, instead of letting Neomake handle it?! - XXX: done, or in stash
 - [x] enhancements to tempfiles (https://github.com/neomake/neomake/pull/1252 - based on it already)
 - [ ] doc

Related/fixes issues:
 - https://github.com/neomake/neomake/issues/745
 - https://github.com/neomake/neomake/issues/405
 - https://github.com/neomake/neomake/issues/441

/cc @dojoteef


blueyed:
For anyone lurking: this should be usable now again (even though not all tests pass currently).

dojoteef:
@blueyed Hey, sorry I've been out of contact for a while. I have been quite busy lately. Thanks for helping to respond to people regarding [Neomake Autolint] in my absence.

* Should I bother making changes/fixes to [Neomake Autolint], or do you expect this project to be a replacement for it?
* Is there any help you want with regards to this project?

Just a heads up that things still have not quite settled down for me yet, so I may still be slow in responding.

[Neomake Autolint]: https://github.com/dojoteef/neomake-autolint

blueyed:
@dojoteef 
> or do you expect this project to be a replacement for it?

Yes.

> Is there any help you want with regards to this project?

Testing it, i.e. using this branch.
I am using it myself, and rebase it against master regularly.

Should be quite good already, but e.g. missing docs.

blueyed:
Promise: I will merge this on Thursday - even without docs or any feedback.. 💃

Alok:
@blueyed Any word on when this will be merged in?

blueyed:
@Alok 
I am using this most of the time myself, and rebase it often against master.
It should work well, but is missing docs, cleanup and feedback.

blueyed:
(disregard my previous enthusiastic comment about having it merged already by now.. there is too much on my plate, and regarding Neomake I'd rather want to stabilize it before bringing this).

bounceme:
if this is based on timers, you could save one temp creation when not `&modified`

blueyed:
@bounceme 
Sorry, I don't understand. Please elaborate.

bounceme:
https://github.com/neomake/neomake/blob/305d587154bd4470acb06b4ead08b1979f0753bd/autoload/neomake.vim#L408

could work with `%:p` if the current buffer is unmodified

blueyed:
The buffer needs to be modified, empty (no name) or unreadable for temp files to be used.
https://github.com/neomake/neomake/blob/305d587154bd4470acb06b4ead08b1979f0753bd/autoload/neomake.vim#L438
Temp files are currently also still off by default, but very useful!

blueyed:
Ah.. you mean to check/ensure this when the timer times out only?

bounceme:
`Ah.. you mean to check/ensure this when the timer times out only?`

if there are multiple timers concurrently, or the check is being run against a previous version of the buffer. I haven't taken a good look at the code here though.

I took the `text changed` approach when i tried to implement automake.

bounceme:
I am just throwing my impressions out here. Is polling happening only when the buffer is modified? Wouldn't a single , global, variable job request be more efficient? Or is that already what happens? i ask because it seems like there are several caches of timers for starting and killing processes

blueyed:
Thanks for your interest.
IIRC there is a timer per buffer.
TextChanged is used also already, but can trigger a timer.
In https://github.com/neomake/neomake/pull/1370 I have started working to improve the generic action queue, which could be used here also, e.g. when timing out in another buffer: it should not just restart the timer over and over again, but rather only when changing to another window/buffer.

The best might be to just give it a try with `g:neomake_logfile = '/tmp/neomake.log'` and `tail -f /tmp/neomake.log` while editing.

blueyed:
@davidosomething 
Since you have been working on the README/help: what do you think about cutting the README down like this?
I think that #1452 should clear it up / explain it a bit more afterwards, but also mainly in the `doc/neomake.txt` region then probably.

blueyed:
:fireworks:

gunar:
This is awesome. Thank you, Daniel!

blueyed:
@gunar 
You're welcome.
How do you use it?

btw: I hope I can finish #924 soon, which is also really nice!

blueyed:
@gunar 
i.e.. what is your Neomake config?

gunar:
@blueyed
I wanted this mostly so I could lint while typing,
but realized I exit Insert Mode so often that it makes little difference.
Also, you must know I use [907th/vim-auto-save](/907th/vim-auto-save).

[My current config](https://github.com/gunar/dotfiles/blob/ed68dcabd3419d8d169ca86f742dd56a389befcf/nvim/settings/linting.vim).

blueyed:
@gunar 
vim-auto-save is interesting.. therefore having BufWritePost in your case is likely to be enough..
But Neomake should handle that and e.g. skip events when the buffer was not changed.
So you might want to take a closer look at the logs to see if it really works as expected - i.e. does not call the makers too often.

gunar:
@blueyed Oh but I actually want to auto-save (works wonders for development). Are you saying I could use neomake to auto-save?

blueyed:
> I actually want to auto-save (works wonders for development)

I also use Ctrl-S to leave insert mode and save most of the time.
Therefore triggering Neomake on BufWritePost was good for me in general, too.

> Are you saying I could use neomake to auto-save?

No, but triggering it on save is all you need basically then - like you've realized yourself.. ;)

hoschi:
I have it running for some days now, and in general it is really great :+1: 

One thing is the locale temp files which are created right by the source files. I already told four tools or so to ignore hidden JS files, but sometimes it is hard to do that. Right now I have one with I can't fix by myself (eslint plugin) and another one where I already changed a regex to ignore hidden files but it still watches them.

Would it be possible to configure a dedicated path to store the files in it? It even an replicate the whole path for the source file. So when you edit `/home/myUser/repos/foo/bar.js` it and my neomake path is `/tmp/neomake/` it creates the file `/tmp/neomake/home/myUser/repos/foo/bar@neomake_4654-2.js`. With this a user must configure one option, instead of all others.

Another thing I noticed but not yet looked into is: What is the working directory of the linter when it runs? With `:Neomake` all was fine and e.g. stylelint picked up the `.stylelintrc`. With auto lint it seems it dosen't find it anymore.

hoschi:
> Are you saying I could use neomake to auto-save?

Does neomake implements events? With this someone would then be able to configure "save after linter run and there were no errors"? This would be the first "auto save" option which makes sense, in my opinion. Why I want to save a file which is has broken code in it when I leave the window :grinning:

blueyed:
> Would it be possible to configure a dedicated path to store the files in it? It even an replicate the whole path for the source file.

The problem with that is that config files (for the linters) might be missing then.
Should become a new issue (for discussion) - please create one with your suggestions from above.

> What is the working directory of the linter when it runs?

The current working directory by default (IIRC), but some makers specify a custom one.

> With :Neomake all was fine and e.g. stylelint picked up the .stylelintrc. With auto lint it seems it dosen't find it anymore.

That would be a bug then - please create a new issue for this.

> Does neomake implements events?

Yes, there are `NeomakeFinished` and `NeomakeJobFinished` user autocommands.

> save after linter run and there were no errors

Interesting idea - might become an option/setting even, or something in a recipes section!

dojoteef:
Great job @blueyed! I'm glad you were able to check this into the mainline of Neomake. Unfortunately, I've been a bit preoccupied so I haven't had much opportunity to think about autolinting/automaking. Glad you were able to pick up my slack :-D

hoschi:
@blueyed I saw you opened #1593 and I created #1595 for the config stuff. I'll check my setup for the stylelint thing if I think my setup is correct.

PR #1177, opened 2017-04-01
tests: cleanup/improve vimrc

blueyed:
This sets 'noloadplugins' to skip loading of global plugins, but only
user-defined ones (in case HOME would not be set to a temporary dir).

Vader and Neomake are then sourced manually.


blueyed:
In the long run this should get improved after https://github.com/tweekmonster/vim-testbed/issues/8#issuecomment-252536288 is fixed.

PR #1184, opened 2017-04-04
debug: improve msg with "Nothing to make"

blueyed:


maxime-michel:
Please decline ASAP, the ridiculously long branch name makes pulling fail on my encrypted Ubuntu, and probably on Windows as well.

blueyed:
@maxime-michel 
Sorry, hope that helps?!

blueyed:
@maxime-michel 
Out of curiosity: do you use encfs?
it also depends on the prefix probably, i.e. where you clone your plugin?
And then: why is this branch/ref being fetched for you in the beginning?

maxime-michel:
Thanks, yes it does. I had resolved the issue already by deleting the repo and cloning it again, since `core.longpaths` was in my gitconfig, but had been added after the initial cloning of neomake, and was therefore not taken into account (?).

The ref got fetched when vim-plug pulled, and git tried to save a file named after it.

Yes, I'm using encfs, which can be awkward at times. But in this case I don't think anybody wanted to see such a branch long name in the project's Git history, nor on GitHub. It's adding an horizontal scrollbar to this page's layout on my 13-inch display, for instance.

blueyed:
> The ref got fetched when vim-plug pulled

I would have expected it to only fetch the relevant (i.e. `master`) branch?!

> anybody wanted to see such a branch long name in the project's Git history,

Would have been squash-merged anyway.
I get your point though.  Sorry again.

Issue #1191, opened 2017-04-09
Expansion of args does not use temporary file


blueyed:
`neomake#utils#ExpandArgs` does not use any temporary file (when enabled and used).

This is OK for something like `%:h` where it should use the non-temporary file (which is usually in another directory), but we should provide a way to get to the actual (temporary) file being used.

I've thought about adding something like `%f` which would get expanded before any other expansions.
There is no way to use modifiers on it then though.

blueyed:
See https://github.com/neomake/neomake/pull/1192 for something that might help here.

blueyed:
Let's consider https://github.com/neomake/neomake/pull/1207 to be an appropriate / good enough fix for this for now.

PR #1192, opened 2017-04-09
[RFC] Expand %f in neomake#utils#ExpandArgs

blueyed:
Ref: https://github.com/neomake/neomake/issues/1191

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -376,7 +376,7 @@ function! s:command_maker_base._get_fname_for_buffer(jobinfo) abort
     return bufname
 endfunction
 
-function! s:command_maker_base._bind_args() abort dict
+function! s:command_maker_base._bind_args(jobinfo) abort dict
     " Resolve args, which might be a function or dictionary.
     if type(self.args) == type(function('tr'))
         let args = call(self.args, [])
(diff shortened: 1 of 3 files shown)


tweekmonster:
Hmm, I'm not sure if `%f` is the way to go for this, but I'm not against it.  I just want to point out that it might be semantically confusing.

`ExpandArgs()` takes `%` as a filename substitution as defined with Vim's `expand()` and `fnamemodify()`.  We use `%%` to prevent the expansion.  So, if an argument is passing a `printf()` style format string for floats, `%%f` would break in this case.

Maybe it'll be easier and more future-proof to set environment variables?  For example use `$NEOMAKE_TEMPFILE` in the arguments to let the system call sort it out?

blueyed:
> Maybe it'll be easier and more future-proof to set environment variables?

Very nice idea, I'll change it.

blueyed:
Not really sure how to approach this though: my current idea would be to `let $NEOMAKE_TEMPFILE` before starting the jobs, and unletting it afterwards again?!
Also it should be `NEOMAKE_FILENAME` probably (since it is about the file to be used, which usually is not a temporary file).

blueyed:
Closing this in favor of https://github.com/neomake/neomake/pull/1207.

PR #1194, opened 2017-04-09
Add purescript maker

vaibhavsagar:
Instead of #359 this is a simpler PureScript maker based on the Elm maker. This is pretty barebones at the moment but I'd love any and all feedback!

### DIFF
--- a/autoload/neomake/makers/ft/purescript.vim
+++ b/autoload/neomake/makers/ft/purescript.vim
@@ -1,0 +1,63 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#purescript#EnabledMakers() abort
+    return ['pulp']
+endfunction
+
+function! neomake#makers#ft#purescript#pulp() abort
+    " command is `pulp build --no-psa -- --json-errors`
+    " as indicated in https://github.com/nwolverson/atom-ide-purescript/issues/136
+    let maker = {
+        \ 'args': ['build', '--no-psa', '--', '--json-errors'],
+        \ 'append_file': 0,
+        \ 'process_output': function('neomake#makers#ft#purescript#PSProcessOutput'),
+        \ }
+
+    " Find project root, since files are reported relative to it.
+    let bower_file = neomake#utils#FindGlobFile('bower.json')
+    if !empty(bower_file)
+        let maker.cwd = fnamemodify(bower_file, ':h')
+    endif
+
+    return maker
+endfunction
+
+function! neomake#makers#ft#purescript#PSProcessOutput(context) abort
+    let l:errors = []
+    for line in a:context.output
+        if line[0] !=# '{'
+            continue
+        endif
+        let l:decoded = neomake#utils#JSONdecode(line)
+        for [key, values] in items(l:decoded)
+            let l:code = key ==# 'warnings' ? 'W' : 'E'
+            for item in values
+                let l:compiler_error = item['errorCode']
+                let l:message = item['message']
+                let l:position = item['position']
+                let l:filename = item['filename']
+                if  l:position is g:neomake#compat#json_null
+                    let l:row = 1
+                    let l:col = 1
+                    let l:end_col = 1
+                    let l:length = 1
+                else
+                    let l:row = l:position['startLine']
+                    let l:col = l:position['startColumn']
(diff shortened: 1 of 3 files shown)


autoload/neomake/makers/ft/purescript.vim
```diff
@@ -0,0 +1,59 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#purescript#EnabledMakers()
+  return ['pulp']
+endfunction
+
+function! neomake#makers#ft#purescript#pulp()
+    return {
+        \ 'exe': 'pulp',
+        \ 'args': ['build', '--no-psa', '--', '--json-errors'],
```
blueyed: `exe` defaults to `pulp`.
Is `--` required in `args` really?  (typically it's used to separate options from filenames, so `--json-errors` would end up being considered a filename?!)

vaibhavsagar: In this case it's used to separate options to `pulp` from options to `purs compile` which it uses under the hood.

blueyed: I see.  Should have a comment then.

vaibhavsagar: You're right, I've added a comment and a link to the issue where I got this information.

blueyed:
LGTM, should have some test though.
See `tests/ft_rust.vader`.

vaibhavsagar:
Noted, I will look into adapting an existing test.

blueyed:
@vaibhavsagar 
ping

vaibhavsagar:
Hey, thanks for the ping! I got busy with life stuff but will look at this over the weekend 😄

vaibhavsagar:
Still a work in progress!

tests/ft_purescript.vader
```diff
@@ -0,0 +1,48 @@
...
+    \ ]
+
+  let result = neomake#makers#ft#purescript#pulp().process_output({'output': output})
+
+  AssertEqual [], result
+
+Execute (pulp parses error message):
+  new
+  file src/Main.purs
+  file src/Other.purs
```
blueyed: Don't call `file` twice on the same buffer, but should call `new` inbetween maybe?!

blueyed:
Pushed a fixup to fix the indent issue.

What do you plan to do additionally?

vaibhavsagar:
I also need to test the warnings.

tests/ft_purescript.vader
vaibhavsagar: Thanks. This is my first ever Vader test, can you tell :smile:

blueyed:
I will look into the bufnr stuff. It seems to be the same as with https://github.com/neomake/neomake/pull/1218.

blueyed:
https://github.com/neomake/neomake/pull/1257 should address this.

btw: I think there is a conceptual error in using a file_mode maker, when in fact it should be a project-based one: file_mode makers use location lists, while otherwise a quickfix list is used.  I am not really sure about how to address this, and think the whole idea of `file_mode` needs to be revisited.

vaibhavsagar:
Thanks for looking into this! I agree that having this as a project-based maker would make more sense. I will look into it.

blueyed:
It is more a general, conceptual issue I think.
E.g. `cargo` was switched to a file_mode maker recently, but I am not sure what the reason was.

blueyed:
Pushed another commit to use the project root in `cwd`, since the output is relative to it.

Needs more fixed to https://github.com/neomake/neomake/pull/1257 though really.

vaibhavsagar:
You've put much more work into this than I have :smile:. I've added a test for warnings.

blueyed:
Merged master, which contains #1257.

Is this OK from your side?
Do you want to make it a project maker already?
My idea about that would be to allow for makers to live in `ft/`, but allow to specify if they are project makers (or rather if they want to use the quickfix or location list).

vaibhavsagar:
I'm happy with the file-based maker for now, I just wanted something simple to start off with and I'm happy for someone else (or future me) to change it to be project-based further down the line.

blueyed:
Thanks!

blueyed:

vaibhavsagar:
Done, thanks!

PR #1195, opened 2017-04-09
Migrate to cargo-check in cargo.vim maker by default

### LINKED ISSUE #843: Use cargo-check in cargo maker

fuine:
This PR introduces `cargo-check` as the default command run in the `cargo` maker. Users can set the command passed to `cargo` via specified option (`g:neomake_rust_cargo_command`). Most notably to enable test-checking support one would add this line to the .vimrc:
`let g:neomake_rust_cargo_command = ['test', '--no-run']`
This closes #843.

### DIFF
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -22,8 +22,9 @@ function! neomake#makers#ft#rust#rustc() abort
 endfunction
 
 function! neomake#makers#ft#rust#cargo() abort
+    let maker_command = get(g:, 'neomake_rust_cargo_command', ['check'])
     return {
-        \ 'args': ['test', '--no-run', '--message-format=json', '--quiet'],
+        \ 'args': maker_command + ['--message-format=json', '--quiet'],
         \ 'append_file': 0,
         \ 'errorformat':
             \ '[%t%n] "%f" %l:%v %m,'.
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -22,7 +22,8 @@ function! neomake#makers#ft#rust#rustc() abort
 endfunction
 
 function! neomake#makers#ft#rust#cargo() abort
-    let maker_command = get(b:, 'neomake_rust_cargo_command', ['check'])
+    let maker_command = get(b:, 'neomake_rust_cargo_command',
+                \ get(g:, 'neomake_rust_cargo_command', ['check']))
     return {
         \ 'args': maker_command + ['--message-format=json', '--quiet'],
         \ 'append_file': 0,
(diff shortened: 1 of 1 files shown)


fuine:
@blueyed I decided to use the list in order to allow users to use multiple arguments (the case with the `test --no-run` command). I'm not sure how the usage of :b would improve this PR - do you think that setting this option per buffer instead of globally would be useful in any way?

JelteF:
I think `:b` would definitly improve this PR. I would like to force usage of the `test` version when I'm editing my test files. When `:b` is used I can just do a simple regex to set the value for files in the tests directory.

fuine:
@JelteF that is a nice use case, changed.

autoload/neomake/makers/ft/rust.vim
```diff
@@ -22,7 +22,7 @@ function! neomake#makers#ft#rust#rustc() abort
 endfunction
 
 function! neomake#makers#ft#rust#cargo() abort
-    let maker_command = get(g:, 'neomake_rust_cargo_command', ['check'])
+    let maker_command = get(b:, 'neomake_rust_cargo_command', ['check'])
```
blueyed: Should rather look at both:
```diff
diff --git a/autoload/neomake/makers/ft/rust.vim b/autoload/neomake/makers/ft/rust.vim
index ec6ca94f..4892d247 100644
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -22,7 +22,8 @@ function! neomake#makers#ft#rust#rustc() abort
 endfunction
 
 function! neomake#makers#ft#rust#cargo() abort
-    let maker_command = get(b:, 'neomake_rust_cargo_command', ['check'])
+    let maker_command = get(b:, 'neomake_rust_cargo_command',
+                \ get(g:, 'neomake_rust_cargo_command', ['check']))
     return {
         \ 'args': maker_command + ['--message-format=json', '--quiet'],
         \ 'append_file': 0,
```

blueyed:
What do you think about documenting it?
Ideally it would change after #1156 anyway - but then we could also adjust the doc.

There is no maker specific documentation in doc/neomake.txt for know though.
Should it go into the maker list in the wiki (which eventually should move into doc/ (again)).

autoload/neomake/makers/ft/rust.vim
fuine: Changed

blueyed:
@fuine 
Sounds good to me: please add it to the wiki for now.

fuine:
As promised, the documentation comment. At the moment `cargo` maker will by default invoke `cargo-check` command. You can configure the exact command, along with its args via the `g:neomake_rust_cargo_command` and `b:neomake_rust_cargo_command` lists (e.g. to check everything, including test modules, put `let g:neomake_rust_cargo_command = ['test', '--no-run']` in your .vimrc file).

Thanks for help with this PR @blueyed @JelteF

Issue #1196, opened 2017-04-12
Migrate clippy maker to the json-based format


fuine:
`cargo-clippy` supports the json output and I think that migrating to it should be beneficial, if anything for the better errors/warnings highlights.

Issue #793, opened 2016-11-20
Cargo maker should show type hint for mismatched type errors


jonhoo:
When the `cargo` maker is run, the short error shown in the statusline is usually helpful. The exception is for type mismatches, where it will simply show `cargo: mismatched types`, without communicating what those types are. This was discussed briefly with @fuine in https://github.com/neomake/neomake/pull/613#issuecomment-261745615.

Consider the following rust program:
```rust
fn main() {
    let foo: bool = 42;
}
```

Compiling it gives the following error:
```
error[E0308]: mismatched types
 --> test.rs:2:21
  |
2 |     let foo: bool = 42;
  |                     ^^ expected bool, found integral variable
  |
  = note: expected type `bool`
  = note:    found type `{integer}`
```

Unfortunately, the statusline given by neomake just shows `mismatched types`. Given how common type errors are in Rust development, this hinders development speed (you need to navigate to the line, open the quickfix window to find the types, and then close it again).

### Expected behavior

Neomake statusline shows 

```
cargo: mismatched types (expected `bool` found `{integer}`)
```

JelteF:
Would be fixed by #942

JelteF:
This should now be fixed because #1123 is merged

blueyed:
FWIW: ALE has a ALEDetail command, which displays details for an error (https://github.com/w0rp/ale/blob/c7bd5cc0ba799abb7e382751cdbea49c1b98a429/autoload/ale/cursor.vim#L112).
Maybe we could add something like that for this, too?

PR #1197, opened 2017-04-12
Use json output in the clippy maker

fuine:
This is solves #1196. I have a several questions/ideas regarding this PR and so this is at the moment regarded work in progress.

* As `clippy` heavily uses multiple information messages in the children array I decided to leave the first message appended to the parent message, and create subsequent `I` (information) messages for each child. This way we can retain certain important information visible, without the need to open the location list (see #793 for example).

* Currently i left out the information about the spans away from the created children (code for that is commented out on purpose), granting them effectively the same span as the parent. Ideally we could use the provided span if it exists, otherwise the parent's span should be used. However I am not sure if I didn't hit a bug/unwanted behavior in the `neomake#makers#ft#rust#CargoProcessOutput` function. Consider this example:

```
fn main() {
    let a = 1.0;
    let b = 2.0;
    assert_eq!(a, b);
}
```
`clippy` will yield this warning:

```
warning: strict comparison of f32 or f64
 --> src/main.rs:5:5
  |
5 |     assert_eq!(a, b);
  |     ^^^^^^^^^^^^^^^^^
  |
  = note: #[warn(float_cmp)] on by default
help: consider comparing them within some error
  | if ! (* left_val - * right_val).abs() < error {
note: std::f32::EPSILON and std::f64::EPSILON are available.
 --> src/main.rs:5:5
  |
5 |     assert_eq!(a, b);
  |     ^^^^^^^^^^^^^^^^^
  = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#float_cmp
  = note: this error originates in a macro outside of the current crate
```

Despite having a single warning, proposed maker gives two warnings (and 3 children, which are correct), namely:

```
<assert_eq macros>|5 col 6 warning| strict comparison of f32 or f64. #[warn(float_cmp)] on by default
src/main.rs|4 col 5 warning| strict comparison of f32 or f64. #[warn(float_cmp)] on by default
src/main.rs|4 col 5 info| consider comparing them within some error: if ! (* left_val - * right_val).abs() < error {
src/main.rs|4 col 5 info| std::f32::EPSILON and std::f64::EPSILON are available.
src/main.rs|4 col 5 info| for further information visit https://github.com/Manishearth/rust-clippy/wiki#float_cmp
```

It looks like either the processing function can't deal with the span that is given in the json, or the span itself is misleading. I wasn't able to reproduce this with the `cargo` maker and so I don't know which of these options is at play here. @JelteF do you have any ideas as to what might be the problem here?

@blueyed let me know what do you think about the additional `info` messages created from children. I think this is somewhat connected to [your comment in #793](https://github.com/neomake/neomake/issues/793#issuecomment-291766606)

### DIFF
--- a/autoload/neomake/makers/clippy.vim
+++ b/autoload/neomake/makers/clippy.vim
@@ -4,8 +4,6 @@
 let s:rustup_has_nightly = -1
 
 function! neomake#makers#clippy#clippy() abort
-    let errorfmt = neomake#makers#ft#rust#rustc()['errorformat']
-
     " When rustup and a nightly toolchain is installed, that is used.
     " Otherwise, the default cargo exectuable is used. If this is not part
     " of a nightly rust, this will fail.
@@ -22,17 +20,22 @@ function! neomake#makers#clippy#clippy() abort
         endif
     endif
 
+    let cargo_maker = neomake#makers#ft#rust#cargo()
+    let json_args = ['--message-format=json', '--quiet']
+
     if s:rustup_has_nightly
         return {
             \ 'exe': 'rustup',
-            \ 'args': ['run', 'nightly', 'cargo', 'clippy'],
-            \ 'errorformat': errorfmt,
+            \ 'args': ['run', 'nightly', 'cargo', 'clippy'] + json_args,
+            \ 'errorformat': cargo_maker.errorformat,
+            \ 'process_output': cargo_maker.process_output,
             \ }
     else
         return {
             \ 'exe': 'cargo',
-            \ 'args': ['clippy'],
-            \ 'errorformat': errorfmt,
+            \ 'args': ['clippy'] + json_args,
+            \ 'errorformat': cargo_maker.errorformat,
+            \ 'process_output': cargo_maker.process_output,
             \ }
     endif
 endfunction
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -82,6 +82,29 @@ function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
                         \ span.expansion.span)
             call add(errors, error)
         endif
+
+        if len(children) > 1
+            for child in children[1:]
+                if !has_key(child, 'message')
+                    continue
+                endif
+                let info = deepcopy(error)
+                let info.type = 'I'
+                let info.text = child.message
+                if has_key(child, 'rendered') && !(child.rendered is g:neomake#compat#json_null)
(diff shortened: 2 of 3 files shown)


JelteF:
@fuine can you post the json (or a link to it), otherwise it's hard to see what the issue is. It can definitely be the case that there is a bug in the process_output function, I've only tested it on errors I found. Also can you put the errors in a code block, makes them easier to read?

fuine:
@JelteF this is a prettified version of the message: http://sprunge.us/OZEZ

JelteF:
After reading better I'm not entirely sure what the problem is. Do you mean that this line shouldn't be there:
```
<assert_eq macros>|5 col 6 warning| strict comparison of f32 or f64. #[warn(float_cmp)] on by default
```

Because I can agree with that. Looking at the json it's because the actual file names for std macros are not available. The best thing is probably to add a regex check for the weird filename format (`<.*>`). If it matches just ignore the error.

fuine:
I have added the span support for children and check against suspicious filenames (macros), as suggested by @JelteF. I've also fixed tests, but I can see that while they pass on my local machine, they fail on the docker. The only difference is the `bufnr` field for the child info message. I have no experience with vimscript testing environment and so I don't really know how to solve this - should I just change it accordingly to the value that it gets on the docker? @blueyed could you help me with this one?

blueyed:
@fuine 
Rebased, and added two fixups.. the main issue was not using `bufnr('%')`, but a fixed number (7) here - which works when only running the Vader file by itself, but not when running all tests (e.g. `make testnvimx`).

fuine:
Thanks a lot! Can you review the code. as I think that there won't be any other significant changes?

blueyed:
Added some more more fixups: one optimization, and code cleanup otherwise.

I am not sure if having additional (info) entries is really good here, since it's still about a single issue, isn't it? - and then using `:cnext` etc makes it harder to go to the next issue etc..

I like the idea about something like `:ALEDetail` (https://github.com/neomake/neomake/issues/793#issuecomment-291766606) in general.
In the long run it's about having additional information for entries, which then could also be displayed in tooltips etc later.

Having said that, I think this is good for now anyway.

blueyed:
Auto-squashed the fixups.

blueyed:
@fuine 
Let me know if it's good from your side and then we can merge it.

fuine:
@blueyed I think it's fine. Once again thanks for the help.

blueyed:
Thanks!
You're welcome! :)

PR #1198, opened 2017-04-13
Fix readme markdown

bronzehedwick:
Github changed it's markdown parser, breaking headings that do not separate the hash from the text.

**Works**

`# Heading`

**Doesn't Work**

`#Heading`


blueyed:
Thanks!

PR #1202, opened 2017-04-14
Use empty()/!empty() in place of !len()/len()

jamessan:
VimL knows how to quickly return from empty(), so use it when
appropriate instead of unnecessarily calculating the length.


ViViDboarder:
LGTM

blueyed:
Thanks!

PR #1213, opened 2017-04-18
Go: serialize / prefer gometalinter

blueyed:
Some suggestions based on comments in https://github.com/neomake/neomake/pull/795.

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -67,6 +67,6 @@ function! neomake#makers#ft#go#gometalinter() abort
         \ 'append_file': 0,
         \ 'cwd': '%:h',
         \ 'mapexpr': 'neomake_bufdir . "/" . v:val',
-        \ 'errorformat': '%W%f:%l:%c:%m',
+        \ 'errorformat': '%f:%l:%c:%t%*[^:]: %m',
         \ }
 endfunction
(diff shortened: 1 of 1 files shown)


bradleyfalzon:
This looks good to me.

blueyed:
> I still don't believe that linters should ever be errors, these should be reserved for the compiler

If gometalinter calls "go vet" and this "Reports potential errors that otherwise compile.", those should be marked as errors, no?  (have not checked if gometalinter would actually report them as errors though)

> as long as they aren't run if there's a Go compiler error.

That should be provided through `serialize` / `serialize_abort_on_error` now.
Please note that this should be improved, but it should work if `go test -c` (the `go` maker) exits non-zero (#1214).

bradleyfalzon:
> If gometalinter calls "go vet" and this "Reports potential errors that otherwise compile.", those should be marked as errors, no? (have not checked if gometalinter would actually report them as errors though)

I don't believe gometalinter could reliably determine which issues are actual errors, but ultimately, if the code compiles, there are no errors - it's just warnings. Most of gometalinters checks are linters that help you out, "did you forget to check this error" and therefore aren't errors, just warnings as the code still compiles.

blueyed:
Let's merge this then as-is for now (will squash the fixup)?!
/cc @novas0x2a @JelteF @mhinz @tweekmonster

PR #1215, opened 2017-04-19
Improved sign handling

blueyed:
One of the main ideas here is to clean old signs only after new jobs have finished, where existing signs were re-used/recycled.  Only at the end old/left-over signs are removed.
This keeps existing signs for longer, which is useful while waiting for the new make run to finish, and avoids flickering in general.


blueyed:
There are issues, but also on master (AFAIK).  So I am pushing this forward.

PR #1218, opened 2017-04-19
rust: remove neomake#utils#get_or_create_buffer

blueyed:
Setting `filename` should be enough (Vim will add the buffer as unlisted).
This might need some fixes in Neomake itself to also look at `filename`,
and not only `bufnr`.

/cc @JelteF 
I've not found the previous discussion about it, but IIRC there was some.
I've stumbled upon this when noticing that Vim automatically adds buffers from `.filename` as unlisted buffers, so it might be something about the `cwd` not being correct etc only?!

### DIFF
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -88,7 +88,6 @@ endfunction
 
 function! neomake#makers#ft#rust#FillErrorFromSpan(error, span) abort
     let a:error.filename = a:span.file_name
-    let a:error.bufnr = neomake#utils#get_or_create_buffer(a:error.filename)
     let a:error.col = a:span.column_start
     let a:error.lnum = a:span.line_start
     let a:error.length = a:span.byte_end - a:span.byte_start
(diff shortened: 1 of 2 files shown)


blueyed:
This was added in https://github.com/neomake/neomake/pull/1123, but no tests are failing without it (?!).

JelteF:
I'll try this PR out somewhere this week. IIRC there were issues when not supplying a `bufnr` and the file was not opened yet. I think in all tests the files are actually opened before runnning the maker so it makes sense that they do not fail. It should be quite simple to make a new test case where the file is not opened before, unless something has changed that will cause an error.

blueyed:
@JelteF 
Any feedback already?

blueyed:
Related: https://github.com/neomake/neomake/issues/1257

JelteF:
I just tried this PR out and it indeed didn't work. I created a new pull request with a test that shows the issue: https://github.com/neomake/neomake/pull/1285

I hope you can track down what's causing it or that #1257 already fixes it. Feel free to close my PR and move the test case here, it was just the easiest way for me to share it.

PR #1234, opened 2017-04-24
Removing g:let as I'm pretty sure it's a typo

jdivock:
Pretty sure that's not a thing, at least it's not for me.


blueyed:
Thanks, merging already since I think the same.. :)
Nice PR number btw.

PR #1235, opened 2017-04-25
speed up rjsx-mode linter

bounceme:
this makes emacs only load the required packages, rather than load everything else the user has installed by package.el, which is far slower in some cases


blueyed:
Thanks!

PR #1239, opened 2017-04-27
Rubocop: add --force-exclusion

jakswa:
Problem: Neomake was running rubocop on a file that had been ignored in `.rubocop.yml`
Solution: Used the recommended flag to indicate that `rubocop <file>` runs should still adhere to these exclusions. To quote a contributor [over in rubocop's github repo](https://github.com/bbatsov/rubocop/issues/2492#issuecomment-164375887):
> There's an option called --force-exclusion, which editor plugins should use. The problem is that when you give a single file name to rubocop, then Exclude parameters don't come into play. Unless you run with --force-exclusion.

### DIFF
--- a/autoload/neomake/makers/ft/ruby.vim
+++ b/autoload/neomake/makers/ft/ruby.vim
@@ -6,7 +6,7 @@ endfunction
 
 function! neomake#makers#ft#ruby#rubocop() abort
     return {
-        \ 'args': ['--format', 'emacs'],
+        \ 'args': ['--format', 'emacs', '--force-exclusion'],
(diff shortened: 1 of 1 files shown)


jakswa:
Unsure why this build failed. Looks like something to do with `vint-diff`, which I didn't touch.

blueyed:
Yes, it enforces vint checks on the diff (in the scope of touched files).
You can rebase it after https://github.com/neomake/neomake/pull/1240 is merged, or just wait for me doing it.

blueyed:
Thanks for your contribution!

PR #1251, opened 2017-04-28
vue: extend upstream (javascript) makers

blueyed:
@jacoborus 
Please verify that this still works.

This way you get the benefits of https://github.com/neomake/neomake/pull/1250 etc.


jacoborus:
Hi, I just realized about this issue, I'll check it tonight

blueyed:
Merging as-is, please create an issue in case of problems.

Issue #1212, opened 2017-04-18
temporary files: alternative location(s)


blueyed:
I've noticed that the current handling of temporary files is problematic if there is tool configuration involved that relies on the path/name of files.

The concrete example is `luacheck`, where the `.luacheckrc` can contain something like `files["spec"].std = "+busted"`, which adds some configuration for files in the `spec` directory.
If you know lint a temporary file (e.g. after `:Gdiff`) this will be created somewhere below `tempname()` (i.e. `/tmp`), and the configuration does not get applied by `luacheck`.  (The `.luacheckrc` file itself gets used from the original root/location)

I see two possible solutions for this problem:
1. create the temporary file(s) in the same directory as the original file (`foo.lua` => `foo.lua.neomaketmp-$PID-$BUFNR.lua`).  This has the main drawback of adding files to the working tree, where `git add` etc might pick it up accidentally.  Making the files hidden there (i.e. prepend a dot on Linux) might help, but then they might not get matched by some pattern again (this also applies to the version without a dot already).
2. re-create the directory structure when creating temporary files in `/tmp` (`foo/bar.lua` => `/tmp/…/foo/bar.lua`).  For this it might still be necessary to copy over the tool's config file, too (`.luacheckrc` in this case).

blueyed:
@tweekmonster 
Any feedback on this?

I am currently going with temporary files in the original directory:
```vim
let temp_file = fnamemodify(orig_file, ':h')
            \ .slash.'.'.fnamemodify(orig_file, ':t')
            \ .'.neomaketmp_'.s:pid.'_'.make_id
            \ .'.'.fnamemodify(orig_file, ':e')
```

The hidden status (dot prefix) is really necessary with e.g. `rg`, since otherwise you might get results with a temporary file when searching with e.g. vim-grepper while linting is happening.

blueyed:
Using the new pattern now in https://github.com/neomake/neomake/issues/1252.
Will add some more tests there, and then merge it from there.

PR #1252, opened 2017-04-28
tempfiles: use a list and keep track of created dirs

blueyed:
Ref: https://github.com/neomake/neomake/issues/1212

TODO:
 - [x] tests for multiple temporary files
 - [ ] refactor generation of files, so that it can be used by e.g. #1287

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -365,7 +365,6 @@ function! s:command_maker_base._get_fname_for_buffer(jobinfo) abort
         else
             throw 'Neomake: no file name.'
         endif
-
     elseif getbufvar(bufnr, '&modified')
         let temp_file = self._get_tempfilename(a:jobinfo)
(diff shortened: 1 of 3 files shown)


blueyed:
Some TODO:
```vim
diff --git i/tests/tempfiles.vader w/tests/tempfiles.vader
index 344057f9..860f699c 100644
--- i/tests/tempfiles.vader
+++ w/tests/tempfiles.vader
@@ -131,6 +131,9 @@ Execute (Maker can specify temporary file to use via fn):
     AssertNeomakeMessage "Starting: true ".fnameescape(g:neomake_test_tempfile)
   endif
   AssertNeomakeMessage 'Removing temporary file: '.g:neomake_test_tempfile
+
+  " TODO: assert removal of dir(s)!
+
   bwipe
 
 Execute (maker._get_argv uses jobinfo for bufnr):
```

PR #1263, opened 2017-04-30
Fix haskell: hdevtools: empty output from stack

blueyed:
Based on https://github.com/neomake/neomake/pull/1261.

### DIFF
--- a/autoload/neomake/makers/ft/haskell.vim
+++ b/autoload/neomake/makers/ft/haskell.vim
@@ -73,9 +73,12 @@ function! neomake#makers#ft#haskell#hdevtools() abort
     if !exists('s:uses_cabal')
         let s:uses_cabal = 0
         if executable('stack')
-            let rootdir = systemlist('stack --verbosity silent path --project-root')[0]
-            if !empty(glob(rootdir . '/*.cabal'))
-                let s:uses_cabal = 1
+            let output = neomake#compat#systemlist(['stack', '--verbosity silent', 'path', '--project-root'])
+            if !empty(output)
+                let rootdir = output[0]
+                if !empty(glob(rootdir . '/*.cabal'))
+                    let s:uses_cabal = 1
+                endif
             endif
         endif
     endif


autoload/neomake/makers/ft/haskell.vim
```diff
@@ -73,9 +73,12 @@ function! neomake#makers#ft#haskell#hdevtools() abort
...
         if executable('stack')
-            let rootdir = systemlist('stack --verbosity silent path --project-root')[0]
-            if empty(glob(rootdir . '/*.cabal'))
-                let s:uses_cabal = 1
+            let output = neomake#utils#systemlist(['stack', '--verbosity silent', 'path', '--project-root'])
+            if !empty(output)
+                let rootdir = output[0]
+                if empty(glob(rootdir . '/*.cabal'))
+                    let s:uses_cabal = 1
+                endif
```
blueyed: @bonds 
Please confirm that this is OK.
It looks weird to me to set `s:uses_cabal = 1` when `rootdir . '/*.cabal'` is not found.

bonds: Yah, I think the meaning got inverted here:

https://github.com/neomake/neomake/commit/7542a1c92fa71f99ea102499e00826c406762793#diff-6d67a70692ff8830b142f7e3631900f4

So, should be !empty instead of empty. That said...I don't think hdevtools needs to told to use Stack anymore (its smart enough to use stack on its own), nor does it need the workaround for setting GHC-PACKAGE-PATH anymore, so we could perhaps remove that whole section of code.

bonds: Yah, just checked with a couple Cabal and non-Cabal projects that hdevtools if the maker is just reduced to:

````
function! neomake#makers#ft#haskell#hdevtools() abort
    let params = {
        \ 'exe': 'hdevtools',
        \ 'args': ['check', '-g-Wall'],
        \ 'mapexpr': s:CleanUpSpaceAndBackticks(),
        \ 'errorformat':
            \ '%-Z %#,'.
            \ '%W%f:%l:%v: Warning: %m,'.
            \ '%W%f:%l:%v: Warning:,'.
            \ '%E%f:%l:%v: %m,'.
            \ '%E%>%f:%l:%v:,'.
            \ '%+C  %#%m,'.
            \ '%W%>%f:%l:%v:,'.
            \ '%+C  %#%tarning: %m,'
        \ }
    return params
endfunction
````

bonds: Hmm. Maybe spoke too soon: https://github.com/hdevtools/hdevtools/issues/50

bonds: Well, I'm having trouble getting hdevtools working through any versions of this maker. Its working fine from the command line, but not through the maker. I'll keep banging on this for a bit and report back when I have something.

bonds: Ok, throwing up my hands for now...I'm struggling to keep hdevtools working in my environment, and I'm not convinced its worth it when ghc-mod works fine. I say, just invert the empty to !empty for now and I'll bang my head against this more when I can.

blueyed: Rebased, with the fix for `empty`.

blueyed:
Based on https://github.com/SpaceVim/SpaceVim/issues/475#issuecomment-301389328 you might end up here with no output at all, since you have not setup stack.

PR #1279, opened 2017-05-02
java: use neomake#utils#shellescape

blueyed:


blueyed:
I have not inlined it really, since there were a few places.  Would you prefer that?
The function/pattern is the same otherwise.

wsdjeg:
:] This PR is better.

PR #1280, opened 2017-05-02
java: use builtin tempname()

blueyed:
I've noticed that invoking neomake#makers#ft#java#javac during tests
leaves temporary directories behind.

Using `tempname()` simplifies the code a lot, and will make (Neo)Vim
remove those automatically.

### DIFF
--- a/autoload/neomake/makers/ft/java.vim
+++ b/autoload/neomake/makers/ft/java.vim
@@ -86,35 +86,8 @@ let s:has_maven = executable(expand(g:neomake_java_maven_executable, 1))
 let s:has_gradle = executable(expand(g:neomake_java_gradle_executable, 1))
 
 function! s:tmpdir() abort
-    let tempdir = ''
-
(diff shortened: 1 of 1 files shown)


blueyed:
@wsdjeg 
I wonder if the idea here was too have it more persistent - or if it would benefit from having it e.g. per buffer, assuming that things get cached better by `javac` then?!

We could add a helper then that would provide this, or we could just store it in the `b:` config right away?!

wsdjeg:
I remove tempname for some reason. but I do not remember why I remove it. will try to find it out.

blueyed:
FWIW, it's still used directly in `s:GetGradleClasspath`.

wsdjeg:
Most time I use maven instead. I need to find out the old commit message in my repo.

wsdjeg:
I have used this PR for more than  10 days, and it works well for me

blueyed:
Thanks for testing!
Back to master for you then.. :)

PR #1285, opened 2017-05-03
Clone of #1218 with added test case

JelteF:
This highlights the issue that I was having and shows why I created
get_or_create_buffer. If another file is open an no bufnr is given the current
bufnr is used.

### DIFF
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -131,7 +131,6 @@ endfunction
 
 function! neomake#makers#ft#rust#FillErrorFromSpan(error, span) abort
     let a:error.filename = a:span.file_name
-    let a:error.bufnr = neomake#utils#get_or_create_buffer(a:error.filename)
     let a:error.col = a:span.column_start
     let a:error.lnum = a:span.line_start
     let a:error.length = a:span.byte_end - a:span.byte_start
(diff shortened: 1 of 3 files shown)


blueyed:
Hmm, the 3 commits from you all seem to fail in the same way?!

I am looking into https://github.com/neomake/neomake/pull/1257 now again.. my guess is that rebasing this on top of it would fix it.

JelteF:
The revert is of the wrong commit should be of yours, but is of my second one

JelteF:
Should be fixed now

JelteF:
Hmm, still fails. Locally it works for me with `nvim +Vader tests/ft_rust.vader`.

blueyed:
How about just `make tests/ft_rust.vader`?
`nvim +Vader tests/ft_rust.vader` works,, but there are teardown checks that it does not catch: `(6/7) [EXECUTE] (X) Vim(let):E94: No matching buffer for src/add_assign_like.rs`.

blueyed:
(a new check added)

blueyed:
Might be running into something odd like https://github.com/vim/vim/issues/1676 after all.
I can reproduce this to be OK with `nvim +Vader tests/ft_rust.vader` however.

JelteF:
With the last commit it should be OK. The one before should be broken.

blueyed:
Should be fixed now.. #1257 helped, and then it was also affected by https://github.com/vim/vim/issues/1676.

I will clean it up and merge it when it's really green.

blueyed:
Pushed, thanks!

PR #1329, opened 2017-05-18
Don't conditionally disable fast-parser for python mypy

miedzinski:
Since mypy 0.510 there is no point to conditionally enable fast-parser - it is the only one available. Windows is supported by mypy.

```
λ mypy --no-fast-parser
Warning: --no-fast-parser no longer has any effect.  The fast parser is now mypy's default and only parser.
```


blueyed:
Thanks!

Issue #1109, opened 2017-03-14
Integrating custom maker for language plugin


mhartington:
### Expected behavior
Hi, I am the author of [nvim-typescript](https://github.com/mhartington/nvim-typescript) and I'm working on replacing the built in typescript tsc maker, with one that uses the typescript client.

![nvim-geterr](https://cloud.githubusercontent.com/assets/2835826/23880818/8cd03a00-082c-11e7-8a7c-8770c79891f7.gif)

Currently, it's called through it's own command `TSGetErr` but would love to add a custom makers that integrates with neomake.

What would be the recommended way to handle this?
The command is written in python and calls a ts-client that returns some raw output.

https://gist.github.com/mhartington/94b2940c77a8f5b619b8439d20c4c76f

I will extract this out into a function through the neovim-function decorator, but is there a return value that neomake expect?

Similar to how tsuquyomi integrates with syntastic.
https://github.com/Quramy/tsuquyomi/blob/master/syntax_checkers/typescript/tsuquyomi.vim

blueyed:
I think you should wait for https://github.com/neomake/neomake/pull/1050.
Basically you would define a `get_list_entries` method for the maker, that returns a list of entries that are then processed by Neomake, without starting a job.
I might look into getting this in later tonight, otherwise during the next days.
Given that we also can handle JSON by now, this allowed me to write a redpen-server maker, which uses `curl` - this was the reason for me to add auto-linting etc.. (with tempfile handling (just merged)).

mhartington:
Alright, and the list of entries would just match something from a quickfix entry?

As for handling json, that would be even better, since the return value from the ts-client is just json

```
{
  'seq': 0,
  'type': 'event',
  'event': 'semanticDiag',
  'body': {
    'file': '/Users/mhartington/myApp/src/pages/home/home.ts',
    'diagnostics': [

    {
      'start': { 'line': 7, 'offset': 3},
      'end': {'line': 7,'offset': 12},
      'text': 'Module \'"/Users/mhartington/myApp/node_modules/@angular/core/index"\' has no exportedmember \'component\'.'
    },

    {
      'start': {'line': 15,'offset': 2},
      'end': {'line': 15, 'offset': 11},
      'text': "Cannot find name 'Component'."
    }]
  }
}
```
I'm just processing it and creating a quickfix list item currently, but if I can avoid that, that would be great!

ahstro:
@mhartington FYI, #1050 is merged :)

[...]

PR #1361, opened 2017-05-23
Add tsuquyomi maker for TypeScript

### LINKED ISSUE #1356: Maker for Tsuquyomi, if possible
I am not sure whether it would be possible to have a maker for [Tsuquyomi](https://github.com/Quramy/tsuquyomi) since it uses a Nvim command to get errors, not a standalone program. I tried to find info on whether this was possible, but couldn't.

ahstro:
Very basic solution to #1356, but seems to work for me.
Two things I'd like to mention before you consider merging this:
  1. Running `tsuquyomi#createFixlist()` here seems to be synchronous. It freezes Neovim. Any ideas on how to prevent that?
  2. I have no idea how to debug VimScript, so couldn't check how the actual list looks. It seemed to work fine for me, but there might be stuff missing.

Close #1356

### DIFF
--- a/autoload/neomake/makers/ft/typescript.vim
+++ b/autoload/neomake/makers/ft/typescript.vim
@@ -25,3 +25,12 @@ function! neomake#makers#ft#typescript#tslint() abort
         \ 'errorformat': '%E%f[%l\, %c]: %m'
         \ }
 endfunction
+
+function! neomake#makers#ft#typescript#tsuquyomi() abort
+    let maker = {
+        \ }
+    function! maker.get_list_entries(jobinfo) abort
+        return tsuquyomi#createFixlist()
+    endfunction
+    return maker
+endfunction


blueyed:
1. > Running `tsuquyomi#createFixlist()` here seems to be synchronous.

Yes, it is not async.

Call stack:
1. https://github.com/Quramy/tsuquyomi/blob/cdae8ecde529019d50bdf2f25b782e9ee408a7c0/autoload/tsuquyomi.vim#L569
2. https://github.com/Quramy/tsuquyomi/blob/cdae8ecde529019d50bdf2f25b782e9ee408a7c0/autoload/tsuquyomi/tsClient.vim#L411
3. [tsuquyomi#tsClient#sendCommandSyncEvents](https://github.com/Quramy/tsuquyomi/blob/cdae8ecde529019d50bdf2f25b782e9ee408a7c0/autoload/tsuquyomi/tsClient.vim#L232)

btw: it does not appear to use Neovim's jobs either (https://github.com/Quramy/tsuquyomi/blob/cdae8ecde529019d50bdf2f25b782e9ee408a7c0/autoload/tsuquyomi/tsClient.vim#L24).

2. > I have no idea how to debug VimScript, so couldn't check how the actual list looks. 

`echo getloclist(0)` should give you all the details, and then looking at Neomake's logfile (after `let g:neomake_logfile = '/tmp/neomake.log'` gives you information about the processing.

/cc @Quramy
1. Is there something that could be done to help us here?
Something like the [tsuquyomi#tsClient#sendCommandOneWay](https://github.com/Quramy/tsuquyomi/blob/cdae8ecde529019d50bdf2f25b782e9ee408a7c0/autoload/tsuquyomi/tsClient.vim#L253), that would allow for a callback?
Preferably there would be an async version of `tsuquyomi#createFixlist()` already.

2. What is your take on supporting Neovim's jobs (instead of using vimproc there)?

blueyed:
FWIW: we could do something also like ALE is planning at https://github.com/w0rp/ale/issues/517 to have generic support for the "Language Server Protocol".

ahstro:
Re: not using Neovim's jobs, there's a discussion going on here https://github.com/Quramy/tsuquyomi/issues/57

Thank you for the debugging tips, I'll try it out when I have time :)

blueyed:
Let's maybe keep this simple (and blocking) for now?

ahstro:
Works fine for me. Been using it for a week now and it only seems to be blocking when first opening a file from a project, anything opened from that buffer seems to open asynchronously.

blueyed:
@ahstro 
See https://github.com/w0rp/ale/issues/20#issuecomment-305173783 (about https://github.com/jason0x43/vim-tss).

We might want to provide both of them, and e.g. default to vim-tss for Neovim then?
/cc @jason0x43

jason0x43:
vim-tss integrates itself with neomake by overriding the default tsc maker, so if someone has both neomake and vim-tss installed, neomake will use it. Getting error marks to display faster with neomake was the reason I originally wrote it. :)

mhartington:
Another option would be to provide a public callback that can process async results.
I brought it up https://github.com/neomake/neomake/issues/1109#issuecomment-303911114 as using remote plugins are async by default. 

If a getErrors function is async, but can call a public function as a callback, then it will not block on the error checking.

PR #1373, opened 2017-05-26
Log when cleaning make info

blueyed:


blueyed:
btw: Neovim 0.1.7 core dumped in its first build..

PR #1374, opened 2017-05-26
Maker for matlab linter mlint

dmalt:
By default mlint doesn't output filename when linting a single file.
Workaround is to paste current buffer name to the begining of each line of the linter output. 
This is achieved via
```
'mapexpr': "neomake_bufname.':'.v:val"
```
in maker configuration as suggested by @blueyed.

### DIFF
--- a/autoload/neomake/makers/mlint.vim
+++ b/autoload/neomake/makers/mlint.vim
@@ -1,0 +1,12 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#maker#mlint#mlint() abort 
+    return {
+        \ 'exe': 'mlint',
+        \ 'args': ['-id'],
+        \ 'mapexpr': "neomake_bufname.':'.v:val",
+        \ 'errorformat':
+        \ '%f:L %l (C %c): %m,' .
+        \ '%f:L %l (C %c-%*[0-9]): %m,',
+        \ }
+endfunction
--- a/autoload/neomake/makers/ft/mlint.vim
+++ b/autoload/neomake/makers/ft/mlint.vim
@@ -1,6 +1,6 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#ft#mlint#mlint() abort 
+function! neomake#makers#ft#mlint#mlint() abort
     return {
         \ 'exe': 'mlint',
         \ 'args': ['-id'],
--- a/autoload/neomake/makers/ft/mlint.vim
+++ b/autoload/neomake/makers/ft/mlint.vim
@@ -1,5 +1,9 @@
 " vim: ts=4 sw=4 et
 
+function! neomake#makers#ft#mlint#EnabledMakers() abort
+    return ['mlint']
+endfunction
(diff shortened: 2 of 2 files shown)


blueyed:
Thanks!
Is it guaranteed to have no multiline output?
Otherwise another approach would be needed.
The Golang makers also used this before, but actually produced output across multiple lines.

autoload/neomake/makers/mlint.vim
```diff
@@ -0,0 +1,12 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#maker#mlint#mlint() abort
```
blueyed: > Error: Vim(function):E746: Function name does not match script file name: neomake#maker#mlint#mlint (in /testplugin/autoload/neomake/makers/mlint.vim

You should move it to `./autoload/neomake/makers/ft/mlint.vim` and name it `neomake#makers#ft#mlint#mlint`.

dmalt:
Unfortunately there's no official documentation for mlint but I've checked on my repo and mlint never produced multiline output. But just in case I've found an alternative solution which is used by [MatlabFilesEdition](https://github.com/vim-scripts/MatlabFilesEdition) plugin. Apparently mlint starts to output filenames if more then one file is linted. Here's the solution from the plugin above:
```
" mlint doesn't provide filename information except if multiple
" filenames are given 
" With the following command :
" mlint <filename> <filename without extension>
" mlint produces an output like that :
" ========== <filename> ==========
" L x (C y): ID : Message
" L x (C y): ID : Message
" ..
" ..
" ========== <filename without extension> ==========
" L 0 (C 0): MDOTM :Filename 'filename' must end in .m or .M
"
" The filename can then be parsed
CompilerSet makeprg=mlint\ -id\ %\ %< 

CompilerSet errorformat=
      \%-P==========\ %f\ ==========,
      \%-G%>==========\ %s\ ==========,
      \%-G%>L\ %l\ (C\ %c):\ MDOTM%m,
      \L\ %l\ (C\ %c):\ %m,
      \L\ %l\ (C\ %c-%*[0-9]):\ %m,
      \%-Q
```

autoload/neomake/makers/mlint.vim
dmalt: Will do

blueyed:
Not sure if the hack is better, let's go with the `mapexpr` for now.
Otherwise we should also have something like defaults, or a postprocessor instead.

dmalt:
I agree, ```mapexpr``` solution is cleaner.

blueyed:
You should also add:

```vim
function! neomake#makers#ft#mlint#EnabledMakers() abort
    return ['mlint']
endfunction
```

Ok, you just did it.. :)

dmalt:
Finally :)

blueyed:
Thanks a lot.
Slightly amended in 6182a88.

blueyed:

dmalt:
Just did.
Cheers!

PR #1377, opened 2017-05-28
Fix stack command line options

sboehler:
This fix treats "--verbosity" and "silent" as independent tokens on the command line. 

The problem seems to have been introduced with https://github.com/neomake/neomake/commit/958b3c6f036a69a6cca71f198ae2565a6fcc9482 

@blueyed is probably the best person to review this.

Thanks!


blueyed:
Thanks!

PR #1379, opened 2017-05-29
cpp: add cpplint maker

fszymanski:


blueyed:
Thanks!
Please add it to the wiki's maker list.

PR #1394, opened 2017-06-05
Add support for Styled Components in JavaScript and JSX

albertorestifo:
[Styled Components](https://github.com/styled-components/styled-components) is getting traction.

The community built a pre-processor which allows linting of Styled Components trough StyleLint: https://github.com/styled-components/stylelint-processor-styled-components

This PR adds support of stylelint to JavaScrit and JSX files. I manually configured it in my vimrc and it works perfectly.

PR #1395, opened 2017-06-05
javascript/jsx: add stylelint (via Styled Components)

blueyed:
Followup of https://github.com/neomake/neomake/pull/1394.


blueyed:
@albertorestifo 
Please add it to the wiki's list of makers, with links to the documentation how to configure it, since it appears to be non-trivial / needs some extra packages.

PR #1399, opened 2017-06-05
WIP: async get_list_entries

blueyed:
Via https://github.com/neomake/neomake/issues/1109#issuecomment-306291569.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -199,6 +199,24 @@ function! s:jobinfo_base.get_pid() abort
     endtry
 endfunction
 
+function! s:jobinfo_base.await_async_entries(job) abort
+    if !has_key(a:job, 'cancel')
+        throw 'jobinfo.await_async_entries requires a cancel function.'
+    endif
+    call extend(self, a:job)
+    return self
+endfunction
+
+function! s:jobinfo_base.add_entries(entries, ...) abort
+    let partial = a:0 ? a:1 : 0
+    if !empty(a:entries)
+        call s:ProcessEntries(self, a:entries)
+    endif
+    if !partial
+        call s:CleanJobinfo(self)
+    endif
+endfunction
+
 function! s:MakeJob(make_id, options) abort
     let job_id = s:job_id
     let s:job_id += 1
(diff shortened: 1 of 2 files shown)


mhartington:
Seems like this isn't playing well with remote plugins (python3 in my case). Passing `self` or even `jobinfo` to the remote plugin side will convert any function refs to `None`. According to Justin M Keyes: 

>funcrefs generally won't work over RPC, though maybe we can hack around it since VimL funcrefs are basically just strings. Though, actually, they aren't in recent Nvim/Vim versions

Maybe a general public method would be a better alternative?

blueyed:
So `s:jobinfo_base.await_async_entries` should become a autoload function instead?
Do you have some WIP branch / snippet which shows how you want to use it?
We could have it based on a Python rplugin in the tests then even.

mhartington:
Sure. I’ll push something up in a few hours. 

Mike Hartington

mhartington:
I was thinking something like this.

https://github.com/mhartington/neomake/commit/d7506d8ed88d96f7c8b0ac8f19076cf6a6eb7ab9

blueyed:
1. I do not like that there is a new maker setting, when `get_list_entries` can return something to indicate it already.

2. Your entry point seems to be the following, but you've said that passing the `jobinfo` dict is problematic before?!

```vim
function! neomake#process_async(entries, jobinfo) abort
    if !empty(a:entries)
        call s:ProcessEntries(a:jobinfo, a:entries)
    endif
    call s:CleanJobinfo(a:jobinfo)
endfunction
```

3. With my approach it is possible to add entries on-the-fly, without cleaning the job already, which it gets cleaned/finished in your case always?!


Can you show me the actual (Python?) plugin code that you want to use this with?

mhartington:
>I do not like that there is a new maker setting, when get_list_entries can return something to indicate it already.


Sure thing.

>Your entry point seems to be the following, but you've said that passing the jobinfo dict is problematic before?!

It's only problematic for any function refs that are passed between the remote plugin and viml. 
Any function refs loose reference and are set to `None` in the python side. Looking through the code in `ProcessEntries`, nothing looks like it's referencing any functions on the current job, so it seemed ok in this case. 


Here's the code I used when testing.

https://github.com/mhartington/nvim-typescript/blob/feat/getErrFunc/rplugin/python3/nvim-typescript/__init__.py#L255-L277

blueyed:
@mhartington 
Any new insights in this regard?
I've started back then hacking on this for a bit, but it is stashed deep currently.

mhartington:
@blueyed Hey sorry about this, life kind of got in the way 😄 

I'll get this updated and reworked

mhartington:
@blueyed I have an improved version of this that is working pretty well, but it keeps throwing errors: 
```
Neomake: warning: missing make_info key: 1```

I have the commit in a branch here https://github.com/mhartington/neomake/commit/af367cc83053ff44c62fe54e62c9d48406677bea

Any idea why thats getting shown?

Issue #915, opened 2017-01-07
indicator that neomake is running


nhooyr:
If there is no linter defined, only then should `:Neomake` produce no output. Otherwise, it should give a little message that it is running so I know that something is happening.

blueyed:
I would find this too spammy, but maybe increasing/adjusting `g:neomake_verbose` helps you / could be tamed to fit this use case better.

There is certainly a difference though between running it manually or automatically, which makes it belong to https://github.com/neomake/neomake/projects/1 somehow.

I am using an indicator myself in the statusline: https://github.com/blueyed/dotfiles/blob/ce42f6c9b67642582d490014765e47d94c1cf7a7/vim/plugin/statusline.vim#L402-L413

blueyed:
Related: #337 (also from you).

ahstro:
Would love this, either as part of the plugin, a separate plugin, or in my own statusline. Haven't taken the time to figure out VimScript yet though, so I can't grok enough of your solution to insert it into my own flow, @blueyed. `g:neomake_verbose` is not doing the trick for me, unfortunately.

davidosomething:
i use this:

https://github.com/davidosomething/dotfiles/blob/d6adee60eed5074efedf409ca07b3b332daaa2ae/vim/autoload/dkostatus.vim#L117-L124

in my status as:

https://github.com/davidosomething/dotfiles/blob/d6adee60eed5074efedf409ca07b3b332daaa2ae/vim/autoload/dkostatus.vim#L69

the output is like this:

https://cl.ly/3g301i0M3B3N/Image%202017-05-18%20at%209.23.27%20AM.public.png

blueyed:
Included in #1416.

Not merged yet though - you might want to provide feedback there.

blueyed:
#1416 has been merged.

PR #1416, opened 2017-06-12
[RFC] statusline improvements

blueyed:
Fixes https://github.com/neomake/neomake/issues/915.

### DIFF
--- a/autoload/neomake/statusline.vim
+++ b/autoload/neomake/statusline.vim
@@ -1,3 +1,5 @@
+scriptencoding utf-8
+
 let s:qflist_counts = {}
 let s:loclist_counts = {}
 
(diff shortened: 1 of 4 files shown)


autoload/neomake/statusline.vim
```diff
@@ -89,3 +91,143 @@ endfunction
...
+    if has_key(s:cache, a:bufnr)
+        unlet s:cache[a:bufnr]
+    endif
+endfunction
+
+augroup neomake_statusline
+    autocmd!
+    autocmd User NeomakeJobStarted,NeomakeJobFinished call s:clear_cache(g:neomake_hook_context.jobinfo.bufnr)
+    " Trigger redraw of all statuslines.
+    autocmd User NeomakeJobFinished let &stl = &stl
```
wsdjeg: why not use : 
```
autocmd User NeomakeJobFinished redrawstatus!
```

blueyed:
Need to finish docs, but would appreaciate any feedback already.

/cc @chrisbra @rdnetto regarding integration into vim-airline

chrisbra:
I am missing what kind of airline problem you are trying to solve here.

blueyed:
@chrisbra 
See your comment at https://github.com/vim-airline/vim-airline/pull/1275#issuecomment-273039607 - that's basically it.

It needs adoption in Airline of course, and that's why I ask for feedback.

chrisbra:
I am still missing something. Is this about getting the counts from the quickfix list (instead of the location list, which seems to be currently the only supported way by the airline neomake extension?) If so, can you please point me to a specific commit, that only does that? I see to much neomake specific stuff in your commits and feel it hard to find the relevant code. I could then view a diff with the current neomake extension, which allows me to better review. Or is this about something else?

blueyed:
@chrisbra 
Airline could just use the new `neomake#statusline#get()`.
See the doc added in the diff (WIP).

blueyed:
:fireworks:

PR #1418, opened 2017-06-13
filename for elm maker output item

nyrkovalex:
Neomake complains about missing `bufnr` property in elm maker output. This request adds `bufnr` to error dictionary

### DIFF
--- a/autoload/neomake/makers/ft/elm.vim
+++ b/autoload/neomake/makers/ft/elm.vim
@@ -30,6 +30,7 @@ function! neomake#makers#ft#elm#ElmMakeProcessOutput(context) abort
 
                 let l:compiler_error = item['tag']
                 let l:message = item['overview']
+                let l:filename = item['file']
                 let l:region_start = item['region']['start']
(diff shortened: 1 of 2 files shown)


autoload/neomake/makers/ft/elm.vim
```diff
@@ -42,6 +42,7 @@ function! neomake#makers#ft#elm#ElmMakeProcessOutput(context) abort
                             \ 'lnum': l:row,
                             \ 'col': l:col,
                             \ 'length': l:length,
+                            \ 'bufnr': bufnr('%'),
```
blueyed: I think it would be better to set `filename` instead.

nyrkovalex: Thank you, `filename` it is. Tests are also green.

nyrkovalex:
`filename` is used instead of `bufnr`

blueyed:
Thanks!

nyrkovalex:
Not at all :) Thank you and all neomake contributors for such an awesome project!

PR #1419, opened 2017-06-14
rust: set cwd to the directory of current file

lilydjwg:
`cargo check` can run inside the project, but not outside.


tests/ft_rust.vader
```diff
@@ -1,5 +1,11 @@
 Include: include/setup.vader
 
+Before (create src directory):
+  call mkdir('src')
+
+After (cleanup src directory):
+  !rmdir src
+
```
blueyed: Before/After cannot be used in Neomake's tests like that, we have them globally.
But the `build` dir can be assumed to be there for example.

blueyed: You can run the tests using `make tests/ft_rust.vader` for example.

lilydjwg: Oh....I'll try it. But changing to build/ will require rewrite those fixtures.

lilydjwg: Thanks, `make tests/ft_rust.vader` is much better :-)

But I still get those two kinds of errors.

blueyed: As for the `bufnr`, Vim might create a new buffer now?
You can add `Log neomake#utils#redir('ls!')` to see the buffer list.

"More than 1 window after tests" is because of the previous failure.

blueyed: I've not looked at the tests closely, so maybe it actually fixes something there even..

lilydjwg:
The check finally passed after I adjusted the fixtures correctly :-)

blueyed:
@JelteF 
Please review.

blueyed:
Thanks!

Issue #1287, opened 2017-05-04
Feature: allow linting of range


blueyed:
It should be possible to select lines visually in e.g. a `.travis.yml` file and call `Neomake sh` on them.

PR #1424, opened 2017-06-17
Feature/vimwiki filetype

justfielding:
My attempt at adding makers for [vimwiki](https://github.com/vimwiki/vimwiki/blob/dev/doc/vimwiki.txt) filetypes. I use markdown for my vimwiki syntax, but as I was about to submit a PR with just the first commit I realized markdown isn't even the default syntax.

This implementation defaults to turning on 'proselint' and 'writegood' and then depending on the extension other syntax makers can be added. I initially only added markdown, but obviously (and probably better than how I set it up) it can be extended to include any vimwiki syntax desired. 

First time doing anything in vim script, thankfully it seemed pretty basic. Was a little unsure on adding to a list using the + operator versus cal add() or whatever the alternative was, but overall seemed straight forward.

### DIFF
--- a/autoload/neomake/makers/ft/vimwiki.vim
+++ b/autoload/neomake/makers/ft/vimwiki.vim
@@ -1,0 +1,9 @@
+function! neomake#makers#ft#vimwiki#SupersetOf() abort
+    return 'markdown'
+endfunction
+
+function! neomake#makers#ft#vimwiki#EnabledMakers() abort
+    return neomake#makers#ft#markdown#EnabledMakers()
+endfunction
+
+" vim: ts=4 sw=4 et
--- a/autoload/neomake/makers/ft/vimwiki.vim
+++ b/autoload/neomake/makers/ft/vimwiki.vim
@@ -1,9 +1,21 @@
 function! neomake#makers#ft#vimwiki#SupersetOf() abort
-    return 'markdown'
+    return 'text'
 endfunction
 
 function! neomake#makers#ft#vimwiki#EnabledMakers() abort
-    return neomake#makers#ft#markdown#EnabledMakers()
+    let makers = ['proselint', 'writegood'] + neomake#makers#ft#text#EnabledMakers()
+    if expand('%:e') ==? 'md'
+      let makers = executable('mdl') ? ['mdl'] : ['markdownlint'] + makers
+    endif
+    return makers
+  endfunction
+
+function! neomake#makers#ft#vimwiki#mdl() abort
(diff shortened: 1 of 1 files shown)


blueyed:
So the filetype you are using is `vimwiki`, but with a `.md` extension?

What about using `md.vimwiki` for the filetype, since you already seem to configure it explicitly?

Then `vimwiki` could be just a superset of "text", and you (hopefully) get also the makers from ".md".

blueyed:
(The build failure is due to not using 4 spaces for indentation only.)

justfielding:
Hey, thanks for the feedback.  I meant to push the commit fixing my indent mishap right after seeing the build failure, but in typical ADD fashion I somehow stopped mid commit message lol. Hopefully this takes care of that.

As far as the filetype and extension situation, yeah vimwiki sets the filetype to vimwiki, but has quite a few syntax options available. The default being similar to google's and then popular ones like `markdown` and `mediawiki`. The extension can also be specified along with the syntax, so technically `.md `was another choice specific to my usage and might need to be addressed in the maker setup as well.

More than happy with adjusting my already questionable/unsure of implementation for whatever you recommend. I don't "technically" explicitly state the filetype, but the [vimwiki plugin](https://github.com/vimwiki/vimwiki/blob/dev/doc/vimwiki.txt) does indeed do that for me. I'm not sure what, if any, issues vimwiki would have with me overwriting the filetype, but I will check in to it and update once I find out.

blueyed:
Just a shower thought: Neomake could be made to lint the different code blocks (`{{{filetype
 … }}}` (https://github.com/vimwiki/vimwiki/blob/8cdc1c15388cc7f4edb827ff15dbc31d592a79af/doc/vimwiki.txt#L2074) by themselves (using different makers).  Related: #1287.

As for the filetype, try setting it to `vimwiki.md` manually (when using Markdown).
This should use Markdown linters, and hopefully Vimwiki does not break otherwise.

justfielding:
Planning to get some time to jump back to this over the weekend. Cheers

blueyed:
@justfielding 
Maybe this weekend then? ;)

blueyed:
@justfielding 
Ping?!

jcromero:
Creating a file `autoload/neomake/makers/ft/vimwiki.vim` with the following contents:
```vim
function! neomake#makers#ft#vimwiki#SupersetOf() abort
    return 'markdown'
endfunction
```
is perhaps the simplest solution

Comments?

PR #1446, opened 2017-07-09
do not highlight when line is 0

esquires:
Prior to this pull request, when let `g:neomake_highlight_lines=1` and the problem line was 0 (e.g., with cpplint and a no copyright warning), neomake would generate the following error:

```Neomake: [1.1:1:1] Error during output processing for cpplint: Vim(call):Line number outside range.```

This pull request excludes line highlighting when the line number is less than 0. An example output of cpplint for this type of error is:

```path/to/file:0:  No copyright message found.  You should have a line: "Copyright [year] <Copyright Owner>"  [legal/copyright] [5]```

### DIFF
--- a/autoload/neomake/highlights.vim
+++ b/autoload/neomake/highlights.vim
@@ -57,18 +57,20 @@ else
 endif
 
 function! neomake#highlights#AddHighlight(entry, type) abort
+    " some makers use line 0 for file warnings
+    " (e.g. cpplint with no copyright warnings)
+    " these cannot be highlighted since the line does not exist
+    if a:entry.lnum == 0
+        return
+    endif
+
     if !has_key(s:highlights[a:type], a:entry.bufnr)
         call s:InitBufHighlights(a:type, a:entry.bufnr)
     endif
     let l:hi = get(s:highlight_types, toupper(a:entry.type), 'NeomakeError')
     if get(g:, 'neomake_highlight_lines', 0)
         if s:nvim_api
-            " some makers use line 0 for file warnings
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: autoload/neomake/highlights.vim [hunk 1]
```diff
@@ -63,7 +63,12 @@ function! neomake#highlights#AddHighlight(entry, type) abort
     let l:hi = get(s:highlight_types, toupper(a:entry.type), 'NeomakeError')
     if get(g:, 'neomake_highlight_lines', 0)
         if s:nvim_api
-            call nvim_buf_add_highlight(a:entry.bufnr, s:highlights[a:type][a:entry.bufnr], l:hi, a:entry.lnum - 1, 0, -1)
+            " some makers use line 0 for file warnings
+            " (e.g. cpplint with no copyright warnings)
+            " these cannot be highlighted since the line does not exist
+            if a:entry.lnum > 0
+                call nvim_buf_add_highlight(a:entry.bufnr, s:highlights[a:type][a:entry.bufnr], l:hi, a:entry.lnum - 1, 0, -1)
+            endif
```
blueyed: I think this should be skipped in general (at the top of the function) instead.
Or even with the caller(s) of the function?!

side note: often (with other plugins etc) line numbers < 1 get normalized to 1, but I think it makes sense to have something like global items (with line number 0).

esquires:
It is now moved to the top of the function.

blueyed:
Thank you.
It should also have a test, but I can take care of that soonish™.

blueyed:
Thank you, merged it with tests and amended comment.

PR #1456, opened 2017-07-12
Allow use of .chktexrc for chktex

michaelmior:
This just adds `-l .chktexrc` to the chktex command line so that it's possible to specify some local configuration options.
The file is ignored if it doesn't exist.
This warning is not caught by the parser, so if the file doesn't exist, the behaviour should be just as it is now.

### DIFF
--- a/autoload/neomake/makers/ft/tex.vim
+++ b/autoload/neomake/makers/ft/tex.vim
@@ -5,8 +5,8 @@ function! neomake#makers#ft#tex#EnabledMakers() abort
 endfunction
 
 function! neomake#makers#ft#tex#chktex() abort
-    return {
-                \ 'args': ['-l', '.chktexrc'],
+    let maker = {
+                \ 'args': [],
                 \ 'errorformat':
                 \ '%EError %n in %f line %l: %m,' .
                 \ '%WWarning %n in %f line %l: %m,' .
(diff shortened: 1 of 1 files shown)


blueyed:
I think it would be better to add it to `args` dynamically, like for the pylama maker: https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/python.vim#L246-L249

This way, it uses the nearest config file, and not just the one from the current dir.

michaelmior:
Actually, I realized when the filename is `.chktexrc`, chktex automatically detects this in the working directory, so this PR as written is irrelevant. Your suggestion makes sense though.

blueyed:
According to `strace -f -e file chktex  2>&1|grep chktexrc` it also looks at one in $HOME:

```
access("/usr/share/texmf-dist/chktex/chktexrc", R_OK) = 0
open("/usr/share/texmf-dist/chktex/chktexrc", O_RDONLY) = 3
access("/home/user/.chktexrc", R_OK)  = -1 ENOENT (No such file or directory)
access(".chktexrc", R_OK)               = -1 ENOENT (No such file or directory)
```

blueyed:
Do you plan to update the PR?

michaelmior:
I'd still like to update the PR to put this together, but I probably won't have a chance to do it in the near future. Feel free to close if you like and I can reopen if I'm able to get back to it.

blueyed:
Fair enough, try your updated branch.. :)

blueyed:
Merged in c40bbb4.

michaelmior:
Looks great! Thanks :)

PR #1464, opened 2017-07-21
Add Solidity maker

beaugunderson:
A maker for the [`solium`](https://github.com/duaraghav8/Solium) [Solidity](https://github.com/ethereum/solidity) linter.

Here it is in action:

screenshot


blueyed:
Thanks!
Slightly amended (removed the `exe` entry).

Issue #614, opened 2016-09-17
Python makers: look at shebang to use python2/python3/python


blueyed:
It makes sense to use e.g. `python2` if it is used in the shebang instead of `python`.

See the `sh` maker.

Via https://github.com/neomake/neomake/issues/449#issuecomment-225687472.

blueyed:
Ok, this has two fans already.. ;)

varac:
One more !!!

bonderado:
That would be a welcome addition,
 waiting for built-in support I patch the 
python makers as needed like this:
====================%<------------------------------------
function! Selectpython()
  let l:exename = "python"
  if !empty(matchstr(getline(1), "^#!"))
    let l:exename = matchstr(getline(1), "python[23]*")
  endif
  let l:number = matchstr(l:exename, "[23]$")
  let g:neomake_python_python_maker =  neomake#makers#ft#python#python()
  let g:neomake_python_python_maker.exe = l:exename." -m py_compile"
  
  let g:neomake_python_pylint_maker =  neomake#makers#ft#python#pylint()
  let g:neomake_python_pylint_maker.exe = "pylint".l:number
endfunction

augroup CustomizePythonMaker
  au!
  au BufWrite *.py  call Selectpython()
augroup END
====================%<------------------------------------

I guess it could be more clever, but works for me.

blueyed:
@bonderado 
Thanks for your input which might help to fix this.

I am not really interested in this myself, since I only use Python 3, and with a Python 2 virtualenv, `python` would point at that.

But I still see that this is a nice feature after all.

I might

PR #1470, opened 2017-07-25
Pyswitch

bonderado:
This is my new attempt to address issue #614 and my first pull request, please be patient.
My previous solution (https://github.com/neomake/neomake/issues/614#issuecomment-314583170) would have bee just a note in the documentation, but it seemed too ad hoc, so i decided to try and make it more general.


autoload/neomake/makers/ft/python.vim
```diff
@@ -296,3 +296,45 @@ function! neomake#makers#ft#python#py3kwarn() abort
...
+
+    " Scan the shebang
+    let l:shebang = matchlist(getline(1), '^#!.*python\([0-9.]*\)')
+
+    if empty(l:shebang) || l:shebang[1] == ''
+        " Use the major number from s:default_version
+        " Ip $ the shebang does not contain one or is absent
+        let l:pyversion = s:default_version[0]
+    else
+        let l:pyversion = l:shebang[1]
```
blueyed: This could also call it with `--version`, instead of relying on the name.
Maybe only in case of `l:shebang[1] == ''`?

bonderado: Let see:
only on the first call, the function evaluates `python --version` and stores the version number in s:pyversion (in my system this would be ['2, '7', '13'] right now)

then scan the first line the regex can only match if the line begins wit '#!' and contains python as last word, possibly followed by an explicit version number. The possible outcomes are:
#!<anything> <anythingelse>  ->  []
#!<anything> python  ->  ['#!<anything> python', '', '', '', '', '', '', '', '', '']
#!<anything> python<version>  ->  ['#!<anything> python<version>', '<version>', '', '', '', '', '', '', '', '']

Here <anything> should be '/usr/bin/env', but I made it more generic just in case of strange installations, <version> can be any number with or without dots e.g.: 2, 2.7, 3, 3.6.

If empty(l:shebang), the regex did not match, the second part of the condition at line 315 is not evaluated and the default version is used.
if ! empty(l:shebang) the regex matched, the requested interpreter is 'python<version>' and l:shebang[1] == '' is checked: the latter is only true if no explicit version number is present and again the default version is used.
it is not possible for both parts of the condition to be true and if line 320 is reached, the name of the interpreter contains a version number and I would use just that since it should be the script author responsibility to set the correct shebang (note that in the default I only use the major number, while here more specific requests are allowed).

There is the possibility that the requested interpreter is not present in the system (or evenany python at all, but this feature is meant to be an opt-in, so it should not be an issue.

It is good that I checked, since I spotted a bug that would trigger only if I were to configure options for 'key' == '2'

blueyed:
Thanks for the update.
As you can see now it needs tests.. ;)

As for the design / idea itself: I think it should use the new config interface more, and could be added there in general then maybe?!

On the other hand it also seems to be rather complicated and you might be better off to just configure this per project, e.g. using https://github.com/embear/vim-localvimrc or matching in your vimrc.

The following test shows how it can be done:
```vim
Execute (FileType can be used to override maker exe):
  new
  let b:some_condition = 1
  noautocmd setfiletype python
  AssertEqual neomake#GetMaker('pylint').exe, 'pylint'

  function! s:my_adjust_pylint()
    if get(b:, 'some_condition', 0)
      call neomake#config#set('b:ft.python.pylint.exe', 'pylint3')
    endif
  endfunction
  augroup neomake_tests
    autocmd FileType python call s:my_adjust_pylint()
  augroup END
  doautocmd FileType
  AssertEqual neomake#GetMaker('pylint').exe, 'pylint3'

  new
  noautocmd setfiletype python
  doautocmd FileType
  AssertEqual neomake#GetMaker('pylint').exe, 'pylint'

  bwipe
  bwipe
```

The new config system allows you to easliy set (buffer) settings.
Instead of `neomake#config#set('b:ft.python.pylint.exe', 'pylint3')` you could also use `neomake#config#set('ft.python.pylint.exe', 'pylint3')` to set it globally.

This could maybe go into a recipes section, and the new config system needs to be documented in general.

bonderado:
Hi, thanks for the suggestions,

I had no idea on how to write test cases for vimscript and I was a bit
daunted.
Right now I am quite busy, so I am not sure I can find time to make the
patch better, but I intend to try my hand  at it.

I agree that my approach is a bit complex; in fact I am not sure if
languages other than python would benefit from such an infrastructure. I
will see if it can be simplified.

---------------------------------------------------
Diego Caraffini

Anar caluva tielyanna
  (The Sun shall shine on your path)
   from Unfinished Tales  by JRR Tolkien

blueyed:
@bonderado 
What do you think about skipping adding this complexity (i.e. this PR), but make it easier to configure this given the tools that are there already.

Can you describe your use case in detail, please?

blueyed:
@bonderado 
I see you keep pushing here, but have not answered my questions?!

blueyed:
btw: rebasing on master should get you better test feedback (moved to circleci).

bonderado:
HI, sorry for the delay.

The main motivation for this is that we are using several scripts to perform subsequent stages of data analysis and some of the collaborators cannot be convinced to stick to one version of python. On one hand I could just fix the scripts as they are created, on the other, since each script is run by itself, there is no actual urge to do it. so I came up with a system to select the right checkers automatically, rather than fight windmills.
  
 in the last few days i changed the way I apply the alternative settings, using the neomake native handling of multiple filetypes (eg: jsx.javascript).
This does not require new code, but does not work with a buffer that "needs" a double filetype.
I still left my python selection function (and some test cases) as an example, but it can be easily moved to the documentation if you think it is more suitable, so that anyone can do something similar.
At this point it is really not that hard to maintain it in the init.vim, so I am pushing this version to let you decide whether to keep just the documentation part or more (assuming the tests succeed, I get some failures due to localized strings, but I am confident).

blueyed:
JFI: just created https://github.com/neomake/neomake/pull/2062 as a simpler / more direct approach to make the `python` maker use the shebang.

PR #1485, opened 2017-07-30
Use latest supported version of C++ standard by default for cpp maker

mqudsi:
By default, gcc and clang default to the 2003 C++ version. Regardless of project make settings, I think it's a safe assumption for the default standard that neomake checks syntax against to be the latest supported standard (currently C++1z)

### DIFF
--- a/autoload/neomake/makers/ft/cpp.vim
+++ b/autoload/neomake/makers/ft/cpp.vim
@@ -9,12 +9,14 @@ endfunction
 function! neomake#makers#ft#cpp#clang() abort
     let maker = neomake#makers#ft#c#clang()
     let maker.exe = 'clang++'
+    let maker.args = makers.args + ['-std=c++1z']
     return maker
 endfunction
 
 function! neomake#makers#ft#cpp#gcc() abort
     let maker = neomake#makers#ft#c#gcc()
     let maker.exe = 'g++'
+    let maker.args = makers.args + ['-std=c++1z']
     return maker
 endfunction
 
--- a/autoload/neomake/makers/ft/cpp.vim
+++ b/autoload/neomake/makers/ft/cpp.vim
@@ -9,14 +9,14 @@ endfunction
 function! neomake#makers#ft#cpp#clang() abort
     let maker = neomake#makers#ft#c#clang()
     let maker.exe = 'clang++'
-    let maker.args = makers.args + ['-std=c++1z']
+    let maker.args += ['-std=c++1z']
     return maker
 endfunction
 
 function! neomake#makers#ft#cpp#gcc() abort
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/cpp.vim
```diff
@@ -9,12 +9,14 @@ endfunction
...
     let maker = neomake#makers#ft#c#clang()
     let maker.exe = 'clang++'
+    let maker.args = makers.args + ['-std=c++1z']
     return maker
 endfunction
 
 function! neomake#makers#ft#cpp#gcc() abort
     let maker = neomake#makers#ft#c#gcc()
     let maker.exe = 'g++'
+    let maker.args = makers.args + ['-std=c++1z']
```
blueyed: Typos: s/makers/maker/g

mqudsi:
Thanks @blueyed

autoload/neomake/makers/ft/cpp.vim
blueyed: You can also use `+=`.

blueyed: And then maybe actually test this yourself locally? :)

mqudsi: Sorry, I thought I did and "it worked" because no errors were posted (instead of errors about C++11 keywords).

blueyed: There should have been some error from Neomake then - if not, that's a bug.

blueyed: A test like the following throws:
```
Execute (Error gets reported from maker function):
  function! neomake#makers#ft#neomake_tests#true() abort
      let foo = foo + 'bar'
  endfunction

  set ft=neomake_tests
  RunNeomake true
```

This could be caught somehow and get reported in a better way, but should not happen in the first place, so it is fine.

mqudsi: I think my branch of the fork may have failed to load correctly into neovim or something because I didn't get any errors or warnings and that makes no sense. Sorry once more.

blueyed: No problem - I just want to ensure that Neomake is behaving sane.

Will you push a fixup?

mqudsi: Sure thing.

mqudsi:
OK, I made sure I was actually loading my branch of the plugin and it seems to be working correctly now. Thanks for your patience!

blueyed:
Thank you for your contribution!

mqudsi:
Thanks for being so welcoming!

PR #1488, opened 2017-08-01
C and C++ maker improvements

mqudsi:
I've made some improvements to the C and C++ makers that should be safe to apply for everyone for single-file makers. The most important improvement is include -I./ in the arguments to the single-file makers for C/C++ for clang and gcc to include the current project directory in the header file search path. A lot of build systems generate a top-level config.h that is placed in that directory, regardless of where the rest of the source code actually lives.

Also, the clang-check and clang-tidy makers were broken for a couple of reasons. First, the clang arguments do not go directly to the clang-{check,tidy} tool, they have to be separated by a literal `--`, otherwise they are interpreted as arguments to the tool and not the compiler. Second, for some reason clang-check and clang-tidy were refusing to recognize clang arguments when `-fsyntax-only` (from the default clang c/cpp maker args) was included in the argument list.

Additionally, the file path was being included for these tools via the clang arguments, but the file path needs to go before the `--` and not after. 

All-in-all, it was way saner and much less fragile to just specify the parameters for these tools separately rather than have them inherit from the clang maker.

### DIFF
--- a/autoload/neomake/makers/ft/c.vim
+++ b/autoload/neomake/makers/ft/c.vim
@@ -66,6 +66,7 @@ endfunction
 function! neomake#makers#ft#c#clangtidy() abort
     return {
         \ 'exe': 'clang-tidy',
+        \ 'args': ['%:p'],
         \ 'errorformat':
             \ '%E%f:%l:%c: fatal error: %m,' .
             \ '%E%f:%l:%c: error: %m,' .
--- a/autoload/neomake/makers/ft/c.vim
+++ b/autoload/neomake/makers/ft/c.vim
@@ -8,8 +8,10 @@ function! neomake#makers#ft#c#EnabledMakers() abort
 endfunction
 
 function! neomake#makers#ft#c#clang() abort
+    " as a single-file maker, include the current directory in the default
+    " search path
     return {
-        \ 'args': ['-fsyntax-only', '-Wall', '-Wextra'],
+        \ 'args': ['-fsyntax-only', '-Wall', '-Wextra', '-I./'],
         \ 'errorformat':
             \ '%-G%f:%s:,' .
             \ '%f:%l:%c: %trror: %m,' .
@@ -41,8 +43,10 @@ function! neomake#makers#ft#c#clangcheck() abort
 endfunction
 
 function! neomake#makers#ft#c#gcc() abort
+    " as a single-file maker, include the current directory in the default
+    " search path
     return {
-        \ 'args': ['-fsyntax-only', '-Wall', '-Wextra'],
+        \ 'args': ['-fsyntax-only', '-Wall', '-Wextra', '-I./'],
         \ 'errorformat':
             \ '%-G%f:%s:,' .
             \ '%-G%f:%l: %#error: %#(Each undeclared identifier is reported only%.%#,' .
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks, will review it more in detail tomorrow or so..

> -I./

What about using/preferring `neomake#utils#FindGlobFile('{config.h,someothertypical.h}') there? (so that it works from a subdirectory)

mqudsi:
Thanks, Daniel. I didn't know Neomake had the capacity to do that (it's very powerful!); but I did like the simplicity of just including nvim's current working directory instead of overcomplicating the rule.

What would be really nifty is if neomake could, at some point in the future, provide some sort of retry hooks that could take an error message from the linter and modify the maker args based off the text then retry the lint before bubbling the error back up to the user.

That way any time clang or one of the linters returns "could not locate include file xxxx.h", theoretically the file could be located and its directory added to the clang include path for the remainder of the session...

blueyed:
> I did like the simplicity of just including nvim's current working directory instead of overcomplicating the rule.

Yes, but the `cwd` might be something completely different, and therefore should at least be based on the current buffer.
What about using `-I%:h` to use the dir of the current buffer?
Otherwise `cwd` could also be set on the makers.

/cc @tweekmonster since he did something in this regard already IIRC.

The idea about adjusting linter args based on errors is nice.

mqudsi:
I'd be for cwd + `%:h`, but not for the latter alone. In dealing with `./configure` output, you often have a top-level `configure.h` but everything else under the existing source-code-controlled `src/` subdirectory.

blueyed:
@mqudsi 
Is this ready from your side?
I would review it again then.

blueyed:
Hopefully! ;)

(aggressively merging things now)

mqudsi:
@blueyed yessir, this should be good. Sorry, I got your notification on my phone the first time and forgot to check again once I got to my computer.

Thanks again.

jasonwzz:
with the latest "--" change, clang-tidy no longer picks up the local compilation database anymore. Has anyone had a similar problem?

blueyed:
@jason0x43 
I am not using it myself, sorry.

mqudsi:
I'm looking into it.

mqudsi:
@jason0x43 Please open a new ticket and @mention me. The clang-tidy man page has this to say on the topic of compilation databases:

>When no build path is specified, a search for compile_commands.json will be attempted through all  parent  paths  of first  input  file

Build paths are specified by passing in `-p <build_path>`. We don't pass in `-p` (and it's an error to call `-p` without a parameter) so this should apply.

blueyed:
Isn't the for being passed twice now?
(https://github.com/neomake/neomake/pull/1488/files#diff-fc99e5a401c291fe364e9015c5a2deadR73)

A new issue with debug log would be good indeed.

mqudsi:
@blueyed I think your comment is missing a word, but the highlighted line passes in the path to the file currently being edited (in vim, `%:p`), whereas I was referring to a literal `-p` argument to `clang-tidy` which takes a directory to search for a compilation database in.

blueyed:
Fixed.
The art/file gets appended by Neomake by default, but maybe it detects this?!

jasonwzz:
FYI. A new issue is opened at https://github.com/neomake/neomake/issues/1596 .

PR #1489, opened 2017-08-02
Consider phpstan.neon to be a default project config

still-dreaming-1:
This makes the phpstan file maker automatically check for and use a phpstan.neon file in the current working directory as a phpstan configuration file if one exists. This is still considered a file maker because it is still intended to only analyze the PHP file you are editing, not the entire project.

I also went ahead and added an optional global config value for the strictness level.


blueyed:
You can use `neomake#utils#FindGlobFile` to look up the config file from the current buffer/dir.

autoload/neomake/makers/ft/php.vim
```diff
@@ -42,8 +42,19 @@ function! neomake#makers#ft#php#phpmd() abort
...
+    " Check for the existence of a default PHPStan project configuration file.
+    " Technically PHPStan does not have a concept of a default filename for a
+    " project configuration file, but phpstan.neon is the filename shown in the
+    " example in the PHPStan documentation, so this is the default name expected
+    " by Neomake.
+    let phpStanConfigFilePath = neomake#utils#FindGlobFile('phpstan.neon')
+    if !empty(phpStanConfigFilePath)
+        call extend(maker.args, ['-c', phpStanConfigFilePath])
+    endif
+    return maker
```
blueyed: I assume the config would override the `--level` in this case?
It might be easier to not support a custom `--level`, but only the config (if the level can be configured there) - or otherwise append `--level` at the end always (assuming that it overrides the config then)?!
But maybe `--level` overrides `-c` always already?

still-dreaming-1: I'm not sure if the config file has a level option or not, but it is definitely worth finding out. Even if it did, I'm not sure which one would have higher precedence. I will try to find the answer to that question. For now I just know that the --level option becomes required when you do specify a config. I'm not opposed to just letting people specify it in the config instead if that works. In that case it would make sense to pass a value of 0 when it becomes a required flag.

codeinabox:
This is great. Fantastic to see PHPStan support improving!

still-dreaming-1:
@codeinabox thanks for the encouragement. That gave me the needed motivation to work on this again.

still-dreaming-1:
Here is some information I found out about how you can put the level inside a config file:
https://github.com/phpstan/phpstan/issues/432

I am about to attempt to get the level to be honored from a config file. I will post back here what I find out. From the comments in that link it sounds like we might not need to pass the level option at all.

still-dreaming-1:
I can't get the developer's suggestions to work. What is more the way the neon file works seems way overly complicated as a way to specify the level. I think it is better to let neomake provide the level as the code is here, and then only use the neon file for other things.

blueyed:
Let's go with this then.
Thanks!

PR #1491, opened 2017-08-03
[php.phpstan] Disable progress output

sserbin:
Disables phpstan's progress output to not pollute the location list.


blueyed:
Thanks!

PR #1490, opened 2017-08-03
fix #1481.

tracyone:
remove unnecessary shellescape

PR #1497, opened 2017-08-06
tests: add "NeomakeSh on (emulated) Windows"

blueyed:
Ref: https://github.com/neomake/neomake/pull/1490


blueyed:
Should be covered already by the test above.

Issue #1506, opened 2017-08-07
JS standard maker includes superfluous warnings


tlvince:
Given a file (`Card.js`) that contains lint errors, the following is shown in the quickfix list:

```
src/common/Card.js|22 col 20 warning| Extra semicolon.                                                                                                                        
|| standard: Use JavaScript Standard Style (https://standardjs.com)                                                                                                           
|| standard: Run `standard --fix` to automatically fix some problems. 
```

### Expected behavior

Only the first line is reported (the `|| standard:` lines are superfluous).

### Steps to reproduce

<details>

```
"src/common/Card.js" 22L, 432C written                                                                                                                                        
Neomake [+1.60]: [15.-:1:1] Running makers: standard.                                                                                                                         
Neomake [     ]: Skipping User autocmd NeomakeJobInit: no hooks.                                                                                                              
Neomake [     ]: [15.9:1:1] Starting async job: ['standard', 'src/common/Card.js'].                                                                                           
Neomake [     ]: [15.9:1:1] cwd: /.                                                                                                                           
Neomake [     ]: [15.-:-:1] Restarting already running job (14.8) for the same maker.                                                                                         
Neomake [     ]: [14.8:1:1] Removing already finished job.                                                                                                                    
Neomake [     ]: [14.8:1:1] Cleaning jobinfo.                                                                                                                                 
Neomake [     ]: [14.-:-:1] Cleaning make info.                                                                                                                               
Neomake [     ]: Skipping User autocmd NeomakeJobInit: no hooks.                                                                                                              
Neomake [     ]: [15.10:1:1] Starting async job: ['standard', 'src/common/Card.js'].                                                                                          
Neomake [     ]: [15.10:1:1] cwd: /.                                                                                                                          
Neomake [     ]: Skipping User autocmd NeomakeJobStarted: no hooks.                                                                                                           
Neomake [+0.72]: [15.10:1:1] stdout: standard: ['The react/jsx-space-before-closing rule is deprecated. Please use the react/jsx-tag-spacing rule with the "beforeSelfClosing"
 option instead.', ''].                                                                                                                                                       
Neomake [+0.03]: [15.10:1:1] stderr: standard: ['standard: Use JavaScript Standard Style (https://standardjs.com)', ''].                                                      
Neomake [     ]: [15.10:1:1] stderr: standard: ['standard: Run `standard --fix` to automatically fix some problems.', ''].                                                    
[...]

PR #1508, opened 2017-08-07
js: standard: filter out verbose messages

blueyed:
Fixes https://github.com/neomake/neomake/issues/1506.


tlvince:
This works great, thanks!

> What do you think about using -v in args to get e.g. " (no-unused-expressions)" at the end of messages?

:+1:

blueyed:
Cool, thanks for the report and trying to fix it already!

PR #1523, opened 2017-08-18
Mypy: check untyped defs

zackhsi:
This flag makes mypy behave less surprisingly. A core part of mypy is
that it infers types and performs checks based on these inferences. With
this flag set, more code is covered.

>`--check-untyped-defs` is less severe than the previous option – it
> type checks the body of every function, regardless of whether it has
> type annotations. (By default the bodies of functions without
> annotations are not type checked.) It will assume all arguments have
> type Any and always infer Any as the return type.

http://mypy.readthedocs.io/en/latest/command_line.html


blueyed:
Thank you!

PR #1532, opened 2017-08-29
Use eslint error format for sass-lint

ianks:
sass-lint uses eslint under the hood, thus we can just use the same formatting that is used with js.


blueyed:
Thank you!

PR #1555, opened 2017-09-06
mypy: Remove '--follow-imports=skip' fixes #1528

### LINKED ISSUE #1528: mypy configuration improvement suggestion
Trying out neomake with Python, I've found that "python -m mypy --ignore-missing-imports --follow-imports=skip" breaks even local annotations that reference imported types. Removing the "--follow-imports=skip" shows these while still ignoring errors about missing imports. I therefor propose to drop the "--follow-imports=skip" part which according to the following comment was likely added because the author believed it to only remove the warning about missing imports.

https://github.com/neomake/neomake/blob/d9c1c8b27d1a8bd5a3f2f22882e43028754643ca/autoload/neomake/makers/ft/python.vim#L283

niklas88:
The '--ignore-missing-imports' switch is enough to ignore missing
imports, adding the skip breaks all annotations using imported types,
which isn't desireable in most use cases.

This is also more in line with the comment on replacing the old
--silent-imports with the new flags.

I have been testing this change with large Python code bases since
I opened the issue and have found mypy to work great.

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -280,10 +280,10 @@ function! neomake#makers#ft#python#vulture() abort
 endfunction
 
 " --fast-parser: adds experimental support for async/await syntax
-" --silent-imports: replaced by --ignore-missing-imports --follow-imports=skip
+" --silent-imports: replaced by --ignore-missing-imports
 function! neomake#makers#ft#python#mypy() abort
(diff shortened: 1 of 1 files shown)


blueyed:
Thank you.

blueyed:
I think `--follow-imports=silent` is a better option here for file-mode: it will only display results for the current file then.

PR #1579, opened 2017-09-11
Fix doc/neomake.txt / vimhelplint integration

blueyed:


blueyed:
All good - had another local doc file that triggered failures locally.

PR #1591, opened 2017-09-16
vimlint fix in s:configure_buffer

blueyed:


blueyed:
Valid find.. needs to be fixed instead.

blueyed:
2017-09-21-005918_611x504_scrot

As can be seen by this code coverage report.. :nerd_face:

PR #1647, opened 2017-10-08
README: use CircleCI badge

mhinz:
The TravisCI badge seems outdated. Now we have the choice between:

badge

and

badge

The PR uses the latter, since it better fits with other badges out there.


blueyed:
Thank you!

PR #1652, opened 2017-10-09
Add python2 support to mypy via pyenv

zackhsi:
Let me know if this is a good approach. Should be plug n play friendly.
Alternately, can add a global extra mypy args variable.

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -279,13 +279,15 @@ endfunction
 " --silent-imports: replaced by --ignore-missing-imports
 function! neomake#makers#ft#python#mypy() abort
     let l:args = ['--check-untyped-defs', '--ignore-missing-imports']
-    if !exists('s:python_version')
-        let s:python_version = split(split(system('python -V 2>&1'))[1], '\.')
-    endif
-    if !v:shell_error && s:python_version[0] == '2'
-        call add(l:args, '--py2')
-    endif
-
+    try
+        if !exists('s:python_version')
+            let s:python_version = split(split(system('python -V 2>&1'))[1], '\.')
+        endif
+        if s:python_version[0] == '2'
+            call add(l:args, '--py2')
+        endif
+    catch
+    endtry
     return {
         \ 'args': l:args,
         \ 'errorformat':
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -283,7 +283,7 @@ function! neomake#makers#ft#python#mypy() abort
         let s:python_version = split(split(system('python -V 2>&1'))[1], '\.')
     endif
     if !v:shell_error && s:python_version[0] == '2'
-      call add(l:args, '--py2')
+        call add(l:args, '--py2')
     endif
 
     return {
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/python.vim
```diff
@@ -278,8 +278,14 @@ endfunction
 " --fast-parser: adds experimental support for async/await syntax
 " --silent-imports: replaced by --ignore-missing-imports
 function! neomake#makers#ft#python#mypy() abort
+    let l:args = ['--check-untyped-defs', '--ignore-missing-imports']
+    let l:output = system('pyenv version')
+    if !v:shell_error && l:output[0] == '2'
```
blueyed: Would fail for version 20..!

autoload/neomake/makers/ft/python.vim
```diff
@@ -278,8 +278,14 @@ endfunction
 " --fast-parser: adds experimental support for async/await syntax
 " --silent-imports: replaced by --ignore-missing-imports
 function! neomake#makers#ft#python#mypy() abort
+    let l:args = ['--check-untyped-defs', '--ignore-missing-imports']
+    let l:output = system('pyenv version')
```
blueyed: Should get cached.

blueyed:
related: https://stackoverflow.com/questions/29980673/is-the-output-of-python-v-considered-to-be-stable-usable-as-an-api

zackhsi:
Sure, can move the command to python. Once that is done and caching is added, is this a reasonable thing to do?

blueyed:
I think we should rather have some general helpers for version checks then like Syntastic and ALE have them - I've stayed away from this until now, since Neomake is "neo" i.e. the latest.. ;)

I think it would be OK to have it here for now, and other modules (java IIRC) have stuff like that already.

blueyed:
Maybe `split(split(system('python -V 2>&1'))[1][0])` then?

autoload/neomake/makers/ft/python.vim
blueyed: OTOH caching should get skipped here maybe, since you might be switching virtualenvs / $PATH?!

autoload/neomake/makers/ft/python.vim
```diff
@@ -278,8 +278,16 @@ endfunction
 " --fast-parser: adds experimental support for async/await syntax
 " --silent-imports: replaced by --ignore-missing-imports
 function! neomake#makers#ft#python#mypy() abort
+    let l:args = ['--check-untyped-defs', '--ignore-missing-imports']
+    if !exists('s:python_version')
+        let s:python_version = split(split(system('python -V 2>&1'))[1], '\.')
+    endif
+    if !v:shell_error && s:python_version[0] == '2'
+      call add(l:args, '--py2')
```
blueyed: Should be 4 spaces I guess (checkqa build job failure).

blueyed:
Merged in fc97f11.

Missed the try/catch (https://github.com/neomake/neomake/pull/1652/commits/572b137bf27fca2d83242bf731d069292fa76c9c), but I think that is good - try/catch is not cheap in Vim anyway AFAIK.

zackhsi:
Thanks @blueyed ! 🎉

blueyed:
Thanks! :)

PR #1653, opened 2017-10-10
Fix ä inserted where it shouldn't

Kazy:


blueyed:
Thanks, was meant to be a `#` though.

Issue #1665, opened 2017-10-20
Temporary files should be expanded correctly with `%:p` in args


CptnKirk:
See https://github.com/neomake/neomake/pull/1666#issuecomment-338444276

Original issue:

Neomake can asynchronously run makers against dirty buffers.  This is exposed via enhanced append_file support.  The credo and dogma makers however still use the '%p' expansion, which returns the current filename.  Neomake will still trigger makers on buffer change, but the makers will scan the wrong file. 

### Expected behavior
Type as you go linting happens without having to save changes.

### Steps to reproduce
1. Open file.  
2. Make changes that should be warned.  
3. Wonder why nothing happens.

CptnKirk:
I have a fix.  Just needed an issue to link against.

blueyed:
Thanks.

It might actually indicate a bug also, so that it should get expanded to the tempfile really.

blueyed:
This should be fixed.
There are tests for this being added in 8ac0d09.

PR #1666, opened 2017-10-20
type as you go fix for credo and dogma #1665

CptnKirk:
I'm not sure why nobody noticed this was broken until now.  The fix is pretty straightforward. #1665


blueyed:
Thanks, append_file is the default though.
Might be worth blaming and CCing the author also.

blueyed:
Picked in a25af15, thanks.

Leaving https://github.com/neomake/neomake/issues/1665 open for investigation, since it should have worked without this PR, too.

PR #1668, opened 2017-10-22
Verilog fixes

rbino:
Fix an error caused by an extra space and make it possible to build files with locally defined external modules

### DIFF
--- a/autoload/neomake/makers/ft/verilog.vim
+++ b/autoload/neomake/makers/ft/verilog.vim
@@ -4,7 +4,7 @@ endfunction
 
 function! neomake#makers#ft#verilog#iverilog() abort
     return {
-                \ 'args' : ['-t null', '-Wall'],
+                \ 'args' : ['-tnull', '-Wall'],
                 \ 'errorformat' : '%f:%l: %trror: %m,' .
                 \ '%f:%l: %tarning: %m,' .
                 \ '%E%f:%l:      : %m,' .
--- a/autoload/neomake/makers/ft/verilog.vim
+++ b/autoload/neomake/makers/ft/verilog.vim
@@ -4,7 +4,8 @@ endfunction
 
 function! neomake#makers#ft#verilog#iverilog() abort
     return {
-                \ 'args' : ['-tnull', '-Wall'],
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/verilog.vim
```diff
@@ -4,7 +4,7 @@ endfunction
 
 function! neomake#makers#ft#verilog#iverilog() abort
     return {
-                \ 'args' : ['-t null', '-Wall'],
+                \ 'args' : ['-tnull', '-Wall', '-y./'],
```
blueyed: What does `-y./` do exactly?
Shouldn't it be rather be `%:h`, i.e. the directory of the file?
Or even a project root?

rbino: It includes the dir where other modules are defined. The use case is being able to compile a file even if a module is defined in a separate file.
Honestly, I "copied" it from the C maker that includes the current dir with `-I./`.
If you want I can replace it with `-y%:h`.

rbino:
@blueyed I've replaced `./` with `%:h`, let me know if it's ok

blueyed:
I assume it works for you?
Have you tried it from different dirs? (i.e. after `:cd` or `:lcd`)

There is an issue with tempfiles (https://github.com/neomake/neomake/issues/1665), i.e. when using automake.

rbino:
Yes, it works both with saved and dirty buffers. The previous version (with `./`) didn't work in a different dirr, the actual one works as expected even changing dir or opening a file in another dir.

blueyed:
I see.

Then actually fixing `%:h` to use the temporary file's dir would break it again.
I've only started to look into fixing/changing this though - not sure yet.
I might end up keeping it like it is, but add some new placeholders that would respect the temporary file.
What do you think?

For this PR you could also use `./` again, but then add `'cwd': '%:h'`.

rbino:
I've changed the PR to use `./` + `cwd`.

Regarding `%:h`, IMHO it makes sense to introduce new placeholders for the temporary file folder to avoid breaking plugins that rely on the fact that `%:h` is the folder in which the original file resides (as is probably expected in Vim).

blueyed:
Thank you!

PR #1693, opened 2017-10-31
loclist/qflist: use number of entries for window height

blueyed:
This will take into account the number of list entries (e.g. 3) when
creating a location list or quickfix window, and uses the minimum of
`g:neomake_list_height` and that.

Therefore with only 3 entries it will use 3 instead of 10 now by
default.

It is still a good idea to use something like
https://github.com/blueyed/vim-qf_resize in general, but this improves
the default behavior.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -875,7 +875,7 @@ function! s:restore_prev_windows() abort
 endfunction
 
 let s:ignore_automake_events = 0
-function! s:HandleLoclistQflistDisplay(jobinfo) abort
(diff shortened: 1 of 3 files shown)


blueyed:
@FabioAntunes 
So the actual issue is that it should use "virtual" (i.e. wrapped) lines (assuming you have `:set wrap` already?!) instead?!
This might not be trivial to get though.

btw: you can try `:call neomake#quickfix#enable()` to get a more condensed quickfix list (still an experimental feature).

PR #1719, opened 2017-11-09
improve perl errorformat

euclio:
Align the perl error format more closely with the one provided by `:compiler perl`.


blueyed:
Thanks!

Reminds me of vim-dispatch and how it uses existing compilers.
We could have some kind of similar magic after all probably, maybe even just on top of the existing "makeprg" magic.

PR #1720, opened 2017-11-09
only concat extramsg if nonempty

euclio:
This prevents a stray extra space from appearing in the error message.

### DIFF
--- a/autoload/neomake/makers/ft/perl.vim
+++ b/autoload/neomake/makers/ft/perl.vim
@@ -29,6 +29,6 @@ function! neomake#makers#ft#perl#PerlEntryProcess(entry) abort
     let extramsg = substitute(extramsg, '\\\$', '', '')
 
     if !empty(extramsg)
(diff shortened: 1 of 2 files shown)


autoload/neomake/makers/ft/perl.vim
```diff
@@ -27,5 +27,8 @@ endfunction
 function! neomake#makers#ft#perl#PerlEntryProcess(entry) abort
     let extramsg = substitute(a:entry.pattern, '\^\\V', '', '')
     let extramsg = substitute(extramsg, '\\\$', '', '')
-    let a:entry.text = a:entry.text . ' ' . extramsg
+
+    if !empty(extramsg)
+        let a:entry.text = a:entry.text . ' ' . extramsg
```
blueyed: Could use `.=` instead of `=`, but fine.

blueyed:
Would be nice to cover this with a test actually.
I might do so eventually, if you do not want to do it yourself.

blueyed:
Thank you!

PR #1721, opened 2017-11-09
Add Solhint checker

sohkai:
Adds a new checker for Solidity.

[Solhint](https://github.com/protofire/solhint) uses ESLint's formatters, so I've just reused them here.

Let me know if anything else needs to be added!

### DIFF
--- a/autoload/neomake/makers/ft/solidity.vim
+++ b/autoload/neomake/makers/ft/solidity.vim
@@ -11,9 +11,5 @@ function! neomake#makers#ft#solidity#solium() abort
 endfunction
 
 function! neomake#makers#ft#solidity#solhint() abort
-    return {
-        \ 'args': ['-f', 'compact'],
-        \ 'errorformat': '%E%f: line %l\, col %c\, Error - %m,' .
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/solidity.vim
```diff
@@ -9,3 +9,11 @@ function! neomake#makers#ft#solidity#solium() abort
             \ '%f:%l:%c: %t%s: %m',
         \ }
 endfunction
+
+function! neomake#makers#ft#solidity#solhint() abort
+    return {
+        \ 'args': ['-f', 'compact'],
+        \ 'errorformat': '%E%f: line %l\, col %c\, Error - %m,' .
+        \   '%W%f: line %l\, col %c\, Warning - %m,%-G,%-G%*\d problems%#'
+        \ }
```
blueyed: You could use `return neomake#makers#ft#javascript#eslint()` here instead.
This way it is more likely for fixes to go to both.

blueyed:

sohkai:
@blueyed added it to the wiki now :). Thanks for changing it to use eslint's errorformat for me too!

PR #1727, opened 2017-11-12
Neovim: use "detach" option with `jobstart()`

blueyed:
This helps with https://github.com/neomake/neomake/issues/1704, and
seems like a good idea in general.
Stopping active makes (i.e. jobs) gets done on `VimLeave` already, so
Neovim not stopping the jobs itself on exit seems to be OK.


blueyed:
Closing in favor of https://github.com/neomake/neomake/pull/1741.

PR #1733, opened 2017-11-15
Undo 712c3ac zsh>sh + shellcheck

davidosomething:
It isn't a superset of sh, it is a different shell
Also shellcheck does not have a --shell=zsh (anymore? since 2015 https://github.com/koalaman/shellcheck/blob/102683ab04bcb8a0a6f5d492e70fc8d3703edcc1/ShellCheck/Parser.hs#L2800-L2809)
https://github.com/koalaman/shellcheck/issues/809

being a superset enables shellcheck by default

todo:
- [ ] update wiki entry

### DIFF
--- a/autoload/neomake/makers/ft/zsh.vim
+++ b/autoload/neomake/makers/ft/zsh.vim
@@ -1,13 +1,10 @@
 " vim: ts=4 sw=4 et
 
-function! neomake#makers#ft#zsh#SupersetOf() abort
-    return 'sh'
-endfunction
-
 function! neomake#makers#ft#zsh#EnabledMakers() abort
     return ['zsh']
 endfunction
 
+" Note: newer versions of shellcheck do not support zsh.
 function! neomake#makers#ft#zsh#shellcheck() abort
     let maker = neomake#makers#ft#sh#shellcheck()
     let maker.args += ['--shell', 'zsh']


autoload/neomake/makers/ft/zsh.vim
```diff
@@ -1,19 +1,9 @@
...
-
 function! neomake#makers#ft#zsh#EnabledMakers() abort
     return ['zsh']
 endfunction
 
-function! neomake#makers#ft#zsh#shellcheck() abort
-    let maker = neomake#makers#ft#sh#shellcheck()
-    let maker.args += ['--shell', 'zsh']
-    return maker
-endfunction
```
blueyed: I would keep this.
https://github.com/neomake/neomake/commit/fbc0b9fb001071afbbba1a6aebcc0c0f49338697 was meant to allow using it, but I've not intended to have it enabled by default.

So I suggest to just remov the `SupersetOf` for now.

davidosomething: sounds good

blueyed:
Nice, thanks!

blueyed:
Please update the wiki.

btw: I think we should move the wiki back to doc in the long run.  What do you think=

davidosomething:
The wiki's all set since I only removed the superset and not the configuration

In the long run, yes, it is especially nice to have docs as part of the codebase when you use a plugin manager or git and review the plugin updates.

PR #1738, opened 2017-11-17
format_loclist_ok is actually just format_ok

davidosomething:
Consider changing in code though, i like format_loclist_ok better. Will open an alternative PR


blueyed:
Yes, should be changed in code I think - to match `format_quickfix_ok`.

davidosomething:
Preferred over #1738

PR #1740, opened 2017-11-17
Support nightly rust 1.23-Nov-11-2017

rusterize:
Cargo changed the format a little bit, and warnings were not detected properly. This is a quick patch to solve that.

### DIFF
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -47,7 +47,7 @@ function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
 
         let error = {'maker_name': 'cargo'}
         let code_dict = get(data, 'code', -1)
-        if code_dict is g:neomake#compat#json_null
+        if code_dict is g:neomake#compat#json_null || code_dict['code'][0] != 'E' || code_dict['code'][0] != 'W'
             if get(data, 'level', '') ==# 'warning'
                 let error.type = 'W'
             else


blueyed:
Thanks.
Do you have details on how it changed?

rusterize:
Unfortunately I have not seen any documentation , I made a diff (for a particular program) between rust stable and the new nightly and there several different json objects. However, what broke the previous maker was an entry like this:

```diff
-    "code": null,
+    "code": {
+      "code": "unused_variables",
+      "explanation": null
+    },

Where the `-` indicates the stable and the `+` the new nightly.
```

As far as I have found (I may be wrong) the only way to figure out the changes in the format would be to scan the sources. In order to make it as generic as possible I simply put a condition that would prevent breakage and use the `level` field instead.

blueyed:
/cc @lilydjwg @lucc @JelteF @fuine 
Any input from you?

lucc:
@blueyed I never used rust and did not contribute much to neomake (I use it though). Did you mix me up with someone?

blueyed:
@lucc 
Got you from the Git history of the file in question.

lilydjwg:
I don't see what is broken for neomake master and cargo 0.26.0-nightly (1d6dfea44 2018-01-26). I also didn't notice issues in the past. Neomake marks warnings and errors as expected (without this patch).

lucc:
@blueyed no, (sadly) my only commit here was 50124a705f5ad68c5a4e377c91b576beb07c7700 :)

blueyed:
@lucc 
Oh, sorry - I've meant @fuine (via ccc36f91, already mentioned before).

fuine:
@rusterize can you provide an example code for us to test against? I haven't developed in nightly for some time now, so I'm not sure how things look atm. AFAIK we decided to support the stable channel rather than beta/nightly so far.

wezm:
This change is now in stable rust. It results in neomake reporting everything as errors. Here is a sample *warning* (from rust 1.24.0):

```json
{
  "message": {
    "children": [
      {
        "children": [],
        "code": null,
        "level": "note",
        "message": "#[warn(unused_imports)] on by default",
        "rendered": null,
        "spans": []
      }
    ],
    "code": {
      "code": "unused_imports",
      "explanation": null
    },
    "level": "warning",
    "message": "unused import: `std::collections::HashSet`",
    "rendered": "warning: unused import: `std::collections::HashSet`\n  --> src/bin/add-url.rs:16:5\n   |\n16 | use std::collections::HashSet;\n   |     ^^^^^^^^^^^^^^^^^^^^^^^^^\n   |\n   = note: #[warn(unused_imports)] on by default\n\n",
    "spans": [
      {
        "byte_end": 342,
        "byte_start": 317,
        "column_end": 30,
        "column_start": 5,
        "expansion": null,
        "file_name": "src/bin/add-url.rs",
        "is_primary": true,
        "label": null,
        "line_end": 16,
        "line_start": 16,
        "suggested_replacement": null,
        "text": [
          {
            "highlight_end": 30,
            "highlight_start": 5,
            "text": "use std::collections::HashSet;"
          }
        ]
      }
    ]
  },
  "package_id": "read-rust 0.1.0 (path+file:///home/wmoore/Projects/read-rust)",
  "reason": "compiler-message",
  "target": {
    "crate_types": [
      "bin"
    ],
    "kind": [
      "bin"
    ],
    "name": "add-url",
    "src_path": "/home/wmoore/Projects/read-rust/src/bin/add-url.rs"
  }
}
```

Here is what an _error_ looks like:

```json
{
  "message": {
    "children": [],
    "code": {
      "code": "E0425",
      "explanation": "\nAn unresolved name was used.\n\nErroneous code examples:\n\n```compile_fail,E0425\nsomething_that_doesnt_exist::foo;\n// error: unresolved name `something_that_doesnt_exist::foo`\n\n// or:\n\ntrait Foo {\n    fn bar() {\n        Self; // error: unresolved name `Self`\n    }\n}\n\n// or:\n\nlet x = unknown_variable;  // error: unresolved name `unknown_variable`\n```\n\nPlease verify that the name wasn't misspelled and ensure that the\nidentifier being referred to is valid for the given situation. Example:\n\n```\nenum something_that_does_exist {\n    Foo,\n}\n```\n\nOr:\n\n```\nmod something_that_does_exist {\n    pub static foo : i32 = 0i32;\n}\n\nsomething_that_does_exist::foo; // ok!\n```\n\nOr:\n\n```\nlet unknown_variable = 12u32;\nlet x = unknown_variable; // ok!\n```\n\nIf the item is not defined in the current module, it must be imported using a\n`use` statement, like so:\n\n```\n# mod foo { pub fn bar() {} }\n# fn main() {\nuse foo::bar;\nbar();\n# }\n```\n\nIf the item you are importing is not defined in some super-module of the\ncurrent module, then it must also be declared as public (e.g., `pub fn`).\n"
    },
    "level": "error",
    "message": "cannot find function `asfd` in this scope",
    "rendered": "error[E0425]: cannot find function `asfd` in this scope\n  --> src/bin/add-url.rs:31:5\n   |\n31 |     asfd();\n   |     ^^^^ not found in this scope\n\n",
    "spans": [
      {
        "byte_end": 696,
        "byte_start": 692,
        "column_end": 9,
        "column_start": 5,
        "expansion": null,
        "file_name": "src/bin/add-url.rs",
        "is_primary": true,
        "label": "not found in this scope",
        "line_end": 31,
        "line_start": 31,
        "suggested_replacement": null,
        "text": [
          {
            "highlight_end": 9,
            "highlight_start": 5,
            "text": "    asfd();"
          }
        ]
      }
    ]
  },
  "package_id": "read-rust 0.1.0 (path+file:///home/wmoore/Projects/read-rust)",
  "reason": "compiler-message",
  "target": {
    "crate_types": [
      "bin"
    ],
    "kind": [
      "bin"
    ],
    "name": "add-url",
    "src_path": "/home/wmoore/Projects/read-rust/src/bin/add-url.rs"
  }
}
```

gdouezangrard:
Issue #1919 is related.

For reference, the output format is serialized right from the following struct (see: https://github.com/rust-lang/rust/blob/698b956a9fca9688632d617dd6d73cae834237a3/src/libsyntax/json.rs#L104):

```rust
struct Diagnostic {
    /// The primary error message.
    message: String,
    code: Option<DiagnosticCode>,
    /// "error: internal compiler error", "error", "warning", "note", "help".
    level: &'static str,
    spans: Vec<DiagnosticSpan>,
    /// Associated diagnostic messages.
    children: Vec<Diagnostic>,
    /// The message as rustc would render it.
    rendered: Option<String>,
}
```

The proposed changes do not fully address the issue:

* first of all, the checks should be for `E` **and** `W` as shown in the following diff (otherwise in the case we have an `EXXX` error code, we will also check the level which probably isn't what was intended):
```diff
- if code_dict is g:neomake#compat#json_null || code_dict['code'][0] !=# 'E' || code_dict['code'][0] !=# 'W'`
+ if code_dict is g:neomake#compat#json_null || (code_dict['code'][0] !=# 'E' && code_dict['code'][0] !=# 'W'`)
```

* even then, `note` and `help` are labeled as errors which is the fallback:

```vim
if get(data, 'level', '') ==# 'warning'
    let error.type = 'W'
else
    let error.type = 'E'
endif
```

Maybe we could just first check the `level` attribute which is mandatory for `error`, `warning`, `note` and `help`, and then fill the appropriate code number if the `code` attribute is present?

PR #1741, opened 2017-11-18
Add nvim_job_opts/vim_job_opts maker settings

blueyed:
This allows for configuring e.g. `pty=1` for makers.

Helps with https://github.com/neomake/neomake/issues/1704.


jamessan:
It'd be nice if this were documented. :) I was going to ask for something like this, since I didn't see anything about it in the help.

blueyed:
@jamessan 
Ok.  It was considered to be experimental, but I agree now.
What is your use case?

jamessan:
I have a maker that requires running in a pty.

Issue #1728, opened 2017-11-13
New Maker: languagetool


somini:
I have a POC maker for languagetool.
https://github.com/languagetool-org/languagetool

Here it is, a python wrapper and the Neomake definition.
https://gist.github.com/somini/26c845d4f7f6dc2630427a67cb371122
It's a bit ugly, but it works. It turn misspellings into errors with possible replacements in the message, and all other issues are turned into warnings.
It's the perfect maker to run async, since being Java it eats 300MB of RAM and takes several seconds for a couple dozen lines of text.

I'm certain it can be implemented in pure Vimscript, with the new JSON parsing capabilities, it's just that I couldn't find an existing maker that parsed JSON in Neomake.
Clippy does it, but I got lost in the code, and went for the simple version.

blueyed:
Interesting.
It would be nice to use its HTTP server (http://wiki.languagetool.org/http-server).
See #1551, where I've done this using `curl` then for the maker itself.

As for parsing the JSON, using Vimscript would be indeed better than using Python, but that could be done later.

somini:
I did some testing with the HTTP server. The first check is equally slow, but after everything gets into cache, it starts being bearable.
This also allows for multiple language checking without having to launch multiple instances. It still uses over 300MB of RAM, but it's much more efficient.

I will implement the curl version.

somini:
It should be something like this:
https://gist.github.com/somini/26c845d4f7f6dc2630427a67cb371122#file-text-vim
The JSON is retrieved from the server and parsed, and I transliterated some of the Python into vimscript, but there are still missing fields.
The main issue is that languagetool only spits out a byte offset and a length, no line/column number. Vim has `byte2line` and it works, but for the column line extra processing must be done.

blueyed:
Once you have the line, you could use `line2byte()` to get its offset, and from there the col then.

somini:
I did that yesterday, but it needs some more work because the offsets are sometimes off.

blueyed:
Yes, byte vs utf8/multibyte chars probably.

somini:
It almost works now.
https://gist.github.com/somini/26c845d4f7f6dc2630427a67cb371122#file-text-vim

The root cause for some offsets being incorrect is errors in the beginning of the line. I'm trying to hack something, but this needs more testing.

somini:
Sucess!
https://gist.github.com/somini/26c845d4f7f6dc2630427a67cb371122#file-text-vim
[...]

PR #1766, opened 2017-11-23
Add languagetool maker for HTTP server

somini:
This adds a new text maker for languagetool, using the HTTP server.

This is #1728.

Defaults to a server @ `localhost:8081`.

See https://gist.github.com/somini/b3421a3c89bd6a6716de1123e39cefad for
socket activation systemd configuration.

### DIFF
--- a/autoload/neomake/makers/ft/text.vim
+++ b/autoload/neomake/makers/ft/text.vim
@@ -27,3 +27,80 @@ function! neomake#makers#ft#text#writegood() abort
                 \ 'postprocess': function('neomake#makers#ft#text#PostprocessWritegood'),
                 \ }
 endfunction
+
+function! s:fn_languagetool_curl(jobinfo) abort dict
+    " XXX: update method to get filename!
+    let self.args = ['-s',
+                \ '--data-urlencode', printf('text@%s', fnameescape(fnamemodify(bufname(a:jobinfo.bufnr), ':p'))),
+                \ '--data-urlencode', printf('language=%s', get(split(&spelllang, ','), 0, 'en')),
+		\ '--data-urlencode', printf('motherTongue=%s', 'pt-PT'),
+                \ 'http://localhost:8081/v2/check']
+endfunction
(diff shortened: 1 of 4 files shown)


somini:
This is still a WIP, list of things to do:
- [ ] Allow server URL configuration. Just a dumb prefix to replace `http://localhost:8081/`, it's possible to run the service on HTTPS.
- [ ] Allow forcing language instead of reading from `spelllang`
- [ ] Allow configuration of "motherTongue". Default to nothing, to disable false friend detection.
- [ ] Tests

somini:
There's a public languagetool server with some hefty rate limits, but it should be good to test this.

Use:
```vim
let g:neomake_text_languagetool_curl_server = "https://languagetool.org/api"
```

somini:
Well, this works for ASCII text (1 character = 1 byte), but any non-ASCII text will break the offsets, making this worthless.
The languagetool output mentions an "offset", but it's not bytes, it's characters. The regular CLI output sends the correct line and column numbers, but the API won't.

That means this needs yet another overhaul.

somini:
I added a very ugly hack to that non-ASCII bug I found. I'll be using this for a while to see if it's usable.
It's a terrible piece of code, I'm almost ashamed of pushing this into a public repo.

It still fails on edge cases such as an error alone in a line.

blueyed:
Maybe it's easier to require/use Python for parsing/handling this?

somini:
I could include a small python script that used urllib to make HTTP requests, mangle the JSON into a better JSON and the vimscript function to put that into entries.

There's even precedent:
https://github.com/neomake/neomake/tree/master/autoload/neomake/makers/ft/java
https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/java.vim#L122-L137

It would even eliminate the dependency on cUrl, it could be used on Windows.

I think I will go that way instead of the current horrible code. I'll open a new PR when I have something to show.

PR #1772, opened 2017-11-30
Add option for cursormoved handle

### LINKED ISSUE #1769: Feature request: bubble info about cursor error
I just see the pic of jedi-vim, and there is a pic 

I think we can use `conceal` feature to implement a bubble to show the error info under the cursor, just like

and maybe use `echo` should be default, add a option like `g:neomake_cursor_hold_func` would be better,

```vim
if !empty(g:neomake_cursor_hold_func)
   call call(g:neomake_cursor_hold_func, ['cursor error info'])
else
   echo 'cursor error info'
endif
```

wsdjeg:
close #1769 

- [x] do not add autocmd if user do not need the feature

- [x]  add option for set cursor moved func

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -2444,9 +2444,16 @@ function! neomake#EchoCurrentError(...) abort
     call neomake#utils#WideMessage(message)
 endfunction
 
-function! neomake#CursorMoved() abort
-    call neomake#EchoCurrentError()
-endfunction
+if !empty(get(g:, 'neomake_cursormoved_func', ''))
+    " this func will be called with one argv ( the current error message)
+    function! neomake#CursorMoved() abort
+        call call(g:neomake_cursormoved_func, [neomake#GetCurrentErrorMsg()])
+    endfunction
+else
+    function! neomake#CursorMoved() abort
+        call neomake#EchoCurrentError()
+    endfunction
+endif
 
 function! s:cursormoved_delayed_cb(...) abort
     if getpos('.') == s:cursormoved_last_pos
--- a/plugin/neomake.vim
+++ b/plugin/neomake.vim
@@ -84,13 +84,15 @@ function! s:setup_autocmds() abort
         if !exists('*nvim_buf_add_highlight')
             autocmd BufEnter * call neomake#highlights#ShowHighlights()
         endif
-        if has('timers')
-            autocmd CursorMoved * call neomake#CursorMovedDelayed()
-            " Force-redraw display of current error after resizing Vim, which appears
-            " to clear the previously echoed error.
-            autocmd VimResized * call timer_start(100, function('neomake#EchoCurrentError'))
-        else
-            autocmd CursorMoved * call neomake#CursorMoved()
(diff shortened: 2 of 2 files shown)


plugin/neomake.vim
```diff
@@ -84,13 +84,15 @@ function! s:setup_autocmds() abort
...
-            autocmd CursorMoved * call neomake#CursorMoved()
+        if get(g:, 'neomake_handle_cursormoved', 1)
+            if has('timers')
+                autocmd CursorMoved * call neomake#CursorMovedDelayed()
+                " Force-redraw display of current error after resizing Vim, which appears
+                " to clear the previously echoed error.
+                autocmd VimResized * call timer_start(100, function('neomake#EchoCurrentError'))
+            else
+                autocmd CursorMoved * call neomake#CursorMoved()
+            endif
```
wsdjeg: maybe we should move all of there to autoload, no need to handle cursormove when neomake checker is called.

blueyed: Yes, could be done only when features are used - like I've tried with other autocommands before, e.g. the statusline.

blueyed:
I would prefer to keep the indentation changes out here (or at least in a separate commit).
(I have tons of stashes, and have the feeling that this causes unnecessary conflicts.)

autoload/neomake.vim
```diff
@@ -2444,9 +2444,16 @@ function! neomake#EchoCurrentError(...) abort
...
+if !empty(get(g:, 'neomake_cursormoved_func', ''))
+    " this func will be called with one argv ( the current error message)
+    function! neomake#CursorMoved() abort
+        call call(g:neomake_cursormoved_func, [neomake#GetCurrentErrorMsg()])
+    endfunction
+else
+    function! neomake#CursorMoved() abort
+        call neomake#EchoCurrentError()
+    endfunction
+endif
```
blueyed: I think it would be better to provide an API to get info about the current error (and errors in general then maybe), so that you can just have your own autocommand?!

wsdjeg:
ok,  sorry, I use `gg=G` in vim file, but you have change the vim's indent syntax file?

autoload/neomake.vim
wsdjeg: yeah, what I need is just a plugin API, now `neomake#GetCurrentErrorMsg()` is what I want, and I hope to disable the default autocmd for cursormoved.

blueyed:
I am getting the same result (I think) when using `gg=G`.
Would be nice to address, but let's keep it out for now (to avoid conflicts.

wsdjeg:
ok, I will remove the indent changes.

autoload/neomake.vim
blueyed: then maybe the following will work already?

```vim
au! neomake CursorMoved
au! neomake VimResized
```

I agree though that e.g. a separate augroup (`neomake_current_error`) would be easier / better.

wsdjeg:
I just push a new commit.  there are only few autocmd for neomake, should I separate them?

Issue #1736, opened 2017-11-17
Neomake should not create files in current repo.


wsdjeg:
here is what `git status` show:

```
On branch version-control
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   autoload/SpaceVim/default.vim

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	autoload/SpaceVim/.default.vim@neomake_6232_10.vim
	autoload/SpaceVim/.default.vim@neomake_6232_8.vim
	autoload/SpaceVim/layers/.VersionControl.vim@neomake_6232_3.vim
	autoload/SpaceVim/layers/VersionControl.vim

no changes added to commit (use "git add" and/or "git commit -a")
````

I think neomake should not create files like:

```
autoload/SpaceVim/.default.vim@neomake_6232_10.vim	
autoload/SpaceVim/.default.vim@neomake_6232_8.vim	
autoload/SpaceVim/layers/.VersionControl.vim@neomake_6232_3.vim
````

blueyed:
There are some issues / discussion regarding this already.

Normally it should not really be a problem, but mostly only if you run Neomake and then <kbd>Ctrl</kbd><kbd>z</kbd> Vim while it is running still etc - or in case it was not removed properly (which is a bug).

As a workaround you can add `.*@neomake*` to your global (git) ignore file.

blueyed:
Related issues: https://github.com/neomake/neomake/issues/1212 https://github.com/neomake/neomake/issues/1595 https://github.com/neomake/neomake/issues/1617

wsdjeg:
@blueyed This happens when I use `:wq` in neovim. ok, I will update gitignore of my projects.

blueyed:
It should not happen though: it should get cleaned up on VimLeave - but maybe this is related to #1617 then.
Does it happen always? Do you have steps to reproduce it?

wsdjeg:
no, I do not have reproduce step, but it only sometimes happens when I use `:wq`.

chancez:
Can we get a config flag to tell neomake where to store tmpfiles? I already put files like this in `~/.vim/tmp/{backup,swap,undo}` and would prefer to be able to configure neomake to use something like `~/.vim/tmp/neomake`.

blueyed:
Yes, would make sense.
[...]

Issue #1617, opened 2017-09-25
temporary files might be left around


blueyed:
This can happen if post-processing gets into the action queue, but the window for it never re-appears.

TODO:
- [ ] should ensure that all temporary files are removed on VimLeave
- [ ] check if tempfiles can be removed before postponing cleaning of make info
- [ ] reloading autoload/neomake.vim should not leave the tempfiles, but either keep the action queue (or tempfile) info, or remove them when reloading happens

PR #1788, opened 2017-12-17
Add setting for custom tempfile directory

blueyed:
Fixes https://github.com/neomake/neomake/issues/1736.

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -554,17 +554,13 @@ function! s:command_maker_base._get_tempfilename(jobinfo) abort dict
 
         let dir = neomake#utils#GetSetting('tempfile_dir', self, '', a:jobinfo.ft, a:jobinfo.bufnr)
 
-        let bufname = bufname(a:jobinfo.bufnr)
-
         " Use absolute path internally, which is important for removal.
-        let orig_file = neomake#utils#fnamemodify(a:jobinfo.bufnr, ':p')
+        let orig_fname = neomake#utils#fnamemodify(a:jobinfo.bufnr, ':p')
         if empty(dir)
-            if empty(bufname)
-                let dir = tempname()
-            elseif empty(orig_file)
+            if empty(orig_fname)
                 let dir = tempname()
             else
-                let dir = fnamemodify(orig_file, ':h')
+                let dir = fnamemodify(orig_fname, ':h')
                 if filewritable(dir) != 2
                     let dir = tempname()
                     let s:make_info[make_id].tempfile_dir = dir
@@ -572,14 +568,12 @@ function! s:command_maker_base._get_tempfilename(jobinfo) abort dict
                 endif
             endif
 
-            if empty(bufname)
+            if empty(orig_fname)
                 let filename = 'neomaketmp.'.a:jobinfo.ft
-            elseif empty(orig_file)
-                let filename = fnamemodify(bufname, ':t')
             else
-                let filename = fnamemodify(orig_file, ':t')
+                let filename = fnamemodify(orig_fname, ':t')
                             \ .'@neomake_'.s:pid.'_'.make_id
-                let ext = fnamemodify(orig_file, ':e')
+                let ext = fnamemodify(orig_fname, ':e')
                 if !empty(ext)
                     let filename .= '.'.ext
                 endif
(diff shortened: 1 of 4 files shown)


autoload/neomake.vim
```diff
@@ -524,22 +524,31 @@ function! s:command_maker_base._get_tempfilename(jobinfo) abort dict
...
-            let temp_file = tempname() . slash . 'neomaketmp.'.a:jobinfo.ft
-        else
-            let orig_file = neomake#utils#fnamemodify(a:jobinfo.bufnr, ':p')
-            if empty(orig_file)
+
+        let orig_file = neomake#utils#fnamemodify(a:jobinfo.bufnr, ':p')
+        if empty(dir)
+            if empty(bufname)
+                let dir = tempname()
+            elseif empty(orig_file)
```
chancez: Now it looks like this conditional could be combined with the `if empty(bufname)`, since the body of both just set `let dir = tempname()`

autoload/neomake.vim
```diff
@@ -524,22 +524,31 @@ function! s:command_maker_base._get_tempfilename(jobinfo) abort dict
...
                 let dir = tempname()
-                let filename = fnamemodify(bufname, ':t')
-                let s:make_info[make_id].tempfile_dir = dir
```
chancez: I see `let s:make_info[make_id].tempfile_dir = dir` is removed but not added later.

autoload/neomake.vim
```diff
@@ -524,22 +524,31 @@ function! s:command_maker_base._get_tempfilename(jobinfo) abort dict
...
             else
                 let dir = fnamemodify(orig_file, ':h')
                 if filewritable(dir) != 2
                     let dir = tempname()
                     let s:make_info[make_id].tempfile_dir = dir
```
chancez: Related to the above comment: `let s:make_info[make_id].tempfile_dir = dir` got removed above, should this be done later down?

autoload/neomake.vim
blueyed: Better to have it split for coverage reporting.
It showed that `orig_file` is typically not empty if `bufname` is not at least.

autoload/neomake.vim
blueyed: It was not used elsewhere, but should maybe (#1617).

blueyed:
@chancez 
Thanks for your review already.

I think the default behaviour could be changed to use a temporary directory in the project root (something like `.neomaketmp.$PID.$MAKE_ID`).
Might do that in a follow-up PR then though.

chancez:
@blueyed yeah that would be a reasonable feature. I think being able to customize the location globally is a good first step though and will fit many needs.

chancez:
@blueyed How do you want to proceed with this? I'm glad to do any work needed on this, as I'm currently running with my patch and would love this to be handled upstream.

blueyed:
@chancez 
IIRC I've some local stash (different from this PR) for this.
Need to come back to it.

b-ggs:
Any news on this? Would love to have this feature in the next release

nathanowen42:
When will this be in?  I am having an issue where clangcheck fails due to it detecting the neomake temp files as headers.  This is causing my linter to only work once (first time it runs) so I have to close the file and re-open to see errors.

chancez:
I would also really like the feature, as I currently maintain a fork of neomake rebased with this feature on top. If others require this feature, I advise you do the same until it's supported.

blueyed:
@nathanowen42 
So the tempfiles do not get removed?  Or is it still running then?  Please consider creating a new issue with a logfile.

@chancez 
Rebased for now.  Will work on this shortly again I hope.

blueyed:
Note that you can specify `tempfile_name` already: https://github.com/neomake/neomake/blob/00b17b5a09d1641780afade98c88a8ecff61c12c/doc/neomake.txt#L442-L444

chancez:
Awesome! Thanks.

blueyed:
@chancez 
You're welcome!

How are you using it?

chancez:
I can't even remember _why_ I needed it anymore except I believe I was finding some of these left over occassionally, or tags may have been using them. Anyway, I'm setting `let g:neomake_tempfile_dir = '/Users/chance/.vim/tmp/neomake'` to ensure they live in a place outside my project.

nathanowen42:
> @nathanowen42
> So the tempfiles do not get removed? Or is it still running then? Please consider creating a new issue with a logfile.
> 
> @chancez
> Rebased for now. Will work on this shortly again I hope.

@blueyed They were being removed but the names were interfering with my build environment.  I made some changes to the config and am now running clang through docker and am now having a different issue (no doubt my fault since I have a complex series of scripts) where they are staying behind after I close vim.  I will need to debug that.  let g:neomake_tempfile_dir = '/tmp/' does not move them to the tmp directory sadly.  I managed to change the name (via putting in a hack to change the name, will likely remove that and switch to your method) but it would be nice to have the ability to re-route them to the tmp directory so they go away after a reboot if something goes wrong.

blueyed:
@nathanowen42 
Then maybe some additional config is used that overrides this?
Please consider creating a new issue with a logfile.
Just set `let g:neomake_logfile = '/tmp/neomake.log'` and search there for "temp" when it happens, and/or post that with the new issue.

PR #1810, opened 2018-01-07
Add eslint maker for .json

willxy:
I don't know if this is something useful for others or not:

I wanted to use eslint for my json files. Maker definitions copied from `autoload/neomake/makers/ft/javascript.vim/javascript.vim` but adding `--plugin json` to `args`.

### DIFF
--- a/autoload/neomake/makers/ft/json.vim
+++ b/autoload/neomake/makers/ft/json.vim
@@ -23,3 +23,15 @@ function! neomake#makers#ft#json#jsonlint() abort
             \ '%-G%.%#'
         \ }
 endfunction
+
+function! neomake#makers#ft#json#eslint() abort
+    let maker = neomake#makers#ft#javascript#eslint()
+    let maker.args += ['--plugin', 'json']
+    return maker
+endfunction
+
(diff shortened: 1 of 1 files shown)


willxy:
That makes sense.  Is there an example of that done elsewhere?

Or, something like this?  I didn't know if there's a way to do this more idiomatically, without the interim variable.

```
function! neomake#makers#ft#json#eslint() abort
    let l:cmd = neomake#makers#ft#javascript#eslint()
    let l:cmd.args += ['--plugin', 'json']
    return l:cmd
endfunction
```

blueyed:
Looks fine, but please use just "maker" for the var (without `l:` prefix).

See https://github.com/neomake/neomake/blob/fe456921fe5eba6fd95fa4542cc3e45384e00d01/autoload/neomake/makers/ft/cpp.vim#L9-L14.

To avoid the var you could use `return extend(…)` directly, but it's less readable.

willxy:
Perfect -- Thanks.  Updated.

blueyed:
Thank you!

willxy:
My pleasure, thank you!

PR #1821, opened 2018-01-17
set warning severity for phpcs

tvestelind:
I needed to be able to control how much warnings to show. I hope this is the correct way of doing that.

### DIFF
--- a/autoload/neomake/makers/ft/php.vim
+++ b/autoload/neomake/makers/ft/php.vim
@@ -26,6 +26,11 @@ function! neomake#makers#ft#php#phpcs() abort
         call add(l:args, '--standard=' . expand(g:neomake_php_phpcs_args_standard))
     endif
 
+    "Add warning severity argument if one is set.
+    if exists('g:neomake_php_phpcs_args_warning_severity')
+        call add(l:args, '--warning-severity=' . expand(g:neomake_php_phpcs_args_warning_severity))
+    endif
+
     return {
         \ 'args': args,
         \ 'errorformat':


autoload/neomake/makers/ft/php.vim
```diff
@@ -26,6 +26,11 @@ function! neomake#makers#ft#php#phpcs() abort
         call add(l:args, '--standard=' . expand(g:neomake_php_phpcs_args_standard))
     endif
 
+    "Add warning severity argument if one is set.
+    if exists('g:neomake_php_phpcs_args_warning_severity')
+        call add(l:args, '--warning-severity=' . expand(g:neomake_php_phpcs_args_warning_severity))
+    endif
+
```
blueyed: You could just do `let neomake_php_phpcs_arg = neomake#makers#ft#php#phpcs().args += ['--warning-severity=whatever']`.
I see that there is `g:neomake_php_phpcs_args_standard` already, and it is not necessary in the same way.

Please report back if that works for you.

(there is some idea about having a `post` setting for args, see https://github.com/neomake/neomake/issues/972#issuecomment-275800406)

tvestelind: Hi,
I tried to add it to my config file but that didn't work. Is that how you mean that I should do it?
The error that I got is:
```
E15: Invalid expression: = ['--warning-severity=0']
E15: Invalid expression: neomake#makers#ft#php#phpcs().args += ['--warning-severity=0']
```

blueyed: Sorry, should be `+` only. Fixed above.

blueyed: @tvestelind 
Does this work for you?

tvestelind: Hi,
Sorry for taking so long.
Yea, it seems to work. Is that they way you prefer configuring tools?

Issue #1795, opened 2017-12-27
Add process_json hook


blueyed:
Similar to `process_output`, but gets a dict (parsed from JSON output).
This avoids duplicating boilerplate for parsing/handling JSON, and could handle e.g. un-buffered output, so that only valid JSON is parsed / passed on.

PR #1829, opened 2018-01-22
[blocked upstream] parse tslint JSON output

### LINKED ISSUE #1783: Typescript TSLint ignores warnings
### Expected behavior

When TSLint has a warning, that's displayed as a warning within Neovim.

### Steps to reproduce

1. Set up TSLint for a project
2. Make a change to a file that results in a warning.
3. Nothing shows up.

### The Problem

The problem seems to be in this [line of code](https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/typescript.vim#L27). It looks like the code is only looking for `ERROR` in the output and ignoring lines that begin with `WARNING`. I fixed it locally by changing the code to this:

```
\ 'errorformat': '%EERROR: %f[%l\, %c]: %m,%WWARNING: %f[%l\, %c]: %m',
```

I'd be happy to contribute a PR unless you guys would like to do it. 😃

euclio:
This passes warning and span information on to neomake.

Fixes #1783.


euclio:
Hm, this isn't quite working yet. Looks like tslint doesn't like temporary files.

blueyed:
What is the problem with tempfiles?
I assume that it is because of `'filename': data.name`?

I am seeing "No valid rules have been specified" on stderr with the minimal example I have for testing.

When using `'output_stream': 'stdout'` this gets transformed into an error from Neomake: `Neomake: tslint: unexpected output on stderr: No valid rules have been specified`, and otherwise the decoding of the JSON will fail.
Therefore having `output_stream` makes sense.

btw: there is an open issue for tslint to handle stdin, which would help here: https://github.com/palantir/tslint/issues/1590.

blueyed:
My `tsconfig.json`:
```json
{ "rules": {
    "noUnusedParameters": true
} }
```

Tried an empty one also.  What am I missing?

euclio:
It's not easy to reproduce, but I was getting an intermittent error where tslint would complain that the temp file isn't part of the project, which would cause the JSON parsing to blow up.

blueyed:
Ah, similar to what I was seeing then.
Can you help me with getting a test case for this, i.e. a `tsconfig.json` that should work?

euclio:
Ah, got it to happen. Use this tsconfig.json:

```
{
    "files": [ "t.ts" ]
}
```

And then open t.ts and start typing a bunch. Without `tempfile_enabled: 0`, you'll get an error.

blueyed:
Still getting "No valid rules have been specified"..
Using tslint 5.9.1 on Arch Linux.

euclio:
I'm testing with this tslint.json:

```
{
    "defaultSeverity": "error",
    "extends": [
        "tslint:recommended"
    ],
    "jsRules": {},
    "rules": {
        "no-console": {
            "severity": "warning",
            "items": ["log"]
        }
    },
    "rulesDirectory": []
}
```

blueyed:
That works.

Your error is because it is not included in "files" it seems.

> X is not included in project.

It seems like using `-p` enforces this check..
Adding `"include": [ ".*.ts" ]` to tsconfig.json seems to help here - and appears to be required for an empty "files" setting (to explicitly include hidden files).

Not sure how to proceed here really, but disabling tempfiles is a bid bad, since automaking (without saving) is not possible then.

euclio:
Perhaps disable tempfiles by default and document that you can reenable it to fix automaking if you have `".*.ts` in your tsconfig?

blueyed:
Or not using `-p` by default.. what is the reason for this again?  To have typechecks?  Aren't those provided via `tsc` already?

euclio:
That flag determines the list of files to be linted, and also enables some additional checks that require type information. Imo, it should be enabled if available.

euclio:
Sounds like https://github.com/palantir/tslint/issues/1590 is really the best solution, though.

blueyed:
Yeah.

> the list of files to be linted

IIRC tslint ignores this for when an argument is passed, except for the check apparently if it is in the project.. :/

But I see that it is good to keep it then.

A hack coming to mind would be to create a temp `tsconfig.json` that would add `.*@neomake*.ts` to `include` - not that trivial though.. ;)

I agree that disabling tempfiles with `-p` is good for now then, and document the workaround.
Although we do not really have documentation for makers itself currently AFAIR - might be the best adding it to the wiki then for now.

I would also like to test
1. how the "unexpected output" behaves with automaking - if it shows up all the time that would be annoying
2. how disabled tempfiles interact with automaking: it should just skip the automaking then with a debug msg, possibly warning once.

blueyed:
Pushed some fixup (inline the JSON/output), and covered the usage of tsconfig.json.

blueyed:
btw: I plan to add a wrapper that helps with parsing JSON output: https://github.com/neomake/neomake/issues/1795 (can be done later for this though).

euclio:
What remains to be done here?

blueyed:
The tests mentioned above, but not really related to the PR.
I will have to look into it, might take me a few days.

euclio:
Looks like the unexpected output will pop up whenever you edit a file not in the project. Probably best to just test if the first character is `[`.

blueyed:
We can use `filter_output` to get rid of it.

blueyed:
`process_json` has been merged by now (#1848), so could (and should) be used here.

euclio:
Awesome. Added them in the latest commit.

davidosomething:
i wonder if recently merged #2010 json updates will help here?

euclio:
@davidosomething This is effectively blocked on https://github.com/palantir/tslint/pull/3816

blueyed:
I am closing this for now, given that there appears to be no work done upstream for this.
Happy to re-open later, of course.

PR #1838, opened 2018-01-28
Add languagetool maker for HTTP API

somini:
This is a continuation of #1766.

It uses a python3 script to mangle the LanguageTool API into the correct
offsets for vim.
https://github.com/neomake/neomake/pull/1766#issuecomment-348602778

It is much more robust than the previous attempts.


somini:
This is still a WIP:
- [x] Allow server URL configuration. Just a dumb prefix to replace `http://localhost:8081/`, it's possible to run the service on HTTPS.
- [x] Allow forcing language instead of reading from `spelllang`
- [x] Allow configuration of "motherTongue". Default to nothing, to disable false friend detection.
- [x] Support ["preferredVariants"](https://languagetool.org/http-api/swagger-ui/#!/default/post_check). Only for "language=auto"
- [ ] Tests
- [ ] Documentation

somini:
To test it without needing to install LanguageTool locally, use:
```
let g:neomake_text_languagetool_server = "https://languagetool.org/api"
```

This is not the default because that would send all the plaintext of `text` buffer to an external server.

somini:
Rebased this to the latest master.

blueyed:
Note: removed some useless tests in master in https://github.com/neomake/neomake/pull/1849 - where you had to adjust them.

somini:
Rebased this to avoid those conflicts.

I see that there will be a `process_json` hook which might be interesting to use since I don't handle a malformed JSON output, but that can be punted to another PR for improvements.

somini:
The failing test is a unused argument to the `'fn'` function.

blueyed:
Yeah, vimlint is kind of annoying.
Just ignore this for now - I will improve the CI around vimlint eventually.

Thanks for keeping up with this.
Seeing Python code in here reminds me of my early idea to move most of the processing to Python after all.. :/ - maybe we'll end there in the end, but for now I think that having at least the core in Vimscript is good.

blueyed:
It would be nice indeed btw if codecov could be made to still report when only "checkqa" is failing - or maybe even in general.

somini:
Having the core in vimscript is good enough for most applications. The GCC output format is used in many places.
This kind of ugly hacks are Good Enough, I don't think there is a need to complicate the plugin to make edge cases easier to code.

somini:
Rebased to master

somini:
Since I'm rebasing, clear the worthless commit and fix typos.

somini:
Rebased on master.

somini:
Rebased on master.

autoload/neomake/makers/ft/text.vim
```diff
@@ -27,3 +35,51 @@ function! neomake#makers#ft#text#writegood() abort
                 \ 'postprocess': function('neomake#makers#ft#text#PostprocessWritegood'),
                 \ }
 endfunction
+
+let s:languagetool_fallback_language = 'auto'
+" Public API: https://languagetool.org/api
```
blueyed: 404.
Should be http://wiki.languagetool.org/public-http-api probably.

somini: That is the actual value you can set the variable. I'll rephrase it.

See https://github.com/somini/neomake/blob/a52514fe3f89348026280efc9a9714820c43db59/autoload/neomake/makers/ft/text/languagetool.py#L80

somini:
Rebased on master.

blueyed:
Thanks for keeping up with it.
Pushed a fixup to fix linting.

somini:
Rebased to master. Includes #1997 .

somini:
Rebased yet again.

The vim 7 tests are failing.

blueyed:
> The vim 7 tests are failing.

    (297/596) [EXECUTE] Instantiate all default makers
        (297/596) [EXECUTE] (X) Vim(let):E117: Unknown function: s:fn_languagetool
          > in neomake#core#instantiate_maker (line 18)
          >   let returned_maker = call(Init, [options], maker)
          > in neomake#debug#validate_maker
          > /home/neomake/repo/tests/makers.vader:128

Would have to either be an autoload function, or a dict function on the maker
directly.  (it is because of https://github.com/somini/neomake/blob/6e994292711fc03ec92dc2a5e5c82fc03f23fc26/autoload/neomake/utils.vim#L359-L368)

Pushed two fixups.  Feel free to squash/reject etc.. :)

autoload/neomake/makers/ft/text.vim
```diff
@@ -27,3 +34,51 @@ function! neomake#makers#ft#text#writegood() abort
                 \ 'postprocess': function('neomake#makers#ft#text#PostprocessWritegood'),
                 \ }
 endfunction
+
+let s:languagetool_fallback_language = 'auto'
+" See http://wiki.languagetool.org/public-http-api for a public instance. Use:
+"   :let g:neomake_text_languagetool_server = 'https://languagetool.org/api'
+let s:languagetool_fallback_server = 'http://localhost:8081'
```
blueyed: Those could go into the function directly.

autoload/neomake/makers/ft/text.vim
```diff
@@ -27,3 +34,51 @@ function! neomake#makers#ft#text#writegood() abort
...
+    endif
+    let preferredVariants = s:getVar('neomake_text_languagetool_curl_preferredVariants', '')
+    if !empty(l:preferredVariants) && l:language ==# 'auto'
+        for var in l:preferredVariants
+            let args += ['--preferredVariants', var]
+        endfor
+    endif
+    let l:args += [l:server, l:language]
+    let self.args = l:args
+endfunction
```
blueyed: Instead of extra args this should could just rely on `g:/b:neomake_text_languagetool_args`, and print an error if not set?!

blueyed:
btw: you should also remove the `l:` prefixes.. ;)  (https://github.com/neomake/neomake/pull/2134)

autoload/neomake/makers/ft/text.vim
somini: Done!

autoload/neomake/makers/ft/text.vim
somini: The args are composed of the rest of the settings. How would I compose them, in the same function?

somini: Is the `neomake#utils#GetSetting` still the "recommended" way of accessing the settings? How can I use it, it has a bunch of required arguments I wouldn't want to hardcode.

blueyed: Yeah, `neomake#utils#GetSetting` is still the "recommended" way, since it handles both old-style and new-style settings.
The newer API should be saner, but then you would have to require new-style settings.
You should look at existing usages of it - typically you are not hard-coding settings, but e.g. only passing in the maker.

> The args are composed of the rest of the settings. How would I compose them, in the same function?

I got the impression that there are just multiple args that are all added up, i.e. you would just require `g:neomake_text_languagetool_args` to be set - no defaults there.

This gets automatically turned into arguments for the maker - i.e. there is only one required setting and that is `…_args`.  You should probably just check in InitForJob if `args` is empty, and then suggest that `g:/b:neomake_text_languagetool_args` should be configured.

autoload/neomake/makers/ft/text.vim
```diff
@@ -48,11 +48,11 @@ function! s:languagetool_maker.InitForJob(_jobinfo) abort
                 \ get(split(&spelllang, ','), 0,
                 \ s:getVar('neomake_text_languagetool_fallbackLanguage', 'auto') ) )
     " Optional Arguments
-    let motherTongue = s:getVar('neomake_text_languagetool_curl_motherTongue', '')
+    let motherTongue = s:getVar('neomake_text_languagetool_motherTongue', '')
     if !empty(motherTongue)
         let args += ['--motherTongue', motherTongue]
     endif
```
blueyed: E.g. here there is no need for a specific setting, just let users configure `--motherTongue` in args themselves.

somini:
Rebased this on master once again. I'm still using this :grin:.

There are still some pending issues, I can take a look at them later this week.

somini:
It seems the API changed in the meantime...

```
(346/677) [EXECUTE] (X) Vim(let):E716: Key not present in Dictionary: ft, a:jobinfo.bufnr)
      > in {11} (autoload/neomake/makers/ft/text.vim:43)
      >   let server = neomake#utils#GetSetting('server', s:languagetool_maker, 'http://localhost:8081', a:jobinfo.ft, a:jobinfo.bufnr)
      > in neomake#core#instantiate_maker (autoload/neomake/core.vim:52)
      >   let returned_maker = call(Init, [options], maker)
      > in neomake#debug#validate_maker (autoload/neomake/debug.vim:64)
      >   let maker = neomake#core#instantiate_maker(a:maker, {}, 0)
      > /home/neomake/repo/tests/makers.vader:188: let issues = neomake#debug#validate_maker(extend(maker, {'exe': 'true'}))
```

Odd, it works for me locally.

nomasprime:
Oh no, what happened to this? Really nice idea.

somini:
> Oh no, what happened to this? Really nice idea.

I'm not using this anymore, but I want to.

@blueyed What's the status of getting the settings and the like, were there any overhauls in the rest of the plugin? Should I start this over, or can it be salvaged?

PR #1843, opened 2018-02-01
Fix an eslint_d issue

sei40kr:
# Changes

Fix an issue eslint_d invokes global ESLint for local project files.


blueyed:
Necessary for eslint then also I guess?!

blueyed:
Found an old branch where it gets merged: https://github.com/neomake/neomake/pull/1844.
Do you agree that `cwd` should be used for both?

sei40kr:
@blueyed 
I read eslint-cli source code and diff in the PR above. And I agree that.

blueyed:
Thanks!

PR #1855, opened 2018-02-15
Add additional errorrformat for gometalinter

joshuarubin:
Without this additional format, results from some linters (like [gas](https://github.com/GoASTScanner/gas)) are ignored


blueyed:
Thanks!
Do you have some example output so that this can actually get tested?

blueyed:
Don't worry about the test itself, I will add it then - but example output is required.

joshuarubin:
`git.go:17::warning: Subprocess launching with variable.,HIGH,HIGH (gas)`

blueyed:
Thanks!

Issue #1861, opened 2018-02-19
Statusline functions don't show running checks when Neomake is disabled


somini:
It is possible to `NeomakeDisable*` and then run Neomake manually. This is mentioned in the docs.

When Neomake is disabled, the statusline changes to an unconditional `$SCOPE-`, with scope being the disable scope.
This is triggered by this code:
https://github.com/neomake/neomake/blob/dbd873a988af116017fba2bfffcb27ac92097d9a/autoload/neomake/statusline.vim#L271-L275

This means when Automake is disabled and you run a check manually, there's no feedback.

### Expected behavior

The same feedback for running a check when automake is enabled should be available when automake is disabled.

### Steps to reproduce

1. Disable automake for a buffer: `NeomakeDisableBuffer`
2. Verify the statusline string is now "b-"
1. Run Neomake manually
1. Verify the statusline shows no feedback, as opposed to the behaviour with automake enabled

blueyed:
https://github.com/neomake/neomake/pull/1865 would fix this also, without adding new formats etc - which might be good base for #1862 then.
Please let me know what you think.

PR #1862, opened 2018-02-19
Improve statusline customization

somini:
This is something that can solve #1861

This is a breaking change, it might be useful to review this with extra
care.

### DIFF
--- a/autoload/neomake/statusline.vim
+++ b/autoload/neomake/statusline.vim
@@ -143,7 +143,18 @@ let s:formatter = {
             \ }
 function! s:formatter.running_job_names() abort
     let jobs = get(self.args, 'running_jobs', s:running_jobs(self.args.bufnr))
-    return join(map(jobs, "v:val.name . (v:val.file_mode ? '' : '!')"), ', ')
+    let sep = get(self.args, 'running_jobs_separator', ', ')
+    let format_running_job_file = get(self.args, 'format_running_job_file', '%s')
+    let format_running_job_project = get(self.args, 'format_running_job_project', '%s!')
+    let formatted = []
+    for job in jobs
+        if job.file_mode
+            call add(formatted, printf(format_running_job_file, job.name))
+        else
+            call add(formatted, printf(format_running_job_project, job.name))
+        endif
+    endfor
+    return join(formatted, sep)
 endfunction
 
 function! s:formatter._substitute(m) abort
@@ -191,9 +202,13 @@ function! neomake#statusline#get_status(bufnr, options) abort
     if !empty(running_jobs)
         let format_running = get(a:options, 'format_running', '… ({{running_job_names}})')
         if format_running isnot 0
-            return s:formatter.format(format_running, {
-                        \ 'bufnr': a:bufnr,
-                        \ 'running_jobs': running_jobs})
+            let args = {'bufnr': a:bufnr, 'running_jobs': running_jobs}
+            for opt in ['running_jobs_separator', 'format_running_job_project', 'format_running_job_file']
+                if has_key(a:options, opt)
+                    let args[opt] = a:options[opt]
+                endif
+            endfor
+            return s:formatter.format(format_running, args)
         endif
     endif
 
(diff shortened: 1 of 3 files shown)


doc/neomake.txt
```diff
@@ -720,14 +720,25 @@ You can use the following options:
...
+    Default: not used / `'{{scope}}-%s'`
+  - format_status_enabled: used to wrap the whole status when enabled.
     This is a |printf()| format string, where `%s` gets replaced with the
     whole status (and any literal/non-printf `%` needs to be escaped as `%%`).
     Default: not used / `'%s'`
 
 You can use the following format placeholders:
   - `{{running_job_names}}}`: comma-separated list of running jobs (typically
     their maker names)
+  - `{{scope}}`: When disabled, the scope of the disable. One of "b", "t",
```
blueyed: disabled_scope to be more explicit.

blueyed:
Thanks.
What about just prepending the disabled info, with a space as separator?

This could get reflected through format strings also, but then having something like `format_disabled_info`, defaulting to `{{disabled_scope}}- ` would be better.

There might be use cases for `format_status_enabled` and `*_disabled`, but they should just default to `%s` then also, or rather `format_status`.  This way you can e.g. easily prefix them etc.
OTOH it makes sense to have `{{disabled_info}} %s` then as the default for `format_status_disabled` then.

I hope this makes sense and you get it? :)

blueyed:
This would change the test here to expect `b- ?` then: https://github.com/neomake/neomake/pull/1864/files#diff-a348766d8a02868ac1176f28467796b5R137.

autoload/neomake/statusline.vim
```diff
@@ -168,15 +168,15 @@ function! s:formatter.format(f, args) abort
 endfunction
 
 
-function! s:running_jobs(bufnr) abort
+function! neomake#statusline#running_jobs(bufnr) abort
```
blueyed: @somini 
Do you have a need for this after all?
Seems like a good change by itself then, but I'd rather not expose things on the API if not necessary.

somini: With the new configuration this is not needed.

somini:
I think I understood https://github.com/neomake/neomake/pull/1862#issuecomment-366785492:

- `format_status_enable / disable` are an outer formatting compared to `format_status`

- `format_status_enable` defaults to '%s'

- `format_status_disable` defaults to `{{disabled_info}} %s`
- disabled_info can be controlled with `format_disabled_info` and defaults to `{{disabled_scope}}-`

somini:
This test: https://github.com/somini/neomake/blob/e731e0c488c1c76cddeca9596e7f26a7886f3305/tests/statusline.vader#L141

Shouldn't this be just `?`? Should we care about the scope when automake is enabled? I can do it, though.

blueyed:
> Shouldn't this be just `?`?

`NeomakeEnableBuffer` sets `{'disabled': 0}` explicitly.
Before this it is `?` only.
Yeah, I agree that this (displaying anything if enabled) is not helpful - and previously it was never triggered (although the code was there) IIRC.

autoload/neomake/statusline.vim
```diff
@@ -184,9 +186,13 @@ function! neomake#statusline#get_status(bufnr, options) abort
     if !empty(running_jobs)
         let format_running = get(a:options, 'format_running', '… ({{running_job_names}})')
         if format_running isnot 0
-            return s:formatter.format(format_running, {
-                        \ 'bufnr': a:bufnr,
-                        \ 'running_jobs': running_jobs})
+            let args = {'bufnr': a:bufnr, 'running_jobs': running_jobs}
+            for l:opt in ['running_jobs_separator', 'running_jobs_project_marker']
```
blueyed: No need to use `l:` prefixes.

blueyed: Instead of `running_jobs_project_marker` we could have `format_running_job_project` and `format_running_job_file`.

doc/neomake.txt
```diff
@@ -720,14 +721,32 @@ You can use the following options:
...
     Default: not used / `'%s'`
+  - format_running_job_project: Wrap the running job name for project makers.
+    Default: `'%s!'`.
+  - running_jobs_separator: Separator for formatted running job names.
+    Default: `', '`.
 
 You can use the following format placeholders:
   - `{{running_job_names}}}`: comma-separated list of running jobs (typically
     their maker names)
+  - `{{scope}}`: When manually overriden, the scope of the disable. One of
```
blueyed: s/the disable/the disabling/
But it also is made available for `format_status_enabled`.
I think we should rename it to `disabled_scope` and use it only when disabled for now.
If somebody really wants it for `format_status_enabled` we could add it as `enabled_scope` then.

blueyed: format_disabled_info and format_enabled_info is not documented yet.
Pushed some fixup, which also removes format_enabled_info.
What do you think?

somini: LGTM, I only added the "enabled" version in case anyone wants the old coded behaviour of showing `b+` if "disabled" is not -1. Personally I wouldn't use it.

somini:
Rebased this to the latest master.

I'm using this daily and can't find any outstanding bugs.

blueyed:
Thanks, will review it again / test it myself - might take days/weeks, but don't worry, it is good stuff and will be merged eventually.. :)

blueyed:
Rebased on master, and testing now.
Will merge it soon.

somini:
Thanks. This passes the integration test also know as my personal configuration, so it looks good to me.

Squashed this into a single commit.

blueyed:
Thank you!

somini:
Here's my personal configuration that might serve as "stress test" for this.

https://gitlab.com/somini/dotnvim/blob/9ada3f24cc21f2cfac51a11c1cc0d4277da6d1cf/init.vim#L1421-1447

This will indicate if automake is on or not, the currently running checkers and the results.

PR #1872, opened 2018-02-23
python: flake8: add output_stream, use %:p in args

blueyed:


blueyed:
Need to figure out why the diff is considered to be not covered.
It seems like the hack (https://github.com/blueyed/vader.vim/commit/e66d91dea178d93707a92d42d904a826db8dc346) does not work here.

blueyed:
Fixed / worked around by https://github.com/neomake/neomake/pull/1880.

PR #1890, opened 2018-03-12
Add Swift package manager support

keith:
Previously the only Swift maker was just for compiling a single file
with no arguments, now if there is a Package.swift file in the current
directory, we will use `swift build` instead.

### DIFF
--- a/autoload/neomake/makers/ft/swift.vim
+++ b/autoload/neomake/makers/ft/swift.vim
@@ -1,7 +1,25 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#swift#EnabledMakers() abort
-    return ['swiftc']
+    let package = neomake#utils#FindGlobFile('Package.swift')
+    if !empty(package)
+        return ['swiftpm']
+    else
+        return ['swiftc']
+    endif
+endfunction
+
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/swift.vim
```diff
@@ -1,7 +1,24 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#swift#EnabledMakers() abort
-    return ['swiftc']
+    if filereadable('Package.swift')
+        return ['swiftpm']
+    else
+        return ['swiftc']
+    endif
```
blueyed: Should use `neomake#utils#FindGlobFile` - see usages with other makers.

keith: Updated!

blueyed:
Thank you both!

@keith 
Please add it to https://github.com/neomake/neomake/wiki/Makers.

keith:
Updated! Thanks  @blueyed !

blueyed:
Condensed in https://github.com/neomake/neomake/pull/1891.
You might want to test it, but I'll merge it otherwise once it's green (likely, since there is no real coverage for this).

PR #1891, opened 2018-03-12
swift: condense code

blueyed:


autoload/neomake/makers/ft/swift.vim
```diff
@@ -1,25 +1,19 @@
...
-        \ 'append_file': 0,
-        \ 'errorformat':
-            \ '%E%f:%l:%c: error: %m,' .
-            \ '%W%f:%l:%c: warning: %m,' .
-            \ '%Z%\s%#^~%#,' .
-            \ '%-G%.%#',
-        \ }
+    let maker = neomake#makers#ft#swift#swiftc()
+    call extend(maker, {
+                \ 'exe': 'swiftc',
```
keith: it's `swift`, not `swiftc`

autoload/neomake/makers/ft/swift.vim
```diff
@@ -1,25 +1,19 @@
...
+                \ 'args': ['-parse'],
```
keith: no args

autoload/neomake/makers/ft/swift.vim
```diff
@@ -1,25 +1,19 @@
...
-        return ['swiftc']
     endif
+    return ['swiftc']
 endfunction
 
 function! neomake#makers#ft#swift#swiftpm() abort
-    return {
-        \ 'exe': 'swift',
-        \ 'args': ['build'],
-        \ 'append_file': 0,
```
keith: we cannot pass the file or swift build will fail

keith:
It seems to me like duplicating the errorfmt is actually worth it here since it's not huge, and there's a possibility swiftpm could output errors that aren't the same format as swiftc since it also calls other tools

autoload/neomake/makers/ft/swift.vim
```diff
@@ -1,25 +1,19 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#swift#EnabledMakers() abort
-    let package = neomake#utils#FindGlobFile('Package.swift')
-    if !empty(package)
+    if empty(neomake#utils#FindGlobFile('Package.swift'))
```
keith: you want to reverse this condition?

PR #1915, opened 2018-04-05
Define an Erlang maker aware of header and code path locations

erszcz:
This maker does an educated guess about where `include` and `ebin` directories might be. It tries standard Rebar 3 paths, but also allows to set `g:erlang_extra_deps` to provide a custom file location. ~It uses relative paths, as it assumes that Vim is run in the project's directory and CWD is not changed to the file-being-edited location.~

It heavily follows the 80/20 rule, so the most common false errors in the majority of straightforward Rebar3 (and possibly old Rebar2) projects should be covered, but there are no guarantees. Still, it's way better than running `erlc` with no command line params at all.

It's my first 30 lines of VimScript, so I beg for mercy :)

### DIFF
--- a/autoload/neomake/makers/ft/erlang.vim
+++ b/autoload/neomake/makers/ft/erlang.vim
@@ -1,6 +1,6 @@
 
 function! neomake#makers#ft#erlang#EnabledMakers() abort
-    return ['rebar3_erlc']
+    return ['erlc_glob_paths']
 endfunction
 
 function! neomake#makers#ft#erlang#erlc() abort
@@ -11,24 +11,31 @@ function! neomake#makers#ft#erlang#erlc() abort
         \ }
 endfunction
 
-function! neomake#makers#ft#erlang#rebar3_erlc() abort
+function! neomake#makers#ft#erlang#erlc_glob_paths() abort
     return {
         \ 'exe': 'erlc',
-        \ 'args': function("neomake#makers#ft#erlang#rebar3_glob_paths"),
+        \ 'args': function("neomake#makers#ft#erlang#glob_paths"),
         \ 'errorformat':
             \ '%W%f:%l: Warning: %m,' .
             \ '%E%f:%l: %m'
         \ }
 endfunction
 
-function! neomake#makers#ft#erlang#rebar3_glob_paths() abort
+function! neomake#makers#ft#erlang#glob_paths() abort
     if match(expand('%'), "SUITE.erl$") > -1
         let l:profile = "test"
     else
         let l:profile = "default"
     endif
     let l:ebins = glob("_build/" . l:profile . "/lib/*/ebin", "", 1)
-    let l:args = []
+    " Set g:erlang_extra_deps in a project-local .vimrc, e.g.:
+    "   let g:erlang_extra_deps = ['deps.local']
+    if exists("g:erlang_extra_deps")
+        for extra_deps in g:erlang_extra_deps
+            let l:ebins += glob(extra_deps . "/*/ebin", "", 1)
+        endfor
+    endif
+    let l:args = ['-I', 'include', '-I', 'src']
     for ebin in ebins
         call add(l:args, '-pa')
         call add(l:args, ebin)
--- a/autoload/neomake/makers/ft/erlang.vim
+++ b/autoload/neomake/makers/ft/erlang.vim
@@ -16,7 +16,22 @@ if !exists("g:rebar3_command")
 endif
 
 function! neomake#makers#ft#erlang#rebar3_erlc() abort
-    let l:ebins = systemlist(g:rebar3_command . " path")
+    return {
+        \ 'exe': 'erlc',
+        \ 'args': function("neomake#makers#ft#erlang#rebar3_paths"),
+        \ 'errorformat':
+            \ '%W%f:%l: Warning: %m,' .
+            \ '%E%f:%l: %m'
+        \ }
+endfunction
+
+function! neomake#makers#ft#erlang#rebar3_paths() abort
+    if match(expand('%'), "SUITE.erl$") > -1
+        let l:maybe_profile = "as test"
+    else
+        let l:maybe_profile = ""
+    endif
+    let l:ebins = split(system(g:rebar3_command . " " . l:maybe_profile . " path"), " ")
     let l:args = []
     for ebin in ebins
         call add(l:args, '-pa')
@@ -26,11 +41,5 @@ function! neomake#makers#ft#erlang#rebar3_erlc() abort
     endfor
     call add(l:args, '-o')
     call add(l:args, '_build/neomake')
-    return {
-        \ 'exe': 'erlc',
-        \ 'args': l:args,
-        \ 'errorformat':
-            \ '%W%f:%l: Warning: %m,' .
-            \ '%E%f:%l: %m'
-        \ }
+    return l:args
 endfunction
--- a/autoload/neomake/makers/ft/erlang.vim
+++ b/autoload/neomake/makers/ft/erlang.vim
@@ -17,7 +17,7 @@ function! neomake#makers#ft#erlang#GlobPaths() abort
     let root = fnamemodify(neomake#utils#FindGlobFile('rebar.config'), ':h')
     if empty(root)
         " At least try with CWD
-        root = getcwd()
+        let root = getcwd()
     endif
     let build_dir = root . '/_build'
     let ebins = []
--- a/autoload/neomake/makers/ft/erlang.vim
+++ b/autoload/neomake/makers/ft/erlang.vim
@@ -22,12 +22,15 @@ function! neomake#makers#ft#erlang#erlc_glob_paths() abort
 endfunction
 
 function! neomake#makers#ft#erlang#glob_paths() abort
+    " Pick the rebar3 profile to use.
     if match(expand('%'), 'SUITE.erl$') > -1
         let profile = 'test'
     else
         let profile = 'default'
     endif
-    let ebins = glob('_build/' . profile . '/lib/*/ebin', '', 1)
+    " Find project root directory.
+    let root = fnamemodify(neomake#utils#FindGlobFile('rebar.config'), ':h')
+    let ebins = glob(root . '/_build/' . profile . '/lib/*/ebin', '', 1)
     " Set g:erlang_extra_deps in a project-local .vimrc, e.g.:
     "   let g:erlang_extra_deps = ['deps.local']
     if exists('g:erlang_extra_deps')
--- a/autoload/neomake/makers/ft/erlang.vim
+++ b/autoload/neomake/makers/ft/erlang.vim
@@ -30,14 +30,14 @@ function! neomake#makers#ft#erlang#GlobPaths() abort
         if expand('%') =~# '_SUITE.erl$'
             let profile = 'test'
         endif
-        let ebins += neomake#makers#ft#erlang#Glob(build_dir . '/' . profile . '/lib/*/ebin')
+        let ebins += neomake#compat#glob_list(build_dir . '/' . profile . '/lib/*/ebin')
         let target_dir = build_dir . '/neomake'
     else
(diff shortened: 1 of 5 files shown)


autoload/neomake/makers/ft/erlang.vim
```diff
@@ -10,3 +10,39 @@ function! neomake#makers#ft#erlang#erlc() abort
...
+        \ 'exe': 'erlc',
+        \ 'args': function('neomake#makers#ft#erlang#glob_paths'),
+        \ 'errorformat':
+            \ '%W%f:%l: Warning: %m,' .
+            \ '%E%f:%l: %m'
+        \ }
+endfunction
+
+function! neomake#makers#ft#erlang#glob_paths() abort
+    if match(expand('%'), 'SUITE.erl$') > -1
```
blueyed: Check out `fnamemodify()`, i.e. `fnamemodify('%:t)` to get the basename, if that's what you want.

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -10,3 +10,39 @@ function! neomake#makers#ft#erlang#erlc() abort
...
+    for ebin in ebins
+        call add(l:args, '-pa')
+        call add(l:args, ebin)
+        call add(l:args, '-I')
+        call add(l:args, substitute(ebin, 'ebin$', 'include', ''))
+    endfor
+    call add(l:args, '-o')
+    call add(l:args, '_build/neomake')
+    return l:args
+endfunction
```
blueyed: `l:` prefixes are not necessary / should be removed.

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -10,3 +10,39 @@ function! neomake#makers#ft#erlang#erlc() abort
...
+        for extra_deps in g:erlang_extra_deps
+            let l:ebins += glob(extra_deps . '/*/ebin', '', 1)
+        endfor
+    endif
+    let l:args = ['-pa', 'ebin', '-I', 'include', '-I', 'src']
+    for ebin in ebins
+        call add(l:args, '-pa')
+        call add(l:args, ebin)
+        call add(l:args, '-I')
+        call add(l:args, substitute(ebin, 'ebin$', 'include', ''))
```
blueyed: Can be a single `let args += ['-pa', ebin, …]`.

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -10,3 +10,39 @@ function! neomake#makers#ft#erlang#erlc() abort
...
+        let l:profile = 'default'
+    endif
+    let l:ebins = glob('_build/' . l:profile . '/lib/*/ebin', '', 1)
+    " Set g:erlang_extra_deps in a project-local .vimrc, e.g.:
+    "   let g:erlang_extra_deps = ['deps.local']
+    if exists('g:erlang_extra_deps')
+        for extra_deps in g:erlang_extra_deps
+            let l:ebins += glob(extra_deps . '/*/ebin', '', 1)
+        endfor
+    endif
```
blueyed: Is `g:erlang_extra_deps` something used by other plugins already?

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -10,3 +10,39 @@ function! neomake#makers#ft#erlang#erlc() abort
...
         \ }
 endfunction
+
+function! neomake#makers#ft#erlang#erlc_glob_paths() abort
+    return {
+        \ 'exe': 'erlc',
+        \ 'args': function('neomake#makers#ft#erlang#glob_paths'),
+        \ 'errorformat':
+            \ '%W%f:%l: Warning: %m,' .
+            \ '%E%f:%l: %m'
```
blueyed: Is there a way to get column information btw?

blueyed:
Also check out `neomake#utils#FindGlobFile` for finding the project root (used by other makers already).

blueyed:
To test this manually you can use `:echo neomake#makers#ft#erlang#glob_paths()` btw.

blueyed:
I also think that the default/existing maker should get enhanced with your improvements, instead of creating a new one.

autoload/neomake/makers/ft/erlang.vim
erszcz: Common Test, a testing framework, defines a naming convention for files  containing tests - they should end with `_SUITE.erl`. In this particular case it doesn't matter if we test on the basename or the whole path.

autoload/neomake/makers/ft/erlang.vim
erszcz: AFAIK it's not, I don't know it from my experience and I couldn't google up any suggestion it might be.
I added it because it's a quick way to support Rebar 2 projects or custom project structure in a reasonable way.

erszcz:
Thanks for the review. I'll follow up with further fixes.

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -5,8 +5,45 @@ endfunction
...
+
+function! neomake#makers#ft#erlang#GlobPaths() abort
+    " Pick the rebar3 profile to use.
+    if match(expand('%'), 'SUITE.erl$') > -1
+        let profile = 'test'
+    else
+        let profile = 'default'
+    endif
+    " Find project root directory.
+    let root = fnamemodify(neomake#utils#FindGlobFile('rebar.config'), ':h')
```
blueyed: This should handle the case where `rebar.config` is not found, i.e. `neomake#utils#FindGlobFile('rebar.config')` being empty.
This should be done in the beginning already.

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -5,8 +5,45 @@ endfunction
...
+        for extra_deps in g:erlang_extra_deps
+            let ebins += glob(extra_deps . '/*/ebin', '', 1)
+        endfor
+    endif
+    let args = ['-pa', 'ebin', '-I', 'include', '-I', 'src']
+    for ebin in ebins
+        let args += [ '-pa', ebin,
+                    \ '-I', substitute(ebin, 'ebin$', 'include', '') ]
+    endfor
+    let build_dir = root . '/_build'
```
blueyed: Can be moved up and get used for `let ebins = ` already.

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -5,8 +5,45 @@ endfunction
...
+    " If <project-root>/_build doesn't exist we'll clutter CWD with .beam files,
+    " the same way erlc does by default.
+    if isdirectory(build_dir)
+        let target_dir = build_dir . '/neomake'
+        if !isdirectory(target_dir)
+            call mkdir(target_dir, 'p')
+        endif
+        let args += ['-o', target_dir]
+    endif
```
blueyed: This could fall back to using `tempname()`, e.g.
```vim
if !exists('s:erlc_build_dir')
    let s:erlc_build_dir = tempname()
endif
```

(Neo)Vim will clean up the temporary directories on exit.

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -5,8 +5,45 @@ endfunction
 
 function! neomake#makers#ft#erlang#erlc() abort
     return {
+        \ 'exe': 'erlc',
```
blueyed: `exe` defaults to the name, so is not necessary.

autoload/neomake/makers/ft/erlang.vim
blueyed: I see.  Thought like the file itself needs to be `SUITE.erl`.
You may want to include the `_` then maybe?

Also `if expand('%') =~# '_SUITE.erl$'` could be used.

autoload/neomake/makers/ft/erlang.vim
blueyed: Ok.  Then the variable name should include `neomake_`.
I suggest `neomake_erlang_erlc_extra_deps`.
It should also check a buffer local var.
I.e. `get(b:, 'neomake_erlang_erlc_extra_deps', get(g:, 'neomake_erlang_erlc_extra_deps'))`.

erszcz:
Well, of course my previous comment was nonsense - I did botch up test isolation, therefore tests running after `erlc` couldn't pass.
Finally, all tests are green!

autoload/neomake/makers/ft/erlang.vim
```diff
@@ -5,8 +5,68 @@ endfunction
...
+    if isdirectory(build_dir)
+        " Pick the rebar3 profile to use
+        let profile = 'default'
+        if expand('%') =~# '_SUITE.erl$'
+            let profile = 'test'
+        endif
+        let ebins += neomake#makers#ft#erlang#Glob(build_dir . '/' . profile . '/lib/*/ebin')
+        let target_dir = build_dir . '/neomake'
+    else
+        let target_dir = tempname()
```
blueyed: This will use a new tempdir every time.
Is the target dir helpful when running multiple times, i.e. does it act as a cache?
If not, you could just use `neomake#compat#dev_null` (https://github.com/neomake/neomake/pull/1917).

tests/ft_erlang.vader
```diff
@@ -0,0 +1,95 @@
...
+
+Before (prepare an empty working directory):
+  let s:root = tempname()
+  call mkdir(s:root, 'p')
+  let s:old_cwd = fnamemodify('.', ':p')
+  execute 'cd ' . s:root
+
+After (clean up the working directory):
+  execute 'cd ' . s:old_cwd
+  execute 'silent !rm -rf ' . s:root
```
blueyed: Unfortunately Vader does not support nested Before/After, and there is a global Before/After already that should not be overwritten.
I have to think about how to fix this best.
Apart from that, the `rm -rf` looks a bit scary and is not necessary, since Vim should do the cleanup.

autoload/neomake/makers/ft/erlang.vim
erszcz: It doesn't, `target_dir` is only set not to clutter CWD / the project directory.
I  tried using `/dev/null`, but since it's a file, not a directory, it generates an unnecessary `|| /dev/null/z.bea#: error writing file: not a directory` in the location list.
I think a global / per-buffer target dir would make the most sense.

blueyed: The error when using /dev/null could be filtered out by the errorformat.

tests/ft_erlang.vader
```diff
@@ -0,0 +1,127 @@
...
+  let args = neomake#makers#ft#erlang#GlobPaths()
+  AssertEqual args[:-2], ['-pa', 'ebin', '-I', 'include', '-I', 'src', '-o']
+  let output_dir = args[-1]
+  Assert isdirectory(output_dir)
+  bwipe
+
+Execute (erlc: output to _build/neomake if _build is present):
+  let root = tempname()
+  call mkdir(root, 'p')
+  let s:old_cwd = fnamemodify('.', ':p')
```
blueyed: You can use `getcwd()` to get the current dir.

And then it would be good to wrap it with `try … finally` to `cd` back - so that a failing test restores the cwd.

erszcz:
@blueyed Do you have any other suggestions? Any points that still require fixing?

erszcz:
@blueyed I won't be adding more features to this PR. Do you think it is ready to merge?

blueyed:
Thanks!
Awesome work!

blueyed:
Re options: you could document them in the wiki (https://github.com/neomake/neomake/wiki/Makers).
I'm planning to eventually convert the wiki page into a help file.

erszcz:
I've updated the wiki accordingly. Thanks for the review and all the pointers!

blueyed:
Thanks ([for reference](https://github.com/neomake/neomake/wiki/Makers/_compare/8cd57cda9fc1661449f453c83c5c67441d18c4c2...b1cd0bfc48529795c77b20c13f322f0d0a3a5ef1))!

PR #1927, opened 2018-04-13
Add the xmllint maker

JCavallo:


blueyed:
Thanks!

This appears to come via Syntastic, at least the errorformat is the same (which is good).
Syntastic also uses it for the "xslt" and "docbk" filetypes, and therefore we might want to do the same.

blueyed:
Taking this over (rebased and added fixup).
I'll look into improving the errorformat, since it currently ignores:
```
validity error : no DTD found!
Document /home/user/src/neomake/t/t.xml does not validate
```

blueyed:
Finished for now.

This improves the tests by adding a system to generate real fixtures for tests, which will be helpful in the future.
Would be nice if you could test if it works for you, and if not provide some example .xml files.

I've not yet tested if it works for xslt/docb files as expected.

JCavallo:
@blueyed I just checked on my usual use case, and it works perfectly.

However, I only check basic syntax (tag matching), and support of extra parameters to the linter (use of relaxng to provide a specific validation schema).

JCavallo:
BTW, there are other file formats which as far as I know are actually xml (xsd, fodt, etc...). Should we define xmllint as a maker for those as well ?

blueyed:
> xsd, fodt, etc...
> Should we define xmllint as a maker for those as well ?

Probably, yes.
But that should not hold this PR back I guess.

blueyed:
Merged, thanks!
Updated the wiki: https://github.com/neomake/neomake/wiki/Makers.

PR #1942, opened 2018-04-22
Add new error format for Puppet parser

mandos:
Create test to cover this case and additionally puppet linter


mandos:
I don't know why vim lint failed, I run it locally, and it passed.

blueyed:
The error is:

> Test not included in main.vader: ft_puppet.vader

(from https://circleci.com/gh/neomake/neomake/9532)

You have to include the test file in main.vader.

blueyed:
btw: there is new functionality to automatically generate fixtures from the tools themselves (see 074ce011).
You could use this here, instead of only providing the mocked/copied output.
The benefit is that we would detect changes to output from tools (when re-creating the fixtures).
Let me know if you want to go that route and need help, of course.

mandos:
@blueyed I was blind not seeing it. 
I would like to use this new functionality but a first try was a failure. Could you give me a short info how to use it?

blueyed:
Great.  Pushed a fixup/improvement to the Makefile and prepared it for puppet.

Then another commit that you should amend/remove, which shows how to generate the fixtures: you have to add/change/rename the tests/fixtures/input/puppet/error.pp input file - you should use a more specific name, and probably add one per test case that you have.

Then check tests/ft_xml.vader for how NeomakeTestsFixtureMaker can be used with it.

tests/ft_puppet.vader
```diff
@@ -0,0 +1,22 @@
+Include: include/setup.vader
+
+Execute (puppet parser, syntax error: end of input):
+  new
+  let maker = NeomakeTestsFixtureMaker('neomake#makers#ft#puppet#puppet',
```
blueyed: You should use `file tests/fixtures/input/puppet/syntax-error-eoi.pp` - otherwise errors are not generated for the current buffer, but a new one.

blueyed: Not sure though - maybe `edit/split tests/fixtures/input/puppet/syntax-error-eoi.pp` (possibly prefixed with `noautocmd`) would be better?!

blueyed: It uses a tempfile anyway - would need to check this myself:

> Using tempfile for unnamed buffer: "/tmp/nvimLiCMEo/112/neomaketmp.".

You can use `Log neomake#utils#redir('ls!')` to get a list of buffers to see what's going on.

mandos: > Not sure though - maybe edit/split tests/fixtures/input/puppet/syntax-error-eoi.pp (possibly prefixed with noautocmd) would be better?!

I don't understand it, could you clarify?

blueyed: It only uses `new` to create an unnamed buffer, but the errors are for another file - therefore the assertion in the test fails.  (Vim creates an unlisted buffer when parsing the output)

mandos:
@blueyed  I try to finish this PR but I have strange problem. When I run test locally with
```make tests/ft_puppet.vader``` from root project folder buffers looks ok.
``` 
before asserts:
  1  a   "[No Name]"                    line 1      
  2  a   "[Vader]"                      line 0      
  3 #a   "[Vader-workbench]"            line 0      
  4 %a   "tests/fixtures/input/puppet/syntax-error-eoi.pp" line 1

after bwipe:
  1  a   "[No Name]"                    line 1      
  2 #a   "[Vader]"                      line 0      
  3 %a   "[Vader-workbench]"            line 1
```
But this same test on CircleCI gives me different results:
```
before asserts:
  1  a   "[No Name]"                    line 1
  2  a   "[Vader]"                      line 0
  3 #a   "[Vader-workbench]"            line 1
922 %a   "tests/fixtures/input/puppet/syntax-error-eoi.pp" line 1
923u     "/workspace/vim/neomake/tests/fixtures/input/puppet/syntax-error-eoi.pp" line 0

after bwipe:
  1  a   "[No Name]"                    line 1
  2  a   "[Vader]"                      line 0
  3 %a   "[Vader-workbench]"            line 1
923u     "/workspace/vim/neomake/tests/fixtures/input/puppet/syntax-error-eoi.pp" line 0
```
Any tip how to fix it?

**Comment by blueyed** [comment]
The fixture has `/workspace/vim/neomake/tests/fixtures/input/puppet/syntax-error-eoi.pp` as the filename in the output, which does not match the path on CI - it works only for your path layout.

Fixed this: improved the fixture generation and added a fixup to your WIP commit.

**Comment by blueyed** [comment]
@mandos 
Fixed now.
Thanks for keeping working on it!

**Comment by blueyed** [comment]
> but error format looks ok for me.

There were extra quotes, and the `%t` (type) gets not converted to uppercase.

**Comment by mandos** [comment]
@blueyed I think, the current state is rather complete. 

I had this case I didn't know how to test. Filenames in location list were wrong (double quote before a full path to the file), this leads to not showing marks in a tested file and "Enter" creates a new file instead of jumping to existing one. Do you have any ideas how to test this kind of cases?

**Comment by blueyed** [comment]
> double quote before a full path to the file

Maybe with spaces in the filename, or something else that needs quoting?
Is it with puppet or puppet-lint?

Vim will parse `"file":msg` using `%f:%m` with `"file"` as the file name.
In this case the errorformat would need to be `"%f":%m` - `let &errorformat = '"%\?%f"%\?:%m'` could be used, but would also match `"foo` only.  Grouping cannot be used (to use `\1` as a reference for the closing part), so the existing rule should get duplicated to include the quotes.
Not clear though if it's about the output only, or the actual file name.

**Comment by mandos** [comment]
> Maybe with spaces in the filename, or something else that needs quoting?
> Is it with puppet or puppet-lint?

It's puppet-lint and I fixed it by removing double quotes from command's arguments
```        
\ 'args': ['--log-format', '"%%{path}:%%{line}:%%{column}:%%{kind}:[%%{check}] %%{message}"'],
```
I don't now why it was working before.

Is any problem with codecov?

**Comment by blueyed** [comment]
Codecov might be hanging for some reason, usually a new push might fix it.

Is the output for `tests/fixtures/output/puppet-lint/err_and_warn.pp.stdout` what you get really?  I am seeing more lines:

```
+tests/fixtures/input/puppet-lint/err_and_warn.pp:1:1:warning:[documentation] class not documented
+tests/fixtures/input/puppet-lint/err_and_warn.pp:1:7:error:[autoloader_layout] foo not in autoload module layout
+tests/fixtures/input/puppet-lint/err_and_warn.pp:2:10:warning:[double_quoted_strings] double quoted string containing no variables
```

The quotes are defined in the `--log-format` argument, and should get removed there probably.
So there is a mismatch between the args in the Makefile to generate the fixture (which is not using quotes), and the actual maker.
So the actual errorformat is wrong now, and that's what you are seeing probably when using puppet-lint.
For this it would be good to get the args of the maker for generating the fixtures actually.

I'll push a fixup for now in a couple of minutes.

**Comment by blueyed** [comment]
Oh, I've seen your last commit now.
We appear to get different order for the output of puppet-lint for tests/fixtures/output/puppet-lint/err_and_warn.pp.stdout however.

**Comment by blueyed** [comment]
@stevepeak
Any idea why the codecov report is still hanging?

**Comment by blueyed** [comment]
@mandos 
Please review my last commit, and then I think this is good to go in in general.

**Comment by stevepeak** [comment]
@blueyed looking at Codecov logs: the pull is getting processed but the status is going to `8e72d2e...23da1cb`. 

Was `23da1cb` the old head commit sha?
Why did `423d865` not upload coverage... 😦

**Comment by mandos** [comment]
@blueyed, @stevepeak  Maybe create another PR?

**Comment by blueyed** [comment]
@mandos 
Good idea.  https://github.com/neomake/neomake/pull/1962.

**Comment by blueyed** [comment]
@stevepeak 

> Was 23da1cb the old head commit sha?

It looks like it was some intermediate commit that was rebased/squashed out then:
```
commit 23da1cb7 (puppet-validator-rebased)
Author: Daniel Hahler <…>
Date:   Sun Apr 22 22:44:30 2018 +0200

    WIP: example for puppet error
```

> Why did 423d865 not upload coverage... 

It was uploaded: https://circleci.com/gh/mandos/neomake/131 (https://codecov.io/github/mandos/neomake/commit/423d865ee2d3719458d153c77e32c98c9609ebb9).

PR #1947, opened 2018-04-23
Allow users uses their own phpmd.xml

zhujinxuan:
This PR adds two options allowing users to use their own phpmd.xml for linting.

### DIFF
--- a/autoload/neomake/makers/ft/php.vim
+++ b/autoload/neomake/makers/ft/php.vim
@@ -35,11 +35,11 @@ function! neomake#makers#ft#php#phpcs() abort
 endfunction
 
 function! neomake#makers#ft#php#phpmd() abort
-    if exists('b:neomake_php_phpmd_args_standard') 
+    if exists('b:neomake_php_phpmd_args_standard')
         let l:standard = b:neomake_php_phpmd_args_standard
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/php.vim
```diff
@@ -35,8 +35,16 @@ function! neomake#makers#ft#php#phpcs() abort
...
+    else
+        let l:standard =  'codesize,design,unusedcode,naming'
+    endif
+
     return {
-        \ 'args': ['%:p', 'text', 'codesize,design,unusedcode,naming'],
+        \ 'args': ['%:p', 'text', l:standard],
         \ 'errorformat': '%W%f:%l%\s%\s%#%m'
         \ }
 endfunction
```
blueyed: I do not  like the `l:` prefix (but can amend it).
It would be good to have some example output though to cover those cases.
Also `:let g:neomake_php_phpmd_args = neomake#makers#ft#php#phpmd().args + ['foo']` etc might work already?

zhujinxuan: Thank you.  It is fixable by b:eomake_php_phpmd_args.  I will close this PR.

PR #1950, opened 2018-04-25
exclude vendor directories when running gometalinter

timfeirg:
without this option, gometalinter output will be crazy, so ignoring vendor directories is definitely preferable to most golang users.

https://github.com/alecthomas/gometalinter/issues/344

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -58,7 +58,7 @@ function! neomake#makers#ft#go#gometalinter() abort
     "
     " All linters are only warnings, the go compiler will report errors
     return {
-        \ 'args': ['--disable-all', '--enable=errcheck', '--enable=megacheck'],
+        \ 'args': ['--disable-all', '--enable=errcheck', '--enable=megacheck', '--vendor'],
         \ 'append_file': 0,
         \ 'cwd': '%:h',
         \ 'errorformat':


autoload/neomake/makers/ft/go.vim
```diff
@@ -58,7 +58,7 @@ function! neomake#makers#ft#go#gometalinter() abort
     "
     " All linters are only warnings, the go compiler will report errors
     return {
-        \ 'args': ['--disable-all', '--enable=errcheck', '--enable=megacheck'],
+        \ 'args': ['--disable-all', '--enable=errcheck', '--enable=megacheck', '--exclude=vendor'],
```
blueyed: Hmm..
1. `--exclude` seems to be for messages (nor dirs/files)?

>   -e, --exclude=REGEXP ...     Exclude messages matching these regular expressions.

2. `--vendor` should be used probably?!

> Enable vendoring support (skips 'vendor' directories and sets GO15VENDOREXPERIMENT=1).

blueyed: Not sure if I understood it correctly from the help though?!
It seems to contradict https://github.com/alecthomas/gometalinter/issues/344#issuecomment-326314935 (/cc @dnephin).

dnephin: `--exclude vendor` isn't right. It would run linters on every file/package in `vendor` and only excludes the lint errors.

`--vendor` removes the vendor packages/files from the set of packages that are linted.

If you see errors reported from `vendor/` it usually means that a linter failed to load the go source code, so there is some `go build` error.

timfeirg:
Just learned how to properly use `gometalinter` from https://github.com/alecthomas/gometalinter/issues/467#issuecomment-385162894, so `--vendor` is the correct thing to use here.

This PR has been updated. @blueyed

blueyed:
Yay!
Please consider updating the wiki (https://github.com/neomake/neomake/wiki/Makers) with a note.

PR #1986, opened 2018-06-29
Add SwiftPM errorformat

keith:
This command can produce normal errors with a column, but also those
without, so we need both formats.


keith:
Thanks!


--
Keith Smiley

PR #2010, opened 2018-07-17
add vale linting (process_json)

davidosomething:
Request input particularly on the `let filename = buffer_name(bufnr)` line -- that can't be the best way to go through the JSON object...

### DIFF
--- a/autoload/neomake/makers/ft/markdown.vim
+++ b/autoload/neomake/makers/ft/markdown.vim
@@ -1,13 +1,16 @@
 function! neomake#makers#ft#markdown#SupersetOf() abort
     return 'text'
 endfunction
+
 function! neomake#makers#ft#markdown#EnabledMakers() abort
     let makers = executable('mdl') ? ['mdl'] : ['markdownlint']
+    if executable('vale') | let makers += ['vale'] | endif
     return makers + ['writegood'] + neomake#makers#ft#text#EnabledMakers()
 endfunction
 
 function! neomake#makers#ft#markdown#mdl() abort
     return {
+        \
         \ 'errorformat':
         \   '%W%f:%l: MD%n %m,' .
         \   '%W%f:%l: %m',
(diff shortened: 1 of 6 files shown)


autoload/neomake/makers/ft/markdown.vim
```diff
@@ -1,33 +1,64 @@
...
     return makers + ['writegood'] + neomake#makers#ft#text#EnabledMakers()
 endfunction
 
 function! neomake#makers#ft#markdown#mdl() abort
     return {
-        \ 'errorformat':
-        \   '%W%f:%l: MD%n %m,' .
+        \
+        \ 'errorformat': '%W%f:%l: MD%n %m,' .
         \   '%W%f:%l: %m',
```
blueyed: Why not keep this?

autoload/neomake/makers/ft/markdown.vim
```diff
@@ -1,33 +1,64 @@
...
+        \   .'%-Q,'
+        \   .'%*[ ]%l:%c-%*\d:%n%*[ ]%tarning%*[ ]%m,'
+        \   .'%-G%.%#'
+        \ }
+endfunction
+
+function! neomake#makers#ft#markdown#ProcessVale(context) abort
+    let bufnr = a:context['jobinfo'].bufnr
+    let filename = buffer_name(bufnr)
+    let items = get(a:context['json'], filename, [])
```
blueyed: I think you could just iterate of the filenames and use them as-is?
This way it would be ready to be used as a project maker, too - i.e. it would add entries for all buffers.
Filtering only on the current buffer would be more the responsibility of Neomake, not the maker.

Apart from that `buffer_name` is an obsolete name for `bufnr`.

blueyed:
@davidosomething 
What do you mean?
(just added it for Zsh in #2012)

blueyed:
What's up with the empty `tempfile`?

davidosomething:
Weird. I've removed it and a squashmerge should keep it out of history.
Your JSON fix has fixed the oldVim quoting issue so I'll merge.

PR #2013, opened 2018-07-22
Make SwiftPM build tests

keith:
Previously building only meant building source modules. Now it also
builds tests to find compiler errors.

We might not want to make this the default, since some people might not expect this behavior. Alternatively we could expose a setting that allows users to pass more arguments to `swift build` if they'd like to.

### DIFF
--- a/autoload/neomake/makers/ft/swift.vim
+++ b/autoload/neomake/makers/ft/swift.vim
@@ -12,7 +12,7 @@ endfunction
 function! neomake#makers#ft#swift#swiftpm() abort
     return {
         \ 'exe': 'swift',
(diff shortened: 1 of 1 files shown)


blueyed:
> We might not want to make this the default, since some people might not expect this behavior. 

I expect it to be slower, too?

No idea about it though in general.

keith:
Definitely slower since if you change normal code you'll have to recompile the test code. But right now the downside is if you're iterating on tests you have to manually run the other swiftpmtest check. Not really sure what's ideal either.


--
Keith Smiley

blueyed:
Let's take it for now, thanks!

Issue #1886, opened 2018-03-11
Specify output streams for `rubocop` & `rubocop_rails`


abMatGit:
### Expected behavior

Neomake should not try to handle messages to `stderr` via `rubocop` & `rubocop_rails`. If `rubocop` fails, it is helpful to flag/signal that `rubocop` failed to run. One way to do this is with the default handler:
```
Neomake: [1.3:1:1] rubocop_rails: unexpected output. See :messages for more information.
```

### Steps to reproduce

- Get rubocop (`gem install rubocop`)
- Create a bogus `.rubocop.yml` file with invalid syntax.
- In same directory try to run `:Neomake` on a ruby file (same directory as the bogus `.rubocop.yml` file)
--> No output (because rubocop fails to run silently)

#### Neomake debug information

Async support: 1
Current filetype: ruby
Windows: 0
[shell, shellcmdflag, shellslash]:['/bin/zsh', '-c', 0]

##### Enabled makers

For the current filetype ("ruby", used with :Neomake):
 - mri
   - args: ['-c', '-T1', '-w']
   - auto_enabled: 1
   - errorformat: '%-G%\m%.%#warning: %\%%(possibly %\)%\?useless use of == in void context,%-G%\%.%\%.%\%.%.%#,%-GSyntax OK,%E%f:%l: syntax error\, %m,%Z%p^,%W%f:%l: warning: %m,%Z%p^,%W%f:%l: %m,%-C%.%#'
   - exe: 'ruby'
 - rubocop
   - args: ['--format', 'emacs', '--force-exclusion', '--display-cop-names']
   - auto_enabled: 1
   - errorformat: '%f:%l:%c: %t: %m,%E%f:%l: %m'
   - exe: 'rubocop'
   - postprocess: function('neomake#makers#ft#ruby#RubocopEntryProcess')
 - rubocop_rails
   - args: ['--format', 'emacs', '--force-exclusion', '--display-cop-names', '--rails']
   - auto_enabled: 1
   - errorformat: '%f:%l:%c: %t: %m,%E%f:%l: %m'
   - exe: 'rubocop'
   - postprocess: function('neomake#makers#ft#ruby#RubocopEntryProcess')
 - reek
   - args: ['--format', 'text', '--single-line']
   - auto_enabled: 1
   - errorformat: '%W%f:%l: %m'
   - exe: 'reek'
   - WARNINGS:
     - maker's exe (reek) is not executable.
 - rubylint
[...]

PR #2030, opened 2018-08-03
Adding a cargotest maker for rust

okhin:
It seems that the errorformat provided by rust-vim does not works for cargo test.

The current cargo maker use JSON output, but test results and errors aren't send there. This errorformat should work, but it might need thorough testing.


blueyed:
I can take care of fixing existing tests.  The failure is only due to having a new maker, which is not really cool after all.. ;)

```
    ( 77/537) [EXECUTE] neomake#cmd#complete_makers: completes all (executable) makers, preferring filetype
    ( 77/537) [EXECUTE] (X) Unequal Lists
      ['cargo', 'rustc', 'cabal', 'clippy', 'gradle'] should be equal to
      ['cargo', 'cargotest', 'rustc', 'cabal', 'clippy', 'gradle']
      > /home/neomake/repo/tests/completion.vader:44: AssertEqual neomake#cmd#complete_makers('', 'Neomake! '), [ 'cargo', 'rustc', 'cabal', 'clippy', 'gradle']
```

okhin:
Ok, Vader tests are working, but I don't really understand how I can improve the coverage, since there's a test for it.
I'm obviously missing something.

okhin:
Victory \o/, ready to merge :)

autoload/neomake/makers/ft/rust.vim
```diff
@@ -21,6 +21,33 @@ function! neomake#makers#ft#rust#rustc() abort
         \ }
 endfunction
 
+function! neomake#makers#ft#rust#cargotest() abort
+    let maker = {
+        \ 'exe': 'cargo',
+        \ 'args': ['test', '--quiet'],
+        \ 'auto_enabled': 1,
```
blueyed: `auto_enabled` does not belong here.
Why have you added it?
It gets internally to see if a maker was configured/called explicitly, or via `GetEnabledMakers`.

autoload/neomake/makers/ft/rust.vim
```diff
@@ -21,6 +21,33 @@ function! neomake#makers#ft#rust#rustc() abort
...
+        \ 'cwd': '%:p:h',
```
blueyed: It this necessary really?
If so, why?
(it gets overwritten below for when a cargo.toml is found anyway - so this is just a bit confusing at least - it should be set in one place)

tests/ft_rust.vader
```diff
@@ -53,3 +53,32 @@ Execute (rust: cargo: uses directory with Cargo.toml as cwd):
...
+Execute (cargo test failures):
+  new
+  file build/lib.rs
+  let cargotest = NeomakeTestsGetMakerWithOutput('neomake#makers#ft#rust#cargotest',
+    \ '../test/fixtures/rust/cargotest_with_failures')
+
+  call neomake#Make(1, [cargotest])
+  NeomakeTestsWaitForFinishedJobs
+
+  AssertEqual 1, len(getloclist(0))
```
blueyed: Please use something like `AssertEqual map(getloclist(0), 'v:val.text'), ['error message']` to assert some property at least.

autoload/neomake/makers/ft/rust.vim
okhin: I think I took it from what NeomakeCheck was displaying (something like what's displayed at #1886. But ok, I'll remove that :)

autoload/neomake/makers/ft/rust.vim
okhin: You're right, not necessary. I think it's an artifact for when I tried to test just one file (Neomakemake vs NeomakeProject). But I figured out later that cargo test with a file will do just that.

tests/ft_rust.vader
okhin: Ok, will read a bit more docs for that.

blueyed:
Looks good!
I assume it is ready from your side?

Can you please provide the steps to produce `tests/fixtures/rust/cargotest_with_failures`?
I would like to add this to the automatic fixture generation then (via `make rebuild-fixtures` currently, but I have started to move/improve this, and would then like to include this fixture there, too).

okhin:
Mm, lemme check if I can automate the test creation. But basically it's cargo test --quiet > file in a rust project.

A small bash script like that does the trick (the test will fail, because 2+2 is not equal to 5

```bash
#!/bin/bash
cargo new --lib fixtures
cd fixtures
sed -e s/4/5/ src/lib.rs
cargo test --quiet > tests/fixtures/rust/cargotest_with_failures
cd ..
rm -rf fixtures
```

Will need to change the error message in the test though. I can do it on my side, and/or add the scripts (tell me where I should put it)

blueyed:
Does not work out of the box for me:
```sh
% cargo new --lib fixtures
error: no default toolchain configured
```
I get this error when running `cargo` in general, i.e. also with just `--help`.

(I have no experience with cargo in case you have not noticed yet.. ;))

`cargo` is from rustup 1.13.0-3 (Arch Linux) here.

Yes, please adjust the fixture in the PR output then.
I will use your script then when improving the automated fixture generation.

okhin:
So it happens that the errorformat works only for the compilers error. Test failures needs another format. Working on it.

okhin:
Mmm, you need a cargo binary install, which requires the rust-roolchain. I think that can be automated (rustup.sh can be curl'ed, and it should works, need to test that in a machine without rust :)).

And, cargo test works either globally OR with a test name, but it does not like having a file name passed as an args. So, I'll probably just make a Project maker if it's ok (and in that case, I should put it straight in the makers/ dir ?

blueyed:
> no default toolchain configured

Could fix that using `rustup default stable`.
It does not need to be that automated (e.g. not necessary to run it on Travis, but only in a/my local setup).

I think it is better to use `append_file=0` for now, and leave it in makers/ft.

But btw: have you considered using https://github.com/janko-m/vim-test instead/additionally?
vim-test uses the cargo compiler that is shipped with Vim then for the errorformat etc: https://github.com/vim/vim/blob/master/runtime/compiler/cargo.vim

okhin:
@blueyed I don't know if making the fixtures can imply using a docker image (There's one maintained by [rust-lang](https://github.com/rust-lang-nursery/docker-rust/).
Else, there's the [rustup.sh](https://rustup.rs/) to be downloaded and executed. But it might be quite big to download all the needed dependencies.

Anyway, a more precise errorformat and an extended fixtures which have compiler error and test failures (as well as a warning and an info)

blueyed:
1. What source file have you used for generating the fixture now?
2. What do you think about vim-test (see my previous comment)
3. I will look into fixing tests/completion.vader, so that no changes are required there maybe, because it is not good that those tests need to be changed whenever a new (matching) maker gets added.

okhin:
1. When creating a new project, cargo creates some empty source files. I'm sedding them (sorry, just woke up, I'm grabbing a coffee and will write a small bash script to generate that).
2. It works fine, but it needs a project maker with the neomake strategy. And since there was no such maker for cargo (one which will run test), it used the default neomake project (makeprg + errorformat). And vim.test allow to set makeprg to what I need (cargo test --quiet), but does not allow to override errorformat.
That's why I started working on cargo test maker, to be able to use QF with tests.

blueyed:
> I'm sedding them (sorry, just woke up, I'm grabbing a coffee and will write a small bash script to generate that).

No hurry.  I think the resulting file would be enough for now probably.
I would not want to automate the project setup, but only the `cargo test` run on them.

okhin:
So, script. I'm using cargo, but it might be easier to just cat<<<EOF >fixtures.rs the content of the fixtures (it's only 6 lines).

```
#!/bin/bash

# First, let's setup the default toolchain to stable
rustup default stable

# Upgrade the toolchain
rustup update

# In case a previous make fixtures was there, rm it
rm -rf fixtures

# We're creating a new library projct with cargo
# it will create a Cargo.toml and a src/lib.rs file
# which we're going to alter somehow.
cargo new fixtures --lib

# First test is an equality. To fail it, change the
# expected result.
sed -ie 's/4/5/g' fixtures/src/lib.rs

# We want to insert a compiler error (missing crate is
# easy enough for that)

COMP_ERROR="#[test]\nfn_cannot_load() {\nextern crate non::existent::crate;\nassert_eq!(2 + 2, 4);\n}\n\n#[test]"

# Let's add that to our library
sed -ie "s/\#\[test\]/$COMP_ERROR/" fixtures/src/lib.rs
```

dpc:
@blueyed No comments, but I'm happy to test.

I'm puzzled what's the relationship between what I'm currently doing:

```
let g:neomake_rust_cargo_command = ['test', '--no-run']
call neomake#configure#automake('w')
```

and this.

Does it mean I can now just:

```
autocmd! BufWritePost *.rs Neomake cargotest
```

?

If it looks OK, please land and I'll be happy to work with this for a while a report any issues.

blueyed:
Yes, looks like it.
Cool.
I will wait for the cwd (cargo metadata) fixes, and then rebase it to make use of it here, too (https://github.com/neomake/neomake/pull/2030/files#diff-c1af06e46298cc214473d5e0d7f8b8d4R49).

blueyed:
One main difference here is that it does not use the JSON output, but only the old-school errorformatting.  Might be more performant due to this, and was previously also not affected by not `cd`ing for `process_output` therefore.

blueyed:
Rebased and adjusted.  Previous HEAD: 2591039b

@dpc 
You might want to try this branch locally already, but otherwise I will merge it soon anyway.

dpc:
@blueyed I've just did and the first difference I've noticed is that the whole  cargo output is displayed:

```
  1 src/tests.rs|18 col 5 error 425| cannot find value `f` in this scope
  2 || |
  3 ||     f;
  4 ||     ^ not found in this scope
  5 src/tests.rs|20 col 5 error 425| cannot find function `g` in this scope
  6 || |
  7 ||     g();
  8 ||     ^ not found in this scope
  9 || linking with `cc` failed: exit code: 1
 10 || |
```

Instead of just:

```
  1 src/tests.rs|18 col 5 error 425| cannot find value `f` in this scope
  5 src/tests.rs|20 col 5 error 425| cannot find function `g` in this scope
```

or something like that. I wouldn't mind much, but `:lne` doesn't jump to the next actual error, but to the next line, which is making `:lne` not useful at all.

blueyed:
Yeah, that should be fixed.. likely by something like `%-G%.%#` in the errorformat at the end.
For now you could also `g:neomake_rust_cargotest_remove_invalid_entries = 1` probably.

okhin:
Just saw the activity (sorry, I'm recovering from a burnout so I tend to keep a low-profile on contributions for now :/).

I just wanted to add that the reason I'm not using json formatting is because, at the time I wrote the plugin, cargo didn't echoed everything needed in json (cargo build process ar json'ed, cargo test aren't), that's why I did it old school. And it's not done yet (see https://github.com/rust-lang/cargo/issues/1403 for reference).

Was wondering what was needed to merge this, if I can spend some time on it, I'll gladly do it :)

blueyed:
@okhin 
Sorry to hear that, get well soon!  (not sure if VimScript is the best for recovering though ;))
I think only the issue reported by @dpc needs addressing for now, but maybe more comes up due to it being used.
Having the raw output for this would be good, so we could have a test for it then, too - or adjust an existing one.

okhin:
@blueyed as for raw output what's currently in tests/fixtures/rust/cargotest_with_failures isn't enough ? What's missing ?

okhin:
So, I'm kinda rusted and forgot about vader (incoming push)

dpc:
```
Error detected while processing function neomake#Make[21]..<SNR>110_Make[54]..neomake#map_makers[6]..neomake#GetMaker[9]..neomak│···
e#get_maker_by_name[13]..<SNR>110_GetMakerForFiletype[5]..neomake#makers#ft#rust#cargotest:                                     │···
line    1:                                                                                                                      │···
E722: Missing comma in Dictionary: '%-G%\ %#error: aborting due to %\\d%#%\ %#previous errors,' . '%E%\ %#error[E%n]:\ %m,' . '%│···
E%\ %#error:\ %m,' . '%I%\ %#note:\ %m,'. '%W%\ %#warning:\ %m,' . '%-Z%\ %#-->\ %f:%l:%c,' . '%-G%\\d%# %#|\ %s,' . '%-G%\\d%# │···
%#|,' '%-G\ %#\= %*[^:]:\ %m,'. '%E%\ %#%m,' . '%G%\ %#%s%\\,,' . '%Z%\ %#%s%\\,%\\s%f:%l:%c' }                                 │···
E15: Invalid expression: { 'exe': 'cargo', 'args': ['test', '%:t:r', '--quiet'], 'append_file': 0, 'errorformat': '%-G,' . '%-Gt│···
est %s,' . '%-Grunning %\\d%# test%s,' . '%-Gfailures:%s,' . '%-G----%s,' . '%-G%.%#--verbose%s,' . '%-G%.%#--explain%s,' . '%-G│···
error: aborting due to previous error,' '%-G%\ %#error: aborting due to %\\d%#%\ %#previous errors,' . '%E%\ %#error[E%n]:\ %m,'│···
 . '%E%\ %#error:\ %m,' . '%I%\ %#note:\ %m,'. '%W%\ %#warning:\ %m,' . '%-Z%\ %#-->\ %f:%l:%c,' . '%-G%\\d%# %#|\ %s,' . '%-G%\│···
\d%# %#|,' '%-G\ %#\= %*[^:]:\ %m,'. '%E%\ %#%m,' . '%G%\ %#%s%\\,,' . '%Z%\ %#%s%\\,%\\s%f:%l:%c' }                            │···
Error detected while processing function neomake#Make[21]..<SNR>110_Make:                                                       │···
line   54:                                                                                                                      │···
E171: Missing :endif  
```

okhin:
@blueyed I haven't forced-push :/ I'll reset it and apply the patch.
@dpc Oh shit.... I hate myself, I've reread all the expression, but missed a comma :man_facepalming: 

Taking a :coffee:  and fixing up things.

okhin:
(finally)

dpc:
Simple test, seems to work OK.

BTW. When I use `test` mode I get the same error twice (because it was displayed during both normal `build` and `cargo test`?). It's not a big deal, but maybe there is a way to not display duplicates?

blueyed:
@dpc 
> I get the same error twice (because it was displayed during both normal cargo build and cargo test?). It's not a big deal, but maybe there is a way to not display duplicates?

I will add a generic filter that removes duplicates, and it could be applied here then.
Do I undertand it correctly that `cargo test` runs `cargo build` internally before, and then the same error gets displayed twice?  And that it would not happen when you run it again just afterwards where `build` is not triggered?
Can you provide the raw output for a case like this when running the command that Neomake uses from a terminal?  (ensuring that the build phase gets triggered, i.e. there is a duplicate)

dpc:
I need to investigate this more, because something is terribly odd in my setup.

dpc:
https://github.com/rust-lang/rust/issues/54138

blueyed:
Appears to be a bug in cargo (https://github.com/rust-lang/cargo/issues/5128).
I would say to revert/take out the removal of duplicates, because they are not expected to be there in the first place?

blueyed:
@okhin 
JFI: I see that you have not force-pushed now, but should have reset your local branch after I did so.  Anyway, nothing too bad, and I wanted to squash it anyway.

dpc:
> I would say to revert/take out the removal of duplicates, because they are not expected to be there in the first place?

Yeah, I don't think it's responsibility of neomake to take care of it.

Though if there is any workaround etc. I would be happy to use it. :D

blueyed:
Merged finally.
I've kept the removal of duplicates.
Thank you!

PR #2031, opened 2018-08-03
Add support for always using stdin

blueyed:
`uses_stdin` can be set on the maker or via settings unconditionally.


blueyed:
This allows to fix the hadolint maker: https://github.com/neomake/neomake/pull/1995#issuecomment-410231366.

PR #2048, opened 2018-08-07
Open cwindow botright

micke:
I'm coming from using dispatch and vim-rspec where the quickfix window was always opened using botright and i find it a bit weird how neomake opens lwindow and cwindow.

This PR serves more as a question to you @blueyed. I don't actually know what `cwindow` defaults to? From my testing it seem to open under the rightmost split?

You might be more open to putting this behind a config flag instead of hardcoding it to botright?

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -976,7 +976,7 @@ function! s:HandleLoclistQflistDisplay(jobinfo, loc_or_qflist, ...) abort
         let cmd = 'lwindow'
     else
         call neomake#log#debug('Handling quickfix list: executing cwindow.', a:jobinfo)
-        let cmd = 'cwindow'
+        let cmd = 'botright cwindow'
     endif
     if open_val == 2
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks.

I think it makes more sense for quickfix windows (global) than for location list windows (related to the window), but I assume that you also like it better with location list windows?

Just for reference:
vim-qf could handle this: https://github.com/romainl/vim-qf/blob/1b0d6e2dd6dcf6ac778c473533577a4384bdc25a/doc/qf.txt#L233-L252
It basically does `wincmd J`, but I am not using it due to https://github.com/vim/vim/issues/931.

lisinge:
I will refactor the PR to only affect quickfix, that's a good spot! I don't think I've actually been able to trigger loclist yet.

vim-qf is a good shout, but i don't think i will be able to use it because of the issue you linked.

blueyed:
Just noticed that you have updated it already, merged.
Thanks!

Issue #449, opened 2016-06-01
No way to use `python3` instead of `python` to check code.


SquishyStrawberry:
When running Neomake and writing Python 3 code, sometimes cod eis marked as incorrect when using Python 3 syntax instead of Python 2 syntax.

`python` points to Python2.7, and `python3` to Python3.5 in my system.

I will provide more information if needed.

blueyed:
The workaround (and best practice in general) is to use a virtualenv / pyenv, so that `python` points at the Python version you're using.

blueyed:
Otherwise, you have to configure the `_exe` setting(s) for the relevant makers.

SquishyStrawberry:
How would I configure the `_exe` setting? Just as any other settings such as `args`?

blueyed:
@SquishyStrawberry 
The `_exe` setting can be a callback/function, and there you can look at `getline(1)`.

SquishyStrawberry:
@blueyed Thanks, you can close this now.

dbrodie:
@SquishyStrawberry Could you please share the configuration that worked for you for future google searches?

blueyed:
@dbrodie 
what problem do you have?
Also related to having a shebang that is different from `python`?

I think that you should work in an environment, where `python` is the Python you want to use.

Check out `:h neomake-options`.

But I can see that this could be improved to look at the shebang indeed - we're doing something in this regard for the `sh` maker already.

dbrodie:
I am editing a project which contains some python3 and python2 code, and sadly would rather not have a separate instance of vim for each.

I'll play around with configuring it, but since this is the first result I found when searching in google, I think it would be better to [have the result for people who happen on this issue.](https://xkcd.com/979/)

blueyed:
@dbrodie 
So are you using shebang lines then?!

dbrodie:
Sorry for not being clearer, yes, I'm using them.

kreedz:
I have exactly the same problem as described in the OP post.
And I'm not using any of pyenv or virtualenv.
I've installed pep8 for both python 2 and 3.
```
let g:neomake_python_pep8_exe = 'python3'
let g:neomake_python_enabled_makers = ['pep8']
```
This doesn't help, doesn't give any error, and not making any checking.
Next snippet only checking python2 rules.
`let g:neomake_python_enabled_makers = ['pep8']`

blueyed:
@kreedz 
[...]

PR #2062, opened 2018-08-20
python: get exe/args from shebang

blueyed:
Ref: https://github.com/neomake/neomake/issues/449#issuecomment-414305622

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -343,7 +343,7 @@ function! neomake#makers#ft#python#pylama() abort
 endfunction
 
 function! neomake#makers#ft#python#python() abort
-    return {
+    let maker = {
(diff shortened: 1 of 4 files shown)


euclio:
What's the status of this PR?

blueyed:
@euclio 
I do not really like it to be used by default I guess (because it adds some overhead).
But we could have this as a helper and document how to configure it to be used.

blueyed:
@euclio 
Made some improvements.

Would be good if you could try it and provide feedback.

It can be used via `call neomake#config#set('b:python.InitForJob', function('neomake#utils#set_argv_from_shebang'))` to enable it for the current buffer's python maker.

euclio:
Thanks for this work, but I'm now using `InitForJob` to use `pipenv` for this functionality.

PR #2063, opened 2018-08-20
add correct maker.version_args for javac and checkstyle

davidosomething:


blueyed:
Thanks!

I wonder if we should try at least `-v` always, if `--version` fails?
Then `-version` could also be tried maybe?

davidosomething:
@blueyed  i think it should only run if `version_arg` is provided.

e.g. `bashate` (https://docs.openstack.org/bashate/latest/readme.html) is a linter that uses `-v` for "verbose output" and doesn't have a version flag at all.  
I haven't added it as an upstream maker since there's no way to skip running the linter exe for version retrieval

blueyed:
> i think it should only run if version_arg is provided.

But defaulting to `--version` is OK still, isn't it?

blueyed:
Feel free to merge this as-is then.

davidosomething:
How should I merge without the coverage test pass? Manually merge to master and push?

PR #2065, opened 2018-08-22
[php.phpcs] disable phpcs progress

sserbin:
particularly when no errors found, phpcs would output it's progress unnecessary polluting location list


blueyed:
Thanks!

PR #2066, opened 2018-08-22
[php.psalm] added psalm maker

sserbin:
Basic support for psalm static checker https://github.com/vimeo/psalm


sserbin:
Thanks, added

PR #2070, opened 2018-08-30
[wip] nix: add maker that runs buildPhase

teto:
... or eval "$buildPhase".

The user is supposed to have run the previous steps
configurePhase/patchPhase beforehand.

The maker checks for a 'build' folder to adjust the current folder so that one
can click the quickfix errors and jump to the error.

I could not test it as I am not sure how to run the maker though :s `Neomake! nix` seems useless.

### DIFF
--- a/autoload/neomake/makers/nix-shell-maker.sh
+++ b/autoload/neomake/makers/nix-shell-maker.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+if [ -z $IN_NIX_SHELL ];
+then
+	echo "you are not in a nix-shell"
+	exit 1
+fi
+
+
+echo $stdenv
+source $stdenv/setup
+
+# doesn't seem effective in the context of quickfix
+# if there is a subfolder called "build", most likely we should run it from there
+# if [ -d "build" ]; then
+# 	echo "there is a build/ folder"
+# 	cd build
+# fi
+
+# to get gcc messages in English
+export LANG=C
+
+if [ -z "$buildPhase" ]; then
+	echo "build from function"
+	echo "CWD: $PWD"
+	buildPhase
+else
+	$buildPhase
+fi
--- a/autoload/neomake/makers/nix.vim
+++ b/autoload/neomake/makers/nix.vim
@@ -1,0 +1,26 @@
+" vim: ts=4 sw=4 et
+
+function! Check_build_folder(opts, ) abort dict
+
+  " todo check for nix-shell
+  if isdirectory("build")
+    let self.cwd = getcwd().'/build'
+  endif
+
+  if !exists("$IN_NIX_SHELL")
+    echom "You are not in a nix-shell" 
+  endif
+
+  return self
+endfunction
+
+function! neomake#makers#nix#nix() abort
+    \ 'exe': expand("%:p:h").'/nix-shell-maker.sh',
+    \ 'args': [],
+    \ 'errorformat': '%f:%l:%c: %m',
+    \ 'remove_invalid_entries': 0,
+    \ 'buffer_output': 0,
+    \ 'InitForJob': function('Check_build_folder')
+    \ }
+endfunc
+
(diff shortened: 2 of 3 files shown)


autoload/neomake/makers/nix-shell-maker.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
blueyed: This should go into a subdirectory, e.g. `autoload/neomake/makers/scripts`.

autoload/neomake/makers/nix-shell-maker.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ -z $IN_NIX_SHELL ];
+then
+	echo "you are not in a nix-shell"
```
blueyed: to stderr?  `>&2`

autoload/neomake/makers/nix.vim
```diff
@@ -0,0 +1,26 @@
...
+function! neomake#makers#nix#nix() abort
+    \ 'exe': expand("%:p:h").'/nix-shell-maker.sh',
+    \ 'args': [],
+    \ 'errorformat': '%f:%l:%c: %m',
+    \ 'remove_invalid_entries': 0,
+    \ 'buffer_output': 0,
+    \ 'InitForJob': function('Check_build_folder')
+    \ }
+endfunc
+
```
blueyed: Trailing newline.

teto:
> Neomake! nix seems useless.

Sry that was badly worded: I meant I don't know how to run the maker neomake#makers#nix#nix().
(previous command would return Neomake: Maker not found (for filetype c): nix)

blueyed:
Oh.. sorry, that is broken currently  See https://github.com/neomake/neomake/pull/2072.

But you also need to fix the function.  Here is some diff on top:

```diff
diff --git a/autoload/neomake/makers/nix.vim b/autoload/neomake/makers/nix.vim
index 81249a95..0dd123d3 100644
--- a/autoload/neomake/makers/nix.vim
+++ b/autoload/neomake/makers/nix.vim
@@ -1,25 +1,25 @@
-" vim: ts=4 sw=4 et
-
 function! Check_build_folder(opts, ) abort dict
-
   " todo check for nix-shell
-  if isdirectory("build")
+  if isdirectory('build')
     let self.cwd = getcwd().'/build'
   endif
 
-  if !exists("$IN_NIX_SHELL")
-    echom "You are not in a nix-shell" 
+  if !exists('$IN_NIX_SHELL')
+    echom 'You are not in a nix-shell.'
   endif
 
   return self
 endfunction
 
 function! neomake#makers#nix#nix() abort
-    \ 'exe': expand("%:p:h").'/nix-shell-maker.sh',
+    return {
+    \ 'exe': expand('%:p:h').'/nix-shell-maker.sh',
     \ 'args': [],
     \ 'errorformat': '%f:%l:%c: %m',
     \ 'remove_invalid_entries': 0,
     \ 'buffer_output': 0,
     \ 'InitForJob': function('Check_build_folder')
     \ }
-endfunc
+endfunction
+
+" vim: ts=4 sw=4 et
```

teto:
thanks for the fix ! the maker now runs and it runs in the kernel case but it fails for other projects and I don't know why. I will ask some nix folks now that the code is public.

teto:
I would like to get notifications when the maker completes. According to the doc 
```
The `exit_callback` (which should get replaced by using the
|NeomakeJobFinished| or |NeomakeFinished| hooks) gets the following dictionary
```
one should rely on autocommands but their use was discouraged by Braam and I would like the callback to be only for this maker. If I use autocommands then I have to filter for the maker name. Why was "exit_callback" -  InitForJob's mirror - deprecated ?

blueyed:
> one should rely on autocommands but their use was discouraged by Braam

Can you elaborate?

I think we could keep exit_callback (or some other method on the maker), but one drawback with the current approach appears to be at least that it is called on the job's exit (and therefore not for get_list_entries makers probably), and also not _after_ the job is processed.

I think a new method similar to `InitForJob` could be used, and then invoked where `NeomakeJobFinished` gets invoked.

PR #2099, opened 2018-09-29
Fix args/append_file for various makers

blueyed:
This fixes the filename being passed twice to clangtidy etc, and behaves
as documented now.

### DIFF
--- a/autoload/neomake/makers/ft/javascript.vim
+++ b/autoload/neomake/makers/ft/javascript.vim
@@ -74,7 +74,7 @@ endfunction
 function! neomake#makers#ft#javascript#rjsx() abort
     return {
         \ 'exe': 'emacs',
-        \ 'args': ['%','--quick','--batch','--eval='
+        \ 'args': ['%t','--quick','--batch','--eval='
         \ .'(progn(setq package-load-list ''((js2-mode t)(rjsx-mode t)))(package-initialize)(require ''rjsx-mode)'
         \ .'  (setq js2-include-node-externs t js2-include-rhino-externs t js2-include-browser-externs t js2-strict-missing-semi-warning nil)'
         \ .'  (rjsx-mode)(js2-reparse)(js2-display-error-list)'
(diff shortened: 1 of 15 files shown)


autoload/neomake/makers/ft/javascript.vim
```diff
@@ -74,7 +74,7 @@ endfunction
 function! neomake#makers#ft#javascript#rjsx() abort
     return {
         \ 'exe': 'emacs',
-        \ 'args': ['%','--quick','--batch','--eval='
+        \ 'args': ['%t','--quick','--batch','--eval='
```
blueyed: @bounceme
I've assumed that the file position is relevant here (and kept the `%t`), but maybe it is not? (i.e. it can be appended?)

autoload/neomake/makers/ft/nix.vim
```diff
@@ -7,7 +7,7 @@ endfunction
 function! neomake#makers#ft#nix#nix_instantiate() abort
     return {
         \ 'exe': 'nix-instantiate',
-        \ 'args': ['%:p', '--parse-only'],
+        \ 'args': ['--parse-only'],
```
blueyed: @gilligan 
From the man page / online documentation it looked like the position of the file arg is not relevant, and therefore I've removed it - append_file will append it (and did so already before this change, i.e. the filename was there twice).

autoload/neomake/makers/ft/php.vim
```diff
@@ -37,7 +37,8 @@ endfunction
 
 function! neomake#makers#ft#php#phpmd() abort
     return {
-        \ 'args': ['%:p', 'text', 'codesize,design,unusedcode,naming'],
+        \ 'args': ['%t', 'text', 'codesize,design,unusedcode,naming'],
+        \ 'append_file': 0,
```
blueyed: `phpmd` appears to require the file name to be in the position, correct @winged?

PR #2112, opened 2018-10-19
php: change phpstan's --errorFormat to --error-format

dreamlover:
The latest version of phpstan has marked the --errorFormat parameter as deprecated and recommend --error-format instead.


blueyed:
I assume `--error-format` is supported widely then already (i.e. has not just been added when deprecating `--errorFormat`?

blueyed:
Thanks!

dreamlover:
@blueyed They can both be used now, so I thought ```--error-format ``` was added sometime ago! :)

dreamlover:
`--error-format` was supported starting from phpstan 0.10.3, as specified [here](https://github.com/phpstan/phpstan/releases/tag/0.10.3).

PR #2113, opened 2018-10-20
[proto.prototool] Add proto maker

martinbaillie:
This adds a Protocol Buffers (*.proto) maker and introduces Uber's prototool as the default maker.

### DIFF
--- a/autoload/neomake/makers/ft/proto.vim
+++ b/autoload/neomake/makers/ft/proto.vim
@@ -1,0 +1,11 @@
+function! neomake#makers#ft#proto#EnabledMakers() abort
+    return ['prototool']
+endfunction
+
+function! neomake#makers#ft#proto#prototool() abort
+    return {
+                \ 'exe': 'prototool',
+                \ 'args': ['lint'],
+                \ 'errorformat': '%f:%l:%c:%m',
+                \ }
+endfunction


autoload/neomake/makers/ft/proto.vim
```diff
@@ -0,0 +1,10 @@
+function! neomake#makers#ft#proto#EnabledMakers() abort
+    return ['prototool']
+endfunction
+
+function! neomake#makers#ft#proto#prototool() abort
+    return {
+                \ 'exe': 'prototool',
+                \ 'args': ['lint'],
+                \ }
```
blueyed: I think there should be an errorformat, no?
Can you provide example output?

martinbaillie: I felt the output was useful by default:

```bash
λ prototool lint
example.proto:7:17:Expected field number.
```
screen shot 2018-10-24 at 7 14 49 pm

```bash
λ prototool lint
example.proto:1:1:File option "go_package" is required.
example.proto:1:1:File option "java_multiple_files" is required.
example.proto:1:1:File option "java_outer_classname" is required.
example.proto:1:1:File option "java_package" is required.
example.proto:1:1:No package declaration found.
```
screen shot 2018-10-24 at 7 18 05 pm

martinbaillie: Makes sense! I've updated the PR to include the error format.

Yes I'm using `g:neomake_highlight_lines = 1`. A recent switch for me and I'm still undecided on whether I'll keep it on :)

blueyed:
Thanks, please add it to the wiki: https://github.com/neomake/neomake/wiki/Makers.

autoload/neomake/makers/ft/proto.vim
blueyed: Re `g:neomake_highlight_lines = 1`: I would assume that it is more useful to see the exact column if provided.
Have not checked, but maybe it makes sense to have a mode where columns are always used (if given / not 1), and highlight the line only otherwise?

martinbaillie: Yes I think a mode like that could be handy. Best of both worlds.

martinbaillie:
Done. Thanks!

autoload/neomake/makers/ft/proto.vim
blueyed: Cool, https://github.com/neomake/neomake/pull/2116.

PR #2115, opened 2018-10-25
Erlang - integrate Gradualizer

erszcz:
[Gradualizer](https://github.com/josefs/Gradualizer) implements a gradual typing system for Erlang. Although still under development, it's already useful, while integration with Vim/NeoMake will hopefully foster early adoption a bit.

This PR refactors the existing Erlang `erlc` support code for reuse with the new `gradualizer` maker. The maker is not enabled by default, since Gradualizer still rarely crashes on some code files - this doesn't affect Vim, though. One has to opt-in with:

```
let g:neomake_erlang_enabled_makers = ['erlc', 'gradualizer']
```

~~The maker by default calls `gradualizer.sh` - a shim shell script which processes Gradualizer output so that it is consumable with a simple pattern - this script is not included in this PR, but will be PRed to Gradualizer~~.


tests/ft_erlang.vader
```diff
@@ -30,32 +30,29 @@ Execute (erlc: output to _build/neomake if _build is present):
     execute 'cd ' . old_cwd
   endtry
 
-Execute (erlc: rebar3 default profile - no rebar.config):
+Execute (erlang#EbinDirs: rebar3 default profile - no rebar.config):
   let old_cwd = getcwd()
   try
-    let root = tempname()
+    let root = tempname() . '/'
     call mkdir(root, 'p')
```
blueyed: Seems like vim73 does not like the extra slash here with the mkdir.

blueyed: Please also use `neomake#utils#Slash()` for Windows compatibility in the future.. (related: see utils in 119dc9be)

erszcz:
@blueyed Any remarks / necessary changes?

autoload/neomake/makers/ft/erlang/gradualizer.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+#set -x
+
+# Set to your gradualizer location or make sure the command is in PATH
+GRADUALIZER=gradualizer
```
blueyed: Maybe this could be `gradualizer=${GRADUALIZER:-gradulizer}` (and then use `$gradulizer` below), if this should be more easily configurable.

Also consider using `NEOMAKE_GRADUALIZER` first then, if it makes sense.

autoload/neomake/makers/ft/erlang/gradualizer.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if [ x`uname` = x"Darwin" ]; then
+    SED=gsed
+else
+    SED=sed
+fi
```
blueyed: This doesn't look that good: it shells out to `uname` for once.
Is GNU sed really required?
And shouldn't the errorformat just handle this (the replacement) maybe even?

autoload/neomake/makers/ft/erlang/gradualizer.sh
```diff
@@ -0,0 +1,15 @@
...
+
+$GRADUALIZER $@ | $SED '/\(.*on line \([0-9]*\)\)/ {s//\2:\1/}' | sort -n | uniq
```
blueyed: Additionally: why the `sort` and `uniq`?

blueyed (commented):
I think it looks really decent already, but there is no rush to merge it - as you can see from your last commit there is usually stuff that pops up while using it.

As for `gradualizer.sh` and only skimming this quickly I think it might not really be necessary after all, since `errorfornat` and or post-processing could also handle this?!

autoload/neomake/makers/ft/erlang/gradualizer.sh
erszcz: GNU and BSD sed do behave differently with the present command and I was not able to come up with a sed script/command which would achieve the same but work for both of them.
Then, it's arguable, but I think more probable that a developer using a Mac will have gsed installed, than a Linux person a BSD sed installed.

I'll look into `errorformat`, maybe you're right that it's possible to ditch the shell script altogether.

erszcz:
> [...] as you can see from your last commit there is usually stuff that pops up while using it.
Definitely. I just had some spare time and figured it couldn't hurt to ask. Thanks for the pointers!

autoload/neomake/makers/ft/erlang/gradualizer.sh
erszcz: Ahhh, my mistake. The surplus `let args += [expand('%')]` in `neomake#makers#ft#erlang#GradualizerArgs` was the reason, because the file was processed twice.

PR #2123, opened 2018-11-06
Delete parse-error.zsh

mattdodge:
I'm not sure what this file is supposed to do, it was added in this commit (https://github.com/neomake/neomake/commit/177e2b1f9a13dfe788cc9fd720b9dd82d6e01b6a). Whenever I load a zsh session now though I see this error before the session is created:
```
/Users/matt/.dotfiles/vim/vim.symlink/plugged/neomake/tests/fixtures/input/zsh/parse-error.zsh:2: parse error near `\n'
```
It seems, according to this file at least `tests/fixtures/output/zsh/parse-error.zsh.stderr` that this is somehow expected but I can't imagine why. Removing the `parse-error.zsh` file causes the error to go away.


mattdodge:
Ok, I think I've traced this down a little bit. The problem seems to be that my dotfiles (I based it off of https://github.com/holman/dotfiles) will go through and source anything that has a `.zsh` file ending. This file obviously isn't intended to be run in a real environment, it seems to be there for tests.

In the meantime I can add a line like this to my `.zshrc` that will ignore any zsh files in a test folder:
```bash
config_files=(${config_files:#*test*})
```

Anyone who is using those dotfiles (repo has >5k stars) or any similar setup that looks for any `*.zsh* file will have an issue if they are using neomake. If you want to close this PR since it isn't technically the fault of neomake I'm good with that. But I'm happy to try and find a happy medium if that's the direction you want to go.

Genuine question: what is this testing? It seems like it wants to make sure an exit code 1 is spat out when a malformed zsh script is present but I can't imagine why.

blueyed:
Thanks!

I have `tests/ft_zsh.vader` locally, which uses this, but obviously not committed.

> my dotfiles (I based it off of https://github.com/holman/dotfiles) will go through and source anything that has a .zsh file ending

I do not think that is a good idea.. - because I could have injected any code into your system using this apparently?!
Should probably have some whitelist.

mattdodge:
That's a good idea, I did add a PR to that repo as well (https://github.com/holman/dotfiles/pull/311) but it just blacklists anything that's in a "test" folder. The security concern you raised is definitely a valid one, the whitelist seems like the better way to go.

If you only plan on running that test locally then it might make sense to remove or gitignore the entire `tests/fixtures/*/zsh` subtree too.

blueyed:
The local test file is rather old, and not of much use - it's just that I apparently added the fixture, but without the test using it.

Issue #1342, opened 2017-05-20
Add a command to clear/reset Neomake


blueyed:
This should:
 - clear all signs and errors
 - close the location/quickfix list
 - reset internal caches (like #395)

This should be done for project mode (quickfix list), for the buffer (file mode, location list), or globally.

Ref: https://github.com/SpaceVim/SpaceVim/issues/547

wsdjeg:
here is some idea:
1. neomake should only clear it's own signs

2. if there other info in the quickfix and location list, which is not generate by neomake, it will be very hard to check it.

blueyed:
> neomake should only clear it's own signs

That is the case already.

> if there is other info in the quickfix and location list, which is not generate by neomake, it will be very hard to check it.

Neomake uses its own location/quickfix lists, but user's could have added to it, of course.
This command would not clear any lists, but rather just close the window (and preferably only if it's from Neomake - which would be a new feature; possibly related to #924).

bluz71:
Yes, this would be very handy.

In the meantime, for those interested, this crude example mapping can be used to clear out the signs column:

```
noremap <silent> <leader>cls :sign unplace *<CR>:set signcolumn=auto<CR>
```

This unplaces all signs and then toggles the empty signs column away. Not good if you have other non-Neomake signs, but good enough for those of us who only use signs with Neomake.

wsdjeg:
@bluz71 after clear sign, when move cursor to the error line, the error info will still be printed.

bluz71:
Yeah, I don't know how to clear that.

We have to wait until `NeomakeReset` (or whatever it will be called) is implemented.

My hack mapping just clears the signs column only; which is all I needed.

somini:
Syntastic had this feature, it might be useful to steal some ideas.
https://github.com/vim-syntastic/syntastic/blob/96cc251075f3f9d290c5afd4adf1b64c1542d469/plugin/syntastic.vim#L252-L255

blueyed:
Some WIP I have in my vimrc:
```
function! MyneomakeCleanFile()
  let buf = bufnr('%')
  call neomake#highlights#ResetFile(buf)
  call neomake#signs#ResetFile(buf)
  call neomake#signs#CleanOldSigns(buf, 'file')
  call neomake#statusline#ResetCountsForBuf(buf)
endfunction
com! NeomakeCleanFile call MyneomakeCleanFile()

function! MyneomakeCleanProject()
[...]

PR #2124, opened 2018-11-13
Add commands for cleaning

blueyed:
#1342 

TODO:
- [x] docs

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -1689,26 +1689,30 @@ function! s:clean_for_new_make(make_info) abort
     if get(a:make_info, 'cleaned_for_make', 0)
         return
     endif
-    let file_mode = a:make_info.options.file_mode
     " XXX: needs to handle buffers for list entries?!
     " See "get_list_entries: minimal example (from doc)" in
     " tests/makers.vader.
-    if file_mode
-        let bufnr = a:make_info.options.bufnr
+    call neomake#_clean_errors(a:make_info.options)
+    let a:make_info.cleaned_for_make = 1
+endfunction
+
+" a:context: dictionary with keys:
+"  - file_mode
+"  - bufnr (required for file_mode)
+"  - make_id (used for logging)
+function! neomake#_clean_errors(context) abort
+    if a:context.file_mode
+        let bufnr = a:context.bufnr
         if has_key(s:current_errors['file'], bufnr)
             unlet s:current_errors['file'][bufnr]
         endif
         call neomake#highlights#ResetFile(bufnr)
-        call neomake#log#debug('File-level errors cleaned.',
-                    \ {'make_id': a:make_info.options.make_id, 'bufnr': bufnr})
+        call neomake#log#debug('File-level errors cleaned.', a:context)
     else
-        " TODO: test
-        for buf in keys(s:current_errors.project)
-            unlet s:current_errors['project'][buf]
-            call neomake#highlights#ResetProject(+buf)
-        endfor
+        let s:current_errors['project'] = {}
+        call neomake#highlights#ResetProject()
+        call neomake#log#debug('Project-level errors cleaned.', a:context)
     endif
-    let a:make_info.cleaned_for_make = 1
 endfunction
 
 " Change to a job's cwd, if any.
--- a/autoload/neomake/cmd.vim
+++ b/autoload/neomake/cmd.vim
@@ -68,3 +68,38 @@ endfunction
 function! neomake#cmd#complete_jobs(...) abort
     return join(map(neomake#GetJobs(), "v:val.id.': '.v:val.maker.name"), "\n")
 endfunction
+
+function! s:is_neomake_list(list) abort
+    if empty(a:list)
+        return 0
+    endif
+    return a:list[0].text =~# ' nmcfg:{.\{-}}$'
+endfunction
+
+function! neomake#cmd#clean(file_mode) abort
+    let buf = bufnr('%')
+    call neomake#_clean_errors({
+          \ 'file_mode': a:file_mode,
+          \ 'bufnr': buf,
+          \ })
+    if a:file_mode
+        if s:is_neomake_list(getloclist(0))
+            call setloclist(0, [], 'r')
+            if get(g:, 'neomake_open_list', 0)
+                lclose
+            endif
+        endif
+        call neomake#signs#ResetFile(buf)
+        call neomake#statusline#ResetCountsForBuf(buf)
+    else
+        if s:is_neomake_list(getqflist())
+            call setqflist([], 'r')
+            if get(g:, 'neomake_open_list', 0)
+                cclose
+            endif
+        endif
+        call neomake#signs#ResetProject()
+        call neomake#statusline#ResetCountsForProject()
+    endif
+    call neomake#EchoCurrentError(1)
+endfunction
--- a/autoload/neomake/cmd.vim
+++ b/autoload/neomake/cmd.vim
@@ -85,18 +85,14 @@ function! neomake#cmd#clean(file_mode) abort
     if a:file_mode
         if s:is_neomake_list(getloclist(0))
             call setloclist(0, [], 'r')
-            if get(g:, 'neomake_open_list', 0)
-                lclose
-            endif
+            lclose
         endif
         call neomake#signs#ResetFile(buf)
         call neomake#statusline#ResetCountsForBuf(buf)
     else
         if s:is_neomake_list(getqflist())
             call setqflist([], 'r')
-            if get(g:, 'neomake_open_list', 0)
-                cclose
-            endif
+            cclose
         endif
         call neomake#signs#ResetProject()
         call neomake#statusline#ResetCountsForProject()
--- a/autoload/neomake/highlights.vim
+++ b/autoload/neomake/highlights.vim
@@ -51,11 +51,18 @@ else
         endif
     endfunction
 endif
+
 function! neomake#highlights#ResetFile(buf) abort
     call s:reset('file', a:buf)
 endfunction
-function! neomake#highlights#ResetProject(buf) abort
-    call s:reset('project', a:buf)
+function! neomake#highlights#ResetProject(...) abort
+    if a:0  " deprecated a:buf
+        call neomake#log#warn_once('neomake#highlights#ResetProject does not use a:buf anymore.',
+                    \ 'deprecated-highlight-resetproject')
+    endif
+    for buf in keys(s:highlights['project'])
+        call s:reset('project', +buf)
+    endfor
 endfunction
 
 function! neomake#highlights#AddHighlight(entry, type) abort
--- a/autoload/neomake/quickfix.vim
+++ b/autoload/neomake/quickfix.vim
@@ -263,7 +263,6 @@ function! neomake#quickfix#FormatQuickfix() abort
         let &breakindentopt = 'shift:'.(b:neomake_start_col + 1)
     endif
 
-    call neomake#signs#CleanOldSigns(buf, 'file')
     call neomake#signs#Reset(buf, 'file')
     call neomake#signs#PlaceSigns(buf, signs, 'file')
 
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -74,6 +74,12 @@ dispatch.
 :NeomakeCancelJobs
                         Terminate all jobs.
 
+                                                               *:NeomakeClean*
+                                                              *:NeomakeClean!*
+:NeomakeClean[!]
+                        Clean signs, highlights etc. for file-mode, or
+                        project-mode (with bang).
+
 ==============================================================================
 2.1. Toggle commands                                          *neomake-toggle*
 
--- a/plugin/neomake.vim
+++ b/plugin/neomake.vim
@@ -22,6 +22,8 @@ command! -bang NeomakeCancelJobs call neomake#CancelJobs(<bang>0)
 command! -bang -bar -nargs=? -complete=customlist,neomake#cmd#complete_makers
             \ NeomakeInfo call neomake#debug#display_info(<bang>0, <f-args>)
 
+command! -bang -bar NeomakeClean call neomake#cmd#clean(<bang>1)
+
 " Enable/disable/toggle commands.  {{{
 function! s:handle_disabled_status(scope, disabled, verbose) abort
     if a:scope is# g:
--- a/tests/clean.vader
+++ b/tests/clean.vader
@@ -1,0 +1,72 @@
+Include: include/setup.vader
+
+Execute (NeomakeClean: no-op):
+  NeomakeClean
+  NeomakeClean!
+
+Execute (NeomakeClean: file-mode):
+  Save g:neomake_open_list
+  let g:neomake_open_list = 2
+
+  new
+  let win_count = winnr('$')
+  CallNeomake 1, [g:entry_maker]
+  AssertEqual winnr('$'), win_count + 1, "loclist appeared"
+
+  NeomakeClean
+  AssertEqual winnr('$'), win_count + 1, "loclist was not closed"
+
+  lclose
+  bwipe
+
+Execute (NeomakeClean: file-mode (customqf)):
+  Save g:neomake_open_list
+  let g:neomake_open_list = 2
+
+  try
+    call neomake#quickfix#enable(1)
+
+    new
+    let win_count = winnr('$')
+    CallNeomake 1, [g:entry_maker]
+    AssertEqual winnr('$'), win_count + 1, "loclist appeared"
+
+    NeomakeClean
+    AssertEqual winnr('$'), win_count, "loclist was closed"
+
+    bwipe
+  finally
+    call neomake#quickfix#disable()
+  endtry
+
+Execute (NeomakeClean: project-mode):
+  Save g:neomake_open_list
+  let g:neomake_open_list = 2
+
+  new
+  CallNeomake 0, [g:entry_maker]
+
+  NeomakeClean!
+
+  cclose
+  bwipe
+
+Execute (NeomakeClean: project-mode (customqf)):
+  Save g:neomake_open_list
+  let g:neomake_open_list = 2
+
+  try
+    call neomake#quickfix#enable(1)
+
+    new
+    let win_count = winnr('$')
+    CallNeomake 0, [g:entry_maker]
+    AssertEqual winnr('$'), win_count + 1, "qflist appeared"
+
+    NeomakeClean!
+    AssertEqual winnr('$'), win_count, "qflist was closed"
+
+    bwipe
+  finally
+    call neomake#quickfix#disable()
+  endtry
--- a/tests/integration.vader
+++ b/tests/integration.vader
@@ -966,7 +966,6 @@ Execute (neomake#GetCurrentErrorMsg):
     \ ]
   endfunction
   CallNeomake 1, [maker]
-  NeomakeTestsWaitForFinishedJobs
 
   AssertEqual neomake#GetCurrentErrorMsg(), 'my_maker: error 1.1 (E)'
   normal! o
@@ -976,6 +975,17 @@ Execute (neomake#GetCurrentErrorMsg):
   normal! A12
   normal! l
   AssertEqual neomake#GetCurrentErrorMsg(), 'my_maker: error 3.2 (0)'
+
+  " Successful maker clears error(s) in file mode.
+  CallNeomake 1, [g:true_maker]
+  AssertEqual neomake#GetCurrentErrorMsg(), ''
+
+  " Successful maker clears error(s) in project mode.
+  CallNeomake 0, [maker]
+  AssertEqual neomake#GetCurrentErrorMsg(), 'my_maker: error 3.2 (0)'
+
+  CallNeomake 0, [g:true_maker]
+  AssertEqual neomake#GetCurrentErrorMsg(), ''
   bwipe!
 
 Execute (Neomake! clippy):
--- a/tests/main.vader
+++ b/tests/main.vader
@@ -12,6 +12,7 @@ Include (Cancellation): cancellation.vader
 Include (Compat): compat.vader
 Include (Completion): completion.vader
 Include (Config): config.vader
+Include (Clean command): clean.vader
 Include (Current working dir): cwd.vader
 Include (Custom Quickfix): customqf.vader
 Include (Debug/Checks/Feedback): debug.vader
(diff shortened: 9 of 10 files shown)


blueyed:
@hertogb
Please give it a shot already - the tests need to be improved (obviously).

hertogp:
I'll have a stab at it, but my vim fu is nowhere near what's in neomake... so please be patient.

hertogp:
Ok, I'm an occasional git(hub) user and I'm a little confused.  To test out/add to the clean branch I did:

```zsh
cd ~/.config/nvim/repos/github.com/neomake
git pull
git checkout clean
```
and started looking around, missing the fact that I landed on second to last commit (ie the WIP: add NeomakeClean command'), 
which shows here as:

```
commit ab3f0d0cb8b46c2b080f11763aa2c19213fc640b
Author: <snip>
Date:   Tue Nov 13 20:11:00 2018 +0100
    WIP: add NeomakeClean command
``` 
What did I miss?

hertogp:
Next I added a function to `autoload/neomake.vim` in the style of (and next to) the other functions that manipulate `s:current_errors`:
```vim
function! neomake#ClearErrorInfo(...) abort
    let buf = a:0 ? a:1 : 0
    if buf && has_key(s:current_errors.file, buf)
        unlet s:current_errors['file'][buf]
    else
        for buf in keys(s:current_errors.project)
            unlet s:current_errors['project'][buf]
        endfor
    endif
endfunction
```
and added it to the `neomake#cmd#clean` function:
```vim
function! neomake#cmd#clean(file_mode) abort
    if a:file_mode
        let buf = bufnr('%')
        call neomake#highlights#ResetFile(buf)
        call neomake#signs#ResetFile(buf)
        call neomake#signs#ResetFile(buf)
        call neomake#statusline#ResetCountsForBuf(buf)
        call neomake#ClearErrorInfo(buf)
    else
        call neomake#highlights#ResetProject()
        call neomake#signs#ResetProject()
        call neomake#signs#ResetProject()
        call neomake#statusline#ResetCountsForProject()
        call neomake#ClearErrorInfo()
    endif
    call neomake#EchoCurrentError(1)
endfunction
```
(the double `neomake#signs#Reset...` threw me a bit at first, until I read your comments :)
and initial testing for 'file' shows it works.

The thinking was to keep add/get/del etc.. to/from s:current_errors in 1 place and in the same style.

hertogp:
Since I can't push to the clean branch anyway, I'm more than happy to just test your code if I can actually get it to check out ;)  So any idea what I'm doing wrong so I can checkout the correct commit and test it for you?

hertogp:
Ah, nevermind.  It looks like I checked out clean before you committed your changes for `_clean_errors` ... guess you were faster than me ;)
I'll fix the conflicts I created and test out your version.
Sorry for the noise...

hertogp:
#### `NeomakeClean` works, `NeomakeClean!` doesn't remove the signs.  

Calling `ResetProject()` twice does not have the same effect as calling `ResetFile(buf)` twice due to the `for` loop in ResetProject looping over bufnrs in `placed_signs.project`:

```vim
function! neomake#signs#ResetProject() abort
    for buf in keys(s:placed_signs.project)
        call neomake#signs#CleanOldSigns(buf, 'project')
        call neomake#signs#Reset(buf, 'project')
    endfor
endfunction
```

Note: I am not that familiar with Neomake's code, so take my reasoning below with a grain of salt ;)

 In the first call to `ResetProject()` from `neomake#cmd#clean` the for loop ensures:
- old signs for any buffers in `last_placed_signs.project` get cleared first, and
- current signs for the same buffers get retired (by *moving* them to `last_placed_signs.project`)

In the second call to `ResetProject` from `neomake#cmd#clean` the for loop is skipped
because `placed_signs` doesn't contain any buffer/signs any more (the were retired in previous call) and the call to `CleanOldSigns` never happens, so they stay visible.

##  First solution

Adding an extra `CleanOldSigns`-call in the for loop takes care of this:
```vim
function! neomake#signs#ResetProject() abort
    for buf in keys(s:placed_signs.project)
        call neomake#signs#CleanOldSigns(buf, 'project')
        call neomake#signs#Reset(buf, 'project')
        call neomake#signs#CleanOldSigns(buf, 'project')
    endfor
endfunction
```
and allows for the removal of the second call to ResetProject in:

```vim
function! neomake#cmd#clean(file_mode) abort
    let buf = bufnr('%')
    call neomake#_clean_errors({
          \ 'file_mode': a:file_mode,
          \ 'bufnr': buf,
          \ })
    if a:file_mode
        call setloclist(0, [], 'r')  " TODO: only with Neomake's list.
        lclose

        " HACK: improve API
        call neomake#signs#ResetFile(buf)
        call neomake#signs#ResetFile(buf)
        call neomake#statusline#ResetCountsForBuf(buf)
    else
        call setqflist([], 'r')  " TODO: only with Neomake's list.
        cclose

        call neomake#signs#ResetProject()              " <--- only called once
        call neomake#statusline#ResetCountsForProject()
    endif
    call neomake#EchoCurrentError(1)
endfunction
```
and calling `NeomakeClean!` now removes the signs as well.

Note: `NeomakeClean` already works, but the same type of change could be applied to `ResetFile(buf)`:
```vim
function! neomake#signs#ResetFile(bufnr) abort
    call neomake#signs#CleanOldSigns(a:bufnr, 'file')
    call neomake#signs#Reset(a:bufnr, 'file')
    call neomake#signs#CleanOldSigns(a:bufnr, 'file')
endfunction
```
and remove the 2nd call to `ResetFile(bufnr)` in `neomake#cmd#clean` and it still works.

I guess the idea is `CleanOldSigns` removes retired signs, so first any previously lingering, retired signs are removed first (just in case), then the current signs are retired and immediately removed.

## Second solution

Alternatively, you could also consider giving `Reset` the responsibility of the pre-cleanup before
retiring signs from a buffer like so:

```vim
function! neomake#signs#Reset(bufnr, type) abort
    if has_key(s:placed_signs[a:type], a:bufnr)
       " first - ensure lingering signs are removed first
        call neomake#signs#CleanOldSigns(a:bufnr, a:type) 
       " then - safely retire current signs
        let s:last_placed_signs[a:type][a:bufnr] = s:placed_signs[a:type][a:bufnr]
        unlet s:placed_signs[a:type][a:bufnr]
    endif
endfunction
```
In which case `ResetProject()` would become:

```vim
function! neomake#signs#ResetProject() abort
    for buf in keys(s:placed_signs.project)
        call neomake#signs#Reset(buf, 'project')
        call neomake#signs#CleanOldSigns(buf, 'project')
    endfor
endfunction
```
the same holds true for `ResetFile` and `neomake#cmd#clean` would only have to call them only once:

```vim
function! neomake#cmd#clean(file_mode) abort
    let buf = bufnr('%')
    call neomake#_clean_errors({
          \ 'file_mode': a:file_mode,
          \ 'bufnr': buf,
          \ })
    if a:file_mode
        call setloclist(0, [], 'r')  " TODO: only with Neomake's list.
        lclose
        call neomake#signs#ResetFile(buf)
        call neomake#statusline#ResetCountsForBuf(buf)
    else
        call setqflist([], 'r')  " TODO: only with Neomake's list.
        cclose
        call neomake#signs#ResetProject()
        call neomake#statusline#ResetCountsForProject()
    endif
    call neomake#EchoCurrentError(1)
endfunction
```
I tried the above and both `NeomakeClean` and `NeomakeClean` still work, but haven't mastered the art of vader testing so I'm not sure if this breaks anything but it does make for a cleaner API.

blueyed:
Thanks for your feedback!

First, I am also not _that_ familiar with the interfaces for cleaning signs and highlights, it has been quite like that already when I took over maintainership - and it is quite messy indeed.
So anything to make this better is good here already, and I think adding those commands forces us to get this a bit more straight luckily.. :)

I will look into your suggestions.

Regarding tests: doesn't `make test` work for you?

hertogp:
Oh, `make test` works, but I'm trying to make sense of its output.  As far as I can tell, the change (2nd solution) changes the behaviour of ResetProject(), which makes 1 test in signs.vader fail:

```bash
   (398/616) [EXECUTE] Signs are not re-used / wiped across modes
      > [D +0.01]: [-.-:305:2] Placing sign: sign place 5000 line=1 name=neomake_file_err buffer=305.
      > [debug  ]: [-.-:305:2] Placing sign: sign place 5001 line=1 name=neomake_project_warn buffer=305.
      > [debug  ]: Cleaning old signs in buffer 305: {5001: neomake_project_warn}.
      > [debug  ]: Unplacing sign: sign unplace 5001 buffer=305.
      > [debug  ]: Removing signs: {file: {}, project: {}}.
    (398/616) [EXECUTE] (X) Message 'Removing signs: {file: {}, project: {305: {5001: neomake_project_warn}}}.' not found.
      > in <SNR>23_AssertNeomakeMessage (line 99)
      >   throw "Message '".a:msg."' not found."
      > ~/.config/nvim/repos/github.com/neomake/neomake/tests/signs.vader:73: AssertNeomakeMessage 'Removing signs: {file: {}, project: {'.bufnr.': {5001: neomake_project_warn}}}.'
    > NOTE: 1 error(s) in teardown.
```

probably because previously `ResetProject` would leave lingering signs in `last_placed_signs.project` and the last solution above actually removes them, so `CleanAllOldSigns` will not find any left-over signs after a call to `ResetProject` anymore.

```vim
Execute (Signs are not re-used / wiped across modes):
  new
  let bufnr = bufnr('%')
  call neomake#signs#PlaceSigns(bufnr('%'),
  \ [{'type': 'E', 'bufnr': bufnr('%'), 'lnum': 1}], 'file')
  call neomake#signs#PlaceSigns(bufnr('%'),
  \ [{'type': 'W', 'bufnr': bufnr('%'), 'lnum': 1}], 'project')

  AssertNeomakeMessage 'Placing sign: sign place 5000 line=1 name=neomake_file_err buffer='.bufnr.'.'
  AssertNeomakeMessage 'Placing sign: sign place 5001 line=1 name=neomake_project_warn buffer='.bufnr.'.'
  call neomake#signs#ResetProject()
  call neomake#signs#CleanAllOldSigns('project')
  AssertNeomakeMessage 'Removing signs: {file: {}, project: {'.bufnr.': {5001: neomake_project_warn}}}.'

  AssertEqual neomake#signs#by_lnum(bufnr), {'1': [5000, 'neomake_file_err']}
  call neomake#signs#RedefineSign('neomake_file_warn', {})
  bwipe
```
But I'm deducing/guessing, because I'm not familiar with vader.vim & its testcases.

blueyed:
Your guess is correct apparently, as can be seen by the previous message:

> Unplacing sign: sign unplace 5001 buffer=305.

(a lot of the tests are based on debug log output / messages)

It seems you prefer the 2nd solution (and it also looks cleaner to me).

Please post any git-diff you have and I will apply it here.

hertogp:
git diff
```diff
diff --git a/autoload/neomake/cmd.vim b/autoload/neomake/cmd.vim
index a5f538f..206c302 100644
--- a/autoload/neomake/cmd.vim
+++ b/autoload/neomake/cmd.vim
@@ -78,17 +78,11 @@ function! neomake#cmd#clean(file_mode) abort
     if a:file_mode
         call setloclist(0, [], 'r')  " TODO: only with Neomake's list.
         lclose
-
-        " HACK: improve API
-        call neomake#signs#ResetFile(buf)
         call neomake#signs#ResetFile(buf)
         call neomake#statusline#ResetCountsForBuf(buf)
     else
         call setqflist([], 'r')  " TODO: only with Neomake's list.
         cclose
-
-        " HACK: improve API
-        call neomake#signs#ResetProject()
         call neomake#signs#ResetProject()
         call neomake#statusline#ResetCountsForProject()
     endif
diff --git a/autoload/neomake/signs.vim b/autoload/neomake/signs.vim
index 5d2e821..4fe836b 100644
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -13,22 +13,23 @@ let s:last_placed_signs = {'project': {}, 'file': {}}
 exe 'sign define neomake_invisible'
 
 " Reset signs placed by a :Neomake! call
-" (resetting signs means the current signs will be deleted on the next call to ResetProject)
 function! neomake#signs#ResetProject() abort
     for buf in keys(s:placed_signs.project)
-        call neomake#signs#CleanOldSigns(buf, 'project')
         call neomake#signs#Reset(buf, 'project')
+        call neomake#signs#CleanOldSigns(buf, 'project')
     endfor
 endfunction
 
 " Reset signs placed by a :Neomake call in a buffer
 function! neomake#signs#ResetFile(bufnr) abort
-    call neomake#signs#CleanOldSigns(a:bufnr, 'file')
     call neomake#signs#Reset(a:bufnr, 'file')
+    call neomake#signs#CleanOldSigns(a:bufnr, 'file')
 endfunction
 
 function! neomake#signs#Reset(bufnr, type) abort
     if has_key(s:placed_signs[a:type], a:bufnr)
+        " clean any lingering, already retired signs
+        call neomake#signs#CleanOldSigns(a:bufnr, a:type)
         let s:last_placed_signs[a:type][a:bufnr] = s:placed_signs[a:type][a:bufnr]
         unlet s:placed_signs[a:type][a:bufnr]
     endif
```
I also removed the comment `"(resetting signs means ...)` since that's no longer true.

hertogp:
Regarding qf/ll windows: if `NeomakeClean(!)` clears the lists therein, one might as well close any open qf/ll window 'cause it'll be empty.

The only problem I noticed is that both qf/ll windows seem fair game to any tool dumping (and overwriting) the lists in qf/ll.  If I do:
- `:Neomake!` -> fills quickfix list with err/warn msgs from pylint3 (in my case here),
- `:grep qry *.py` -> overwrites the quickfix list with grep's output
- `:NeomakeClean!` -> that throws away my grep results in quickfix

The other way around (first `:grep qry *.py` and then `:Neomake!`) also replaces any previous content.

So other tools may replace Neomake's contents of the quickfix window (in this case) and I can't see an immediate solution to determine if the contents was last set by Neomake itself before clearing the list.

One approach would be to go through the list and only remove individual entries recognized by Neomake as its own and then only close the window if the resulting list is empty.  Not sure if that's feasible though.

hertogp:
To answer your question re `neomake#quickfix#enable()`, I'm not using that currently so I just tried it out and there's no difference in behaviour described earlier, except that when `:grep qry *.py` replaces the quickfix list, the window still carries signs placed earlier by `:Neomake!`

Its tough to play nice with other tools when they consider qf/ll windows & contents fair game, I guess.

Apart from that, I don't see any error numbers anymore (like C306) which I sometimes use to disable specific linting checks via comments in source code.

blueyed:
> Regarding qf/ll windows: if `NeomakeClean(!)` clears the lists therein, one might as well close any open qf/ll window 'cause it'll be empty.

Yeah, the question if it should be cleared in the first place.

Re custom qflist: I've mentioned this since with this it is possible already to detect if it is from Neomake (it uses markers for makers).  That is quite a hack though, and we should use another method for handling this in general.  For now I would say to not do `:lclose`/`:cclose` etc.

> except that when `:grep qry *.py` replaces the quickfix list, the window still carries signs placed earlier by `:Neomake!`

Oh, that's a bug, will look into it.

> Apart from that, I don't see any error numbers anymore (like C306) 

Works for me:
2018-11-18-110639_952x515_escrotum

hertogp:
Nice! Exactly what I wanted to achieve... but this is what I have at the moment:

screenshot_2018-11-18_11-43-00

I can see the `type` and `nr` in the quickfix list entries, but haven't figured out how to get it to display that information nicely in qf/ll windows.  I setup pylint3 for testing `Neomake! makeprg` like this:
```vim
augroup PyMake
    au!
    au FileType python set makeprg=pylint3\ -rn\ -ftext\ --msg-template='{path}:{msg_id}:{line}:{column}:{msg}'\ *.py
    au FileType python set efm=%f:%t%n:%l:%c:%m,*%#\ %#%m
augroup end

" together with
nnoremap <f3> :Neomake! makeprg<cr>
nnoremap <space><f3> :NeomakeClean!<cr>
```

So `efm` seems to picking up the error codes (given that they're present in the qf--list entries as seen in the topright window of the screenshot), but it won't actually show up in the qf-window.

Any hints on what I'm missing?

hertogp:
btw, without the `call neomake#quickfix#enable()` in my init.vim, I get:
screenshot_2018-11-18_12-10-05

Here, the `type` and `nr` are shown (although type W becomes warning and any leading zero's in the err nr have gone due to parsing by `efm`).

blueyed:
Why don't you use the existing pylint maker?
You can `let g:neomake_python_pylint_exe = 'pytlint3'` to force a specific command.

No need for autocommands then really, but just use `nnoremap <f3> :Neomake! pylint<cr>`.

If you like to use different makers per filetype(s), then set `g:neomake_enabled_makers` via FileType autocommand, and just use `:Neomake!` for the mapping.

However, there might be an issue with the custom qf list there, where it does not handle type/nr correctly - worth checking based on your custom config.

hertogp:
I'll have to do some diggin' then: at the moment I'm trying out nvim without including my vim's runtimepath or its configuration files, so its just nvim & what I install/configure there.. and no custom qf thingies thus far except for 'chemzqm/unite-location's plugin which uses the qf/ll contents as a source for listing/filtering/jumping etc .. but it doesn't seem to manipulate the qf/ll entries or their display.

Till now, it completely escapes me how to manipulate the *display* of entries in qf/ll lists in their respective window.  Last time I checked (its on old itch) it was impossible (in vim) to do so, but you've shown the contrary to be true.  Unfortunately looking through `neomake#quickfix#FormatQuickfix()` leaves me none the wiser.

blueyed:
Re the custom qf list: it does not display the `nr`, but only matches it in the text (https://github.com/neomake/neomake/blob/72a997ad49a4d7fbea43eb40cfa3bd559bca1d91/syntax/neomake/python.vim#L1) - not sure if this is really ideal though.
As for the type: this is meant to be reflected in the sign that is used in the quickfix list itself (which I think is good).
It basically replaces the quickfix entries with custom text, handles/limits moving in the list and applies syntax highlighting.
This was initially done by @tweekmonster.

As for Neomake's configuration: I suggest configuring it primarily for the current file, and to use automaking (see `:help neomake-automake`) - I use this:
```vim
call neomake#configure#automake('nw', 500)
let g:neomake_python_enabled_makers = ['python', 'flake8']
```
(use "pylint" instead of "flake8" if you like that better)

hertogp:
Ah! Seems like I need to concentrate on getting the error codes into the message text so it gets displayed and let syntax highlighting do the rest.  You mind sharing the `efm` setting you use for `flake8`?

re `automake` - I find constantly being yelled at by pylint/flake8 really distracting, so I actually have/use:
```vim
let g:neomake_python_enabled_makers = ['pylint3']
let g:neomake_python_pylint3_maker = {
      \'args': ['-rn', '-ftext',
      \'--msg-template', "{path}:{msg_id}:{line}:{column}:{msg}", '%t'
      \],
      \'exe': 'pylint3',
      \'errorformat': '%f:%t%n:%l:%c:%m',
      \'append_file': 0}
" with
nnoremap <f2> :Neomake<cr>
nnoremap <space><f2> :NeomakeClean<cr>
```
working per module/file and commanding the feedback of the linter when I feel like some constructive criticism ;)

The other stuff I just setup to be able to test `Neomake!` and `NeomakeClean!` using two .py files in a test directory.

blueyed:
A bit vague, but somehow on purpose:

```diff
diff --git a/doc/neomake.txt b/doc/neomake.txt
index 2a8d397d..a8799676 100644
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -74,6 +74,12 @@ dispatch.
 :NeomakeCancelJobs
                         Terminate all jobs.
 
+                                                               *:NeomakeClean*
+                                                               *:NeomakeClean!*
+:NeomakeClean[!]
+                        Clean signs, highlights etc. for file-mode, or
+                        project-mode (with bang).
+
 ==============================================================================
 2.1. Toggle commands                                          *neomake-toggle*
```

hertogp:
Ok, I'll try it out.  No particular reason for the custom pylint maker other than `pylint` is not on my system and `pylint3` is, so I thought I had to create a custom maker instead of setting the `-exe` to pylint3?

Anyway, I figured out (again) that you can't change the way (n)vim initially displays the qf/ll list entries in their respective windows, but you can alter it afterwards via an `au BufReadPost` event, as long as you don't add/del lines.  Sometimes it helps to read the vim docs :grimacing: 

On another note.  In `neomake#quickfix#FormatQuickfix()` I saw the call pattern:
```vim
    call neomake#signs#CleanOldSigns(buf, 'file')
    call neomake#signs#Reset(buf, 'file')
    call neomake#signs#PlaceSigns(buf, signs, 'file')
```
If `neomake#signs#Reset(buf, type)` has indeed been changed to include a call to `CleanOldSigns`,
the 1st call is redundant and will have no effect (anymore).  Perhaps there are more places where this pattern is used...

Anyway, thanks for all your patience and feedback!

blueyed:
> pylint is not on my system and pylint3 is, so I thought I had to create a custom maker instead of setting the -exe to pylint3?

No, you can configure the `exe`: https://github.com/neomake/neomake/pull/2124#issuecomment-439685040

Fixed the "duplicate" calls to `neomake#signs#CleanOldSigns`.

blueyed:
BufReadPost

> Anyway, I figured out (again) that you can't change the way (n)vim initially
> displays the qf/ll list entries in their respective windows, but you can
> alter it afterwards via an `au BufReadPost` event, as long as you don't
> add/del lines. Sometimes it helps to read the vim docs 

Would `BufReadPost` be better than the `FileType` being used currently?

> Anyway, thanks for all your patience and feedback!

You're welcome.

I'm glad to finally have proper clean commands myself.. :)

hertogp:
First I'll try out the setup you've recommended and then dive into the details re qf/ll display/rewrite of entries.  

The `BufReadPost` example was given in `:he quickfix`, I'm not sure at all if that's in any way, better than `FileType` but if I figure that out, I'll let you know.

Come to think of it, it would be nice to have a enhanced `location`-window experience, similar to what is offered for the `quickfix`-window, so working in `file`-mode behaves similar to working in `project`-mode
when wading through a linter's complaints.

blueyed:
> Come to think of it, it would be nice to have a enhanced `location`-window experience, similar to what is offered for the `quickfix`-window, so working in `file`-mode behaves similar to working in `project`-mode
> when wading through a linter's complaints.

What do you mean?

Location list windows are of filetype=qf, i.e. very similar to quickfix
windows, but for a specific window instead of global.

`:Neomake` (file-mode) uses location lists, `:Neomake!` (project-mode) the
quickfix lists.

The custom quickfix handling gets also applied to location list windows, if
that's what you mean.

hertogp:
My bad... you're right.  I must have commented out the `call neomake#quickfix#enable()` and not realised it.  Goes to show what you get when opening your mouth without checking your facts ... :pensive: 

Please ignore the comment.

Issue #2126, opened 2018-11-16
phpmd is not configurable


frazjp65:
I would expect to be able to customize the settings for `phpmd` similar to how `phpcs` is customizable. It currently uses generic rules for `'codesize,design,unusedcode,naming'`.

PR #2127, opened 2018-11-16
Allow phpmd to be customized

frazjp65:
For issue #2126. 

If `g:neomake_php_phpmd_args_standard` is set, use that as the standard
for phpmd. Otherwise, continue to use the existing setting.

### DIFF
--- a/autoload/neomake/makers/ft/php.vim
+++ b/autoload/neomake/makers/ft/php.vim
@@ -36,8 +36,16 @@ function! neomake#makers#ft#php#phpcs() abort
 endfunction
 
 function! neomake#makers#ft#php#phpmd() abort
+    let l:args = ['%t', 'text']
+
+    if exists('g:neomake_php_phpmd_args_standard')
+        call add(l:args, expand(g:neomake_php_phpmd_args_standard))
+    else
+        call add(l:args, 'codesize,design,unusedcode,naming')
+    endif
+
     return {
-        \ 'args': ['%t', 'text', 'codesize,design,unusedcode,naming'],
+        \ 'args': args,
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/php.vim
```diff
@@ -36,8 +36,16 @@ function! neomake#makers#ft#php#phpcs() abort
 endfunction
 
 function! neomake#makers#ft#php#phpmd() abort
+    let l:args = ['%t', 'text']
+
+    if exists('g:neomake_php_phpmd_args_standard')
+        call add(l:args, expand(g:neomake_php_phpmd_args_standard))
```
blueyed: Why `expand()`?
Likely just coming/copied from https://github.com/neomake/neomake/blob/a17e955cf1959daa6a639879ec3851e92c722ccf/autoload/neomake/makers/ft/php.vim#L27, right?

blueyed:
Thanks for the submission, but using `_args` is better I think.

blueyed:
What would make more sense could be a `g:neomake_php_default_standard_args` setting that would then be used with phpcs and phpmd - basically renaming the existing one (but should have some kind of deprecation), and then using the new one also with phpmd.
That might make sense since it appears to be the same setting for different makers, no?

frazjp65:
Thanks, @blueyed. I didn't even realize that was an option. I just saw that phpcs needed a special way to do it, but didn't realize it was because of the `--standard`. It works without my change.

blueyed:
Glad it works.
I think phpcs would not require a special var by itself also, but apparently that slipped in (and is causing confusion already by itself.. ;))

Issue #2085, opened 2018-09-10
rust: cargo: use src_path instead of relative file_name


blueyed:
This sounds like we could avoid the whole `cargo metadata` and Cargo.toml guessing, and just `cd` to the buffer's dir?!

Might be something added only in recent versions though?

Ref: https://github.com/rust-lang/rust.vim/issues/259#issuecomment-419853344

blueyed:
Worked on in https://github.com/neomake/neomake/pull/2143.

PR #2076, opened 2018-09-06
rust: cargo: prefer primary spans

dpc:
This would be great improvement to the current situation, where the
first span is used, which is quite often not the right one.

Note: My VimL is poor, but I'm desperate to get this fixed.

See https://github.com/rust-lang/rust.vim/issues/174

Related to: #2075

PR #2143, opened 2018-11-24
Improve cargo/rustc maker

NickeZ:
I noticed a couple of issues with the cargo maker.

* Using workspaces I got invalid paths to the filenames.
* Using clippy I got wrong levels on the warnings (I think)
* Using clippy I also got "help" notices which doesn't really say much. Like "try this".

Please review this carefully since it is the first time I write in this language.


autoload/neomake/makers/ft/rust.vim
```diff
@@ -135,16 +135,26 @@ function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
...
-            if get(data, 'level', '') ==# 'warning'
-                let error.type = 'W'
-            else
-                let error.type = 'E'
+        let target = get(decoded, 'target', -1)
+        if type(target) == type({})
+            let filename = get(target, 'src_path', -1)
+            if type(filename) == type('')
+                let error.filename = filename
             endif
```
blueyed: Cool.  I assume this fixes/closes https://github.com/neomake/neomake/issues/2085 then.

NickeZ: yeah, as far as I understand

NickeZ:
I didn't run any tests, are there instructions on how to do that somewhere?

blueyed:
`make test` should work.
The failures are in `tests/isolated/highlights.vader` and `tests/ft_rust.vader`, and you can run them directly: `make tests/ft_rust.vader`.

NickeZ:
It seems like many tests fail because `bufnr` is offset by 1. what is `bufnr`?

blueyed:
It is the bufnr used in the quickfix/location list.
I've looked at the failure in ft_rust.vader, and the failing test appears to be the one that asserts that a primary span is used:

With
```diff
diff --git i/tests/ft_rust.vader w/tests/ft_rust.vader
index b6a0bcdc..d5be53b2 100644
--- i/tests/ft_rust.vader
+++ w/tests/ft_rust.vader
@@ -43,6 +43,8 @@ Execute (cargo: uses primary span):
 
   CallNeomake 1, [cargo]
 
+  Log neomake#utils#redir('ls!')
+
   AssertEqualQf getloclist(0), [{
   \ 'lnum': 6, 'bufnr': bufnr('%'), 'col': 5, 'valid': 1, 'vcol': 0, 'nr': 49,
   \ 'type': 'E', 'pattern': '',
```

you get:
```
      > 
  1  a   "[No Name]"                    line 1
  2  a   "[Vader]"                      line 0
  3 #a   "[Vader-workbench]"            line 0
  7 %a   "build/main.rs"                line 1
  8u     "/home/dpc/tmp/testfoo/build/main.rs" line 0
    ( 4/10) [EXECUTE] (X) Unequal Lists
      [{'lnum': 6, 'bufnr': 7, 'col': 5, 'valid': 1, 'vcol': 0, 'nr': 49, 'type': 'E', 'pattern': '', 'text': 'this function takes 1 parameter but 2 parameters were supplied: expected 1 parameter'}] should be equal to
      [{'lnum': 6, 'bufnr': 8, 'col': 5, 'valid': 1, 'vcol': 0, 'nr': 49, 'type': 'E', 'pattern': '', 'text': 'this function takes 1 parameter but 2 parameters were supplied: expected 1 parameter'}]
```

I.e. it expects to use buffer 7 here, but 8 is used - and this seems to be an unlisted buffer that Vim is creating.  It should use the existing one.

This was done in https://github.com/neomake/neomake/pull/2076.

We should also have a test for your change/fix btw, so please provide input/fixtures for what this is trying to fix after all.

blueyed:
Ok, fixing the tests now.  I.e. in ft_rust.vader the expected buffer is correct.

But I wonder about the fixture: https://github.com/neomake/neomake/blob/088bc329558e37091e93e8447d57f46df4cc6bac/tests/fixtures/rust/cargo_warning.json - shouldn't "src_path" match "filename"?  /cc @JelteF

blueyed:
Better now (although the question about the fixture still remains).
We should also have tests to cover the diff completely (https://codecov.io/gh/neomake/neomake/compare/a4cf8677353a391071dade1a7a5108f75314e939...a9ba92c6ff3a81866a54b6c0efa9f0669a941270/diff).

NickeZ:
My guess is that "src_path" is the file you are currently compiling and that "file_name" is where the error is unfortunately.. Or the other way around. I'll dig into rustc/cargo and see if I can find any docs on what the output should be..

I'm guessing though that the "message" field comes from the compiler and that the "target" field comes from cargo.

NickeZ:
For all the manual testing I did it worked to replace `file_name` with `src_path`, but I will generate some test fixtures for different cases (especially clippy, since that is newly supported)

NickeZ:
@blueyed I should probably change it back to use file_name. But how do I get the path to the workspace or project? Is there some global variable with this value or some function I can call to get it?

blueyed:
> But how do I get the path to the workspace or project?

`neomake#utils#get_project_root` (https://github.com/neomake/neomake/blob/8d9ab61e6cd08dfb42e2809407141623011ec322/autoload/neomake/utils.vim#L566)

We/you should use `let g:neomake#makers#ft#rust#project_root_files = ['Cargo.toml']` (in `rust.vim`) for this.

There is also `s:get_cargo_maker_cwd` though already.

NickeZ:
Will this get the top-most Cargo.toml file that contains `[workspace]` otherwise the closest?

NickeZ:
It needs to handle the case when you have an example project inside a project (get the first found when going upward). And when you have a project inside a workspace (find the one with `[workspace]` in it).

Rahix:
`s:get_cargo_maker_cwd` handles workspaces correctly:

https://github.com/neomake/neomake/blob/7f834152836166f533ab4944792c6d0087d60f5a/autoload/neomake/makers/ft/rust.vim#L50-L62

NickeZ:
thanks! I didn't have time to work on this issue lately, but now it seems much easier to solve!

NickeZ:
I think my last commit fixes the invalid paths. If I understand correctly I now set the `cwd` parameter on the "clippy maker" and then I can use that parameter in the process function since all makers should have it. I still have to improve tests..

NickeZ:
Tests pass now, but coverage has decreased.

blueyed:
Re coverage: try rebasing it on master, but it looks just like there was uncovered code before already from a quick glance.

blueyed:
Would be cool to get the diff covered 100% of course.

NickeZ:
Yeah i already rebased.

Do you now if I can simulate that nighlty compiler is available?

I still need to write some more tests to improve coverage

blueyed:
> Do you know if I can simulate that nighlty compiler is available?

You can set the global: https://github.com/neomake/neomake/pull/2143/files#diff-05c39c0cff9fd1fb19ee0bcc89cd9553R4

But that requires to have separate tests in https://github.com/neomake/neomake/tree/master/tests/isolated then: there you can use `runtime` / `source` to reload the file (or just save/set it in the beginning to 1).

The isolated test is necessary because reloading the source break coverage recording.
Search for "monkey.*patch" in the existing tests.

Let me know if you need help and I'd be happy to provide the base for this.

blueyed:
btw: base report in codecov is missing because I used `[ci skip]` for the last commit.
You could rebase it on `master~1` (https://github.com/neomake/neomake/commit/e7cad7a9e25c1060387a165de2957021f1f6578b).

NickeZ:
> You could rebase it on master~1 (e7cad7a).

codecov seems to diff my two latests commits, so that trick didn't work..

NickeZ:
I need some guidance with the "isolated" tests. My attempt did not work.

I also tried closing/opening the PR to see if that triggered codecov but it didn't. It still diffs against the wrong branch point.

blueyed:
> codecov seems to diff my two latests commits, so that trick didn't work..

It looks at the diff coverage, which is OK in general.
But it complains that there is no base to compare against, since that did not upload coverage (`[ci skip]`).  Rebasing against `master~1` should get rid of that, but is not that important.

I've fixed the isolated test: `Save` gets restored at the end of a `Execute` block.. :)

I'm also pushing a change to store the state in a global, so it can be changed/set in non-isolated tests.

blueyed:
I'll merge https://github.com/neomake/neomake/pull/2291 soon, so you rebase (and squash if you like) after that.
The isolatated test should not be needed anymore, at least for the rustup check.
But why do you want/need that in the first place?

autoload/neomake/makers/clippy.vim
```diff
@@ -1,39 +1,39 @@
...
             \ }
     else
-        return {
+        let maker = {
             \ 'exe': 'cargo',
             \ 'args': ['clippy'] + json_args,
             \ 'process_output': cargo_maker.process_output,
             \ }
     endif
+    " InitForJob will lazy initialize cwd for us
```
blueyed:
```suggestion
    " InitForJob will lazily initialize cwd for us.
```

autoload/neomake/makers/ft/rust.vim
```diff
@@ -34,7 +34,7 @@ function! s:get_cargo_workspace_root() abort
         if !empty(cd_back_cmd)
             exe cd_back_cmd
         endif
-        if v:shell_error
+        if v:shell_error || output ==# ''
```
blueyed: Why is the output check needed?
Is there something on stderr then in this case?  (check with `2>&1` at the end of them `cmd`)

autoload/neomake/makers/ft/rust.vim
```diff
@@ -123,6 +123,7 @@ endfunction
 " elements per line.
 function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
     let errors = []
+    let cwd = get(get(get(a:context, 'jobinfo', {}), 'maker'), 'cwd', '')
```
blueyed: I (would) assume that this should be there always.. so just access it directly - it being empty is probably something unexpected further down anyway, isn't it?

autoload/neomake/makers/ft/rust.vim
```diff
@@ -136,15 +137,18 @@ function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
...
-            endif
+        if get(data, 'level', '') ==# 'warning'
+            let error.type = 'W'
         else
-            let error.type = code_dict['code'][0]
-            let error.nr = code_dict['code'][1:]
+            let error.type = 'E'
+        endif
+        if code_dict isnot g:neomake#compat#json_null
+            if code_dict['code'][0] ==# 'W' || code_dict['code'][0] ==# 'E'
```
blueyed: `code_dict` might be -1 here, from the default above?!

blueyed: Maybe `g:neomake#compat#json_null` should be used as default above? (with `get(data, 'code', …)`)

autoload/neomake/makers/ft/rust.vim
```diff
@@ -223,8 +231,9 @@ function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
     return errors
 endfunction
 
-function! neomake#makers#ft#rust#FillErrorFromSpan(error, span) abort
-    let a:error.filename = a:span.file_name
+function! neomake#makers#ft#rust#FillErrorFromSpan(error, span, cwd) abort
+    let slash = neomake#utils#Slash()
```
blueyed: Could become `let s:slash = neomake#utils#Slash()` at the top of the script.

tests/isolated/highlights.vader
```diff
@@ -158,7 +158,7 @@ Execute (cargo error message):
   call neomake#Make(1, [cargo])
   NeomakeTestsWaitForFinishedJobs
 
-  let bufnr = bufnr('%')
+  let bufnr = bufnr('lib.rs')
```
blueyed: Does this pick up the same bufnr?  (it uses `file build/lib.rs` above)

btw: `Log neomake#utils#redir('ls!')` can be handy in this context.

blueyed:
Ok, master should have coverage again.

NickeZ:
Thanks for the review. I'll rebase now, but I'll wait with squashing until the PR is in a good shape for merge

autoload/neomake/makers/ft/rust.vim
NickeZ: This test (https://github.com/neomake/neomake/pull/2143/files#diff-563dcda53b905303c16ad35d06d7b658R12) generated empty output which made the json decode function fail. But the test is probably invalid. I don't really understand how the tests work.

blueyed: Yeah.. that tests creates an exe, but the output is not what you normally get.
I think using `NeomakeTestsGetMakerWithOutput` instead would be better.
You can run it as project maker using `CallNeomake 0, maker`.

The tests are messy/grown in several places, so feel free to ask, add notes etc.

autoload/neomake/makers/ft/rust.vim
NickeZ: fixed

autoload/neomake/makers/ft/rust.vim
NickeZ: fixed

autoload/neomake/makers/ft/rust.vim
NickeZ: fixed

tests/isolated/highlights.vader
NickeZ: `build/lib.rs` does also work and looks more correct.

tests/isolated/rust-nightly.vader
```diff
@@ -0,0 +1,25 @@
+Include: ../include/setup.vader
+
+Execute (Setup: monkeypatch autoload/neomake/clippy.vim):
+  " Monkeypatch to force enable rust nightly
+  Save g:neomake_clippy_rustup_has_nightly
+  let g:neomake_clippy_rustup_has_nightly = 1
+
+  runtime autoload/neomake/makers/clippy.vim
```
blueyed: This does not test anything..

You should be able to just `Save g:neomake_clippy_rustup_has_nightly` and `let g:neomake_clippy_rustup_has_nightly` in a test (`Execute` block) that needs a particular mode for "has nightly".

blueyed:
Looks like my commits (823bff2, 819a2ad, and ec3b53d) got lost when you force-pushed.. (https://github.com/neomake/neomake/compare/ec3b53d297ffb934c2622a75e374c8ac890b4a6d..e18d425194f023f473871b4cb77e40b82956a232)

tests/isolated/rust-nightly.vader
blueyed: The above comment was under the assumption that the global would be used always.
It works / is needed for the previous/current code.

NickeZ:
Sorry about that. I forgot to pull before rebasing on master.

NickeZ:
I've re-added your commits.

Should i use the `Restore` command in the teardown in the isolated test? https://github.com/junegunn/vader.vim/blob/master/test/feature/save-restore.vader#L15

blueyed:
> Should i use the Restore command in the teardown in the isolated test? 

That gets done in the global `After`: https://github.com/neomake/neomake/blob/e5b45bc74b44da5515e97295ac47fa8a6425b29d/tests/include/init.vim#L478

blueyed:
> Sorry about that. I forgot to pull before rebasing on master.

JFI: using `git push --force-with-lease` (instead of `-f`) avoids that.

NickeZ:
> JFI: using git push --force-with-lease (instead of -f) avoids that.

I wasn't aware, thank you!

blueyed:
@NickeZ 
You're welcome.  FWIW, I am using this Zsh function, which translates `-f` into it:
```zsh
gp () {
	local opt
	local -a git_opts push_opts
	while (( ${#@} ))
	do
		opt=$1 
		shift
		if [[ "$opt" == '-c' ]]
		then
			git_opts=(-c $1) 
			shift
		elif [[ "$opt" == '-f' ]]
		then
			push_opts+=('--force-with-lease') 
		else
			push_opts+=($opt) 
		fi
	done
	git $git_opts push $push_opts
}
```

NickeZ:
I rebased, so now there are only two commits, one fixes the paths and the other enables clippy messages.

I don't know how to increase coverage becuase I don't know what rustc errors trigger the code paths that are not tested. I guess the original author of this maker has a better idea.

blueyed:
@NickeZ have you given up / cleaning up your open PRs?
Sorry for the long wait - I have not read up on it again, but would appreciate feedback on if you were using this still etc.

NickeZ:
Cleaning up old PRs. I'm using rust-analyzer with the nvim built in LSP nowadays, so I don't have the context for this PR any more.

PR #2202, opened 2018-12-20
Fix type in java.vim

wsdjeg:
close https://github.com/SpaceVim/SpaceVim/issues/2280


blueyed:
Thanks for the report and PR, fixed in ecfa28e.

PR #2219, opened 2018-12-22
automake: increase delay with canceled timers/makes

blueyed:


blueyed:
Merry christmas! :gift:

PR #2294, opened 2019-01-21
ada: use gnatc to check both syntax and semantics

Earnestly:
The -gnats flag only checks syntax which misses quite a few cases that
will result in a build failure.


blueyed:
Thanks!

PR #1575, opened 2017-09-11
WIP: Appveyor for Windows tests

blueyed:
Have just been playing around with it a bit.
Input / help would be appreciated.

PR #2296, opened 2019-01-22
Set permissions on temporary files

### LINKED ISSUE #2293: failing test 498 in test suite
### Expected behavior
no error

### Steps to reproduce
Have umask 002.

### Actual output
```
    (498/686) [EXECUTE] Neomake uses temporary file for unsaved buffer
      > [debug  ]: [414.-:410:2] Calling Make with options {'ft': '', 'file_mode': 1, 'winid': 1386, 'enabled_makers': [{'errorformat': '%m', 'default_entry_type': '', 'exe': 'cat', 'append_file': 1, 'tempfile_enabled': 1}]}.
      > [debug  ]: [414.-:410:2] Running makers: unnamed_maker.
      > [debug  ]: [414.431:410:2] Using tempfile for unnamed buffer: "/tmp/nvimnvf5Lk/1555/neomaketmp.".
      > [verbose]: [414.431:410:2] Starting async job: cat /tmp/nvimnvf5Lk/1555/neomaketmp..
      > [debug  ]: [414.431:410:2] cwd: /home/niklas/projects/niklas/dotfiles/config/nvim/plugged/neomake.
    (498/686) [EXECUTE] (X) 'rw-r--r--' should be equal to 'rw-rw-r--'
      > /home/niklas/projects/niklas/dotfiles/config/nvim/plugged/neomake/tests/tempfiles.vader:23: AssertEqual getfperm(temp_file), 'rw-r--r--'
    > [debug  ]: === teardown: canceling jobs.
    > [debug  ]: [414.431:410:2] Canceling job.
    > [debug  ]: [414.431:410:2] Stopping Neovim job: 360.
    > [debug  ]: [414.431:410:2] output on stdout: ['line1', 'line2', ''].
    > [debug  ]: [414.431:410:2] Ignoring output (job was canceled).
    > [debug  ]: [414.431:410:2] exit: job was canceled.
    > [debug  ]: [414.431:410:2] Cleaning jobinfo.
    > [debug  ]: [414.-:410:2] Cleaning make info.
    > [debug  ]: Removing temporary file: "/tmp/nvimnvf5Lk/1555/neomaketmp.".
    > NOTE: 2 error(s) in teardown.
```

NickeZ:
This patch sets the file permissions on temporary files to 0644, which I presume was the intention because there was a test testing this. I'm not sure if it is necessary to check if setfperm exists. It seems like it was added to vim quite recently, in 2016: https://github.com/vim/vim/commit/8049253b96838b3584600e5ad229abad37a95b10

fixes #2293

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -508,6 +508,9 @@ endfunction
 
 function! neomake#utils#write_tempfile(bufnr, temp_file) abort
     call writefile(neomake#utils#get_buffer_lines(a:bufnr), a:temp_file, 'b')
+    if exists('*setfperm')
+        call setfperm(a:temp_file, 'rw-r--r--')
+    endif
 endfunction
 
 " Wrapper around fnamemodify that handles special buffers (e.g. fugitive).
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -509,7 +509,11 @@ endfunction
 function! neomake#utils#write_tempfile(bufnr, temp_file) abort
     call writefile(neomake#utils#get_buffer_lines(a:bufnr), a:temp_file, 'b')
     if exists('*setfperm')
-        call setfperm(a:temp_file, 'rw-r--r--')
+        let perms = getfperm(bufname(+a:bufnr))
+        if empty(perms)
+            let perms = 'rw-------'
+        endif
+        call setfperm(a:temp_file, perms)
     endif
 endfunction
 
--- a/tests/tempfiles.vader
+++ b/tests/tempfiles.vader
@@ -20,7 +20,7 @@ Execute (Neomake uses temporary file for unsaved buffer):
     let make_info = values(neomake#GetStatus().make_info)[0]
     let temp_file = make_info.tempfiles[0]
     AssertEqual fnamemodify(temp_file, ':h:h'), fnamemodify(tempname(), ':h')
-    AssertEqual getfperm(temp_file), 'rw-r--r--'
+    AssertEqual getfperm(temp_file), 'rw-------'
     NeomakeTestsWaitForFinishedJobs
   endif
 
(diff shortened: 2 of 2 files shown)


autoload/neomake/utils.vim
```diff
@@ -508,6 +508,9 @@ endfunction
 
 function! neomake#utils#write_tempfile(bufnr, temp_file) abort
     call writefile(neomake#utils#get_buffer_lines(a:bufnr), a:temp_file, 'b')
+    if exists('*setfperm')
+        call setfperm(a:temp_file, 'rw-r--r--')
```
blueyed: What about using 600 after all?
Or better: the original file's permissions?

NickeZ: I can change it to 600 if you want to. If we want to use the users umask we have to remove this assert that fails:

https://github.com/neomake/neomake/blob/856b272bc09850835e3cd6c47ab5b91b995bd993/tests/tempfiles.vader#L23

blueyed: Yes, the assertion needs to be adjusted in both cases (when changing your PR).

I think using `getfperm(bufname(+a:bufnr))` with a fallback to `'rw-------'` if it is empty would be good.

blueyed: btw: the check for if setfperm exists is good, it is not available in 7.3.429 (the oldest version being tested).  Therefore you need to consider this in the test.

blueyed: We could run the tests with a specific `umask` I guess, which could go into the Makefile then, but I am not sure if it is really worth it (and would need to handle Windows then also (Windows tests are on hold, but there is some WIP in #1575))

blueyed: @NickeZ 
Are you going to update the PR?
Otherwise I might pick it up.

NickeZ: Feel free to take it if you have decided what to do. I was waiting for your input.

blueyed: Pushed an update, according to my previous comment(s).

NickeZ: The tests succeed on my machine :+1:

blueyed: Thanks, merged.

PR #2303, opened 2019-02-07
use shellcheck JSON output

euclio:


blueyed:
Thanks!
https://github.com/euclio/neomake/blob/329d77437e2a9c2c56facb8028d1af4f5028b00c/tests/integration.vader#L183 fails, but it is also a not so good test by itself.  I will look into fixing it (removing shellcheck there) later.

blueyed:
https://github.com/neomake/neomake/pull/2304 (merged).
Please rebase.

euclio:
Rebased.

autoload/neomake/makers/ft/sh.vim
```diff
@@ -40,6 +36,30 @@ function! neomake#makers#ft#sh#shellcheck() abort
...
+            let type = 'W'
+        else
+            let type = 'E'
+        endif
+
+        call add(errors, {
+            \ 'text': err.message,
+            \ 'lnum': err.line,
+            \ 'col': err.column,
+            \ 'length': err.endColumn - err.column,
```
blueyed: This needs `+1`.

autoload/neomake/makers/ft/sh.vim
```diff
@@ -40,6 +36,30 @@ function! neomake#makers#ft#sh#shellcheck() abort
...
+            \ 'filename': err.file,
```
blueyed: The `filename` part does not work with temporary files.
Likely setting/using `bufnr` instead would work / be better.

blueyed:
Given the above issues: what is the benefit of using the json output?  Only having the length easily?

euclio:
Yeah, being able to calculate the length is the main reason.

blueyed:
Re `endColumn` support in general: https://github.com/koalaman/shellcheck/issues/966

blueyed:
And regarding tab size: https://github.com/koalaman/shellcheck/issues/1048

blueyed:
Left a comment at https://github.com/koalaman/shellcheck/issues/1048#issuecomment-461550298.

autoload/neomake/makers/ft/sh.vim
blueyed: This should be fixed via https://github.com/neomake/neomake/pull/2305.

euclio:
Since endColumn support is on its way out, I think the downsides of this outweigh the benefits. Closing.

blueyed:
Agreed.

> Since endColumn support is on its way out

Do you have a source?

euclio:
https://github.com/koalaman/shellcheck/issues/966, but I didn't realize that it was from over a year ago. I guess it's no longer accurate.

PR #2316, opened 2019-02-22
Avoid using same color as Normal bg for signs

sebastianvera:
Hello, I'm using this [vim-one](https://github.com/rakr/vim-one) colorscheme and the error sign is using the same background as color, so maybe we can do the same as for gotham colorscheme:

Before:

before

After:

after

I guess this same problem can happen with other colorschemes so this should be a "generic solution", but I'm not sure if maybe I'm missing something since I'm not familiarized with vim script.

Thanks!

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -256,7 +256,13 @@ endfunction
 
 function! neomake#signs#DefineHighlights() abort
     let ctermbg = neomake#utils#GetHighlight('SignColumn', 'bg')
+    if ctermbg ==# 'NONE'
+        let ctermbg =  neomake#utils#GetHighlight('Normal', 'bg')
+    endif
     let guibg = neomake#utils#GetHighlight('SignColumn', 'bg#')
+    if guibg ==# 'NONE'
+        let guibg = neomake#utils#GetHighlight('Normal', 'bg#')
+    endif
     let bg = 'ctermbg='.ctermbg.' guibg='.guibg
 
     for [group, fg_from] in items({


autoload/neomake/signs.vim
```diff
@@ -269,8 +270,9 @@ function! neomake#signs#DefineHighlights() abort
         let ctermfg = neomake#utils#GetHighlight(fg_group, fg_attr)
         let guifg = neomake#utils#GetHighlight(fg_group, fg_attr.'#')
         " Ensure that we're not using SignColumn bg as fg (as with gotham
-        " colorscheme, issue https://github.com/neomake/neomake/pull/659).
-        if ctermfg == ctermbg && guifg == guibg
+        " colorscheme, issue https://github.com/neomake/neomake/pull/659)
+        " or that the sign color is different that the Normal bg.
+        if (ctermfg == ctermbg && guifg == guibg) || ctermfg == normalctermbg
```
blueyed: Not really sure, but it should maybe also check `&& guifg == normalguibg` (with `let normalguibg = neomake#utils#GetHighlight('Normal', 'bg#')` ?
Is that the case with your scheme?

blueyed:
```suggestion
        if (ctermfg == ctermbg && guifg == guibg) || (ctermfg == normalctermbg && guifg == normalguibg)
```

autoload/neomake/signs.vim
```diff
@@ -255,6 +255,7 @@ function! neomake#signs#RedefineInfoSign(...) abort
 endfunction
 
 function! neomake#signs#DefineHighlights() abort
+    let normalctermbg = neomake#utils#GetHighlight('Normal', 'bg')
```
blueyed:
```suggestion
    let normalctermbg = neomake#utils#GetHighlight('Normal', 'bg')
    let normalguibg = neomake#utils#GetHighlight('Normal', 'bg#')
```

blueyed:
I think a more generic check would be to check if it is the same as the `SignColumn`?

blueyed:
Which is what the `ctermfg == ctermbg` is checking for.
Maybe only the `&& guifg == guibg` is not matching / as expected for you?

blueyed:
Just tried the theme myself, and it looks as follows for me:
2019-02-23-080835_612x521_escrotum
2019-02-23-080947_611x516_escrotum

blueyed:
For me the existing condition triggers it to get reversed already:

```
:debug call neomake#signs#DefineHighlights()
…
function neomake#signs#DefineHighlights                                                                   
line 14: let fg_attr = neomake#utils#ReverseSynIDattr(fg_attr)                                            
>echo ctermfg                                                                                             
255                                                                                                       
>echo ctermbg                                                                                             
255                                                                                                       
>echo guifg                                                                                               
#fafafa                                                                                                   
>echo guibg                                                                                               
#fafafa                                                                                                   
>echo fg_group                                                                                            
Error
```

You can just add some `echom string([fg_group, ctermfg, ctermbg, guifg, guibg])` yourself to the source, or use the `:debug` method (with `s` to step into and `n` for next).

blueyed:
What is the following for you?
```
:hi Normal                                                                                                
Normal         xxx ctermfg=23 ctermbg=255 guifg=#494b53 guibg=#fafafa
:hi Error                                                                                                 
Error          xxx cterm=bold ctermfg=166 ctermbg=255 gui=bold guifg=#e45649 guibg=#fafafa
```

blueyed:
I've tested this with Kitty terminal btw, where I've disabled my existing customization for solarized.

sebastianvera:
This is what I get:

```
> echo string([fg_group, ctermfg, ctermbg, guifg, guibg])
['Error', '16', 'NONE', '#282c34', 'NONE']
```

```
:hi Normal
Normal         xxx ctermfg=145 ctermbg=16 guifg=#abb2bf guibg=#282c34
:hi Error
Error          xxx cterm=bold ctermfg=168 ctermbg=16 gui=bold guifg=#e06c75 guibg=#282c34
```

I'm using iTerm2 with [snazzy](https://github.com/sindresorhus/iterm2-snazzy) theme, not sure if that's helpful

sebastianvera:
> Thanks for the detailed issue and PR!
> 
> Is it only a problem with `NeomakeErrorSign` for you?

Yes, it is only a problem with the `NeomakeErrorSign`.

> Which is what the `ctermfg == ctermbg` is checking for.
> Maybe only the `&& guifg == guibg` is not matching / as expected for you?

Yes, this is the problem, they are different:
- guifg: `#282c34`
- guibg:  `NONE`

sebastianvera:
Or maybe we can also do something like:

```vim
let ctermbg = neomake#utils#GetHighlight('SignColumn', 'bg')
let ctermbg = ctermbg ==# 'NONE' ? neomake#utils#GetHighlight('Normal', 'bg') : ctermbg
let guibg = neomake#utils#GetHighlight('SignColumn', 'bg#')
let guibg = guibg ==# 'NONE' ? neomake#utils#GetHighlight('Normal', 'bg#') : guibg
let bg = 'ctermbg='.ctermbg.' guibg='.guibg
...
    if ctermfg == ctermbg && guifg == guibg
    ...
```

and then we leave the `if` intact

sebastianvera:
@blueyed not sure which approach would you prefer, but I look forward to your response!

blueyed:
@sebastianvera 
Thanks for all the details and suggestions, will look over them later.
I wonder why you get different values with the same (Vim) theme though - are you using something like base16 also (which maps colors to the first 16/21 terminal colors)?

blueyed:
The current patch looks useful/good in general though.

sebastianvera:
> @sebastianvera
> Thanks for all the details and suggestions, will look over them later.
> I wonder why you get different values with the same (Vim) theme though - are you using something like base16 also (which maps colors to the first 16/21 terminal colors)?

No @blueyed , I'm not using `base16`, but I noticed that `airblade/vim-gitgutter` is the problem:


```vim
call plug#begin('~/.config/nvim/plugged')
Plug 'airblade/vim-gitgutter'
Plug 'rakr/vim-one'
Plug 'neomake/neomake'
call plug#end()

filetype plugin indent on
syntax enable

colorscheme one
set bg=dark
```

With that you should be able to reproduce getting `neomake#utils#GetHighlight('SignColumn', 'bg')` as `NONE`

blueyed:
Yes, I could reproduce it now.

Please try https://github.com/neomake/neomake/pull/2322.

sebastianvera:
@blueyed now it doesn't work even if I deactivate `airblade/vim-gitgutter` 🙁

blueyed:
Also with the first commit already?  https://github.com/neomake/neomake/pull/2322/commits/d23e4036349bef612910c96c552b324445bb6617

sebastianvera:
@blueyed let me try, but the problem is that:

`guifg` is not changing from `#282c34` to `#e06c75`

Now I'll try the first commit

sebastianvera:
Yes @blueyed, d23e403 works

blueyed:
The second the also I suppose, but only 1404e3d fails then.

I've just tried it myself, and it appears to work for me.

I am getting this:
```
:hi NeomakeErrorSignDefault | hi SignColumn | hi Normal
NeomakeErrorSignDefault xxx ctermfg=168 ctermbg=16 guifg=#282c34 guibg=#282c34
SignColumn     xxx term=standout ctermfg=14 ctermbg=16 guifg=Cyan guibg=#282c34
                   links to LineNr
Normal         xxx ctermfg=145 ctermbg=16 guifg=#abb2bf guibg=#282c34
```


> `guifg` is not changing from `#282c34` to `#e06c75`

For NeomakeErrorSignDefault?

blueyed:
Oh, I have the same as you probably: `NeomakeErrorSignDefault xxx ctermfg=168 ctermbg=16 guifg=#282c34 guibg=#282c34`.

You are using `set termguicolors`?

blueyed:
Can reproduce, will fix it later.

sebastianvera:
@blueyed yes! I have `set termguicolors` 🙂

blueyed:
Pushed a fix.. a var was replaced twice..

sebastianvera:
now works like a charm, I thought it had to do with that but I couldn't check. Thank you very much @blueyed for your time and for this amazing plugin :)

blueyed:
> I have `set termguicolors`

Not in the minimal vimrc you've posted though.. ;)

sebastianvera:
Yes, I just used the minimal to reproduce the error, I’m sorry!

blueyed:
Merged https://github.com/neomake/neomake/pull/2322.

Thanks for all your valuable input, and the suggested patches!

PR #2334, opened 2019-03-08
hide [no test files] line for go maker

cenkalti:
When there are no test files, go command prints an output line similar to:
```
?   	github.com/cenkalti/rain/rainrpc	[no test files]
```

This line does not need to be visible on location list. This PR makes a change to hide that line.

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -16,6 +16,8 @@ function! neomake#makers#ft#go#go() abort
             \ 'test', '-c',
             \ '-o', g:neomake#compat#dev_null,
         \ ],
+        \ 'output_stream': 'stdout',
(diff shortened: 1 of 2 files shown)


blueyed:
Not sure why the test fails, but it should be modernized anyway: you can cherry-pick / use https://github.com/neomake/neomake/pull/2335 already, and then adjust the output for what you are testing maybe (or add new test base on this).

cenkalti:
This is done I guess.

blueyed:
ok.. I've renamed the test - or do I miss something?
can you fix the conflict now, please?
(move it to the other "go: go" test)

blueyed:
Sorry for noticing it only now, but this should rather go into the errorformat, using `%-G…`.

cenkalti:
I didn't know that `%-G` existed for this purpose. I guess it can be merged now.

blueyed:
Thanks!

PR #2339, opened 2019-03-21
Add eslint to typescript markers

topaxi:
eslint can now be used for typescript aswell, see: https://github.com/typescript-eslint/typescript-eslint

### DIFF
--- a/autoload/neomake/makers/ft/typescript.vim
+++ b/autoload/neomake/makers/ft/typescript.vim
@@ -39,3 +39,7 @@ function! neomake#makers#ft#typescript#tslint() abort
     endif
     return maker
 endfunction
+
+function! neomake#makers#ft#typescript#eslint() abort
+    return neomake#makers#ft#javascript#eslint()
+endfunction


autoload/neomake/makers/ft/typescript.vim
```diff
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#typescript#EnabledMakers() abort
-    return ['tsc', 'tslint']
+    return ['tsc', 'tslint', 'eslint']
```
blueyed: How does it behave if you have eslint installed, but not https://github.com/typescript-eslint/typescript-eslint?
(I wonder if it should be enabled by default).

topaxi: Good question, it would fail end probably cause an error. I removed eslint from the default makers.

blueyed:
Ok.

Please add it to the wiki (https://github.com/neomake/neomake/wiki/Makers), with a link to https://github.com/typescript-eslint/typescript-eslint I guess.

topaxi:
Done :+1: Thanks for the fast response <3

PR #2341, opened 2019-03-27
php: Correctly display psalm error type

sica07:
Before this change, all psalm-generated messages where of "Error" type.
Now the error type is correctly determined/parsed from the output message.

### DIFF
--- a/autoload/neomake/makers/ft/php.vim
+++ b/autoload/neomake/makers/ft/php.vim
@@ -66,7 +66,7 @@ function! neomake#makers#ft#php#psalm() abort
         \ 'args': [
             \ '--output-format=pylint'
         \ ],
-        \ 'errorformat': '%E%f:%l:%m',
+        \ 'errorformat': '%A%f:%l:%\s[%t%n]%\s%m',
         \ }
     return maker
 endfunction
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks!

Do you have some example output, which covered different types?
I would like to have a test for this then.

sica07:
Sure! Here they are:
```
app/Services/QuestionnaireService.php:85: [E0001] MismatchingDocblockParamType: Parameter $questionnaire has wrong type 'App\Services\Service', should be 'App\Service' (column 15)
app/Services/QuestionnaireService.php:179: [W0001] RedundantCondition: Found a redundant condition when evaluating $existingQuestion and trying to reconcile it with a non-falsy assertion (column 44)
```
The result (in vim) is: 
```
app/Services/QuestionnaireService.php|85 error   1| MismatchingDocblockParamType: Parameter $questionnaire has wrong type 'App\Services\Service', should be 'App\Service' (column 15)
app/Services/QuestionnaireService.php|179 warning   1| RedundantCondition: Found a redundant condition when evaluating $existingQuestion and trying to reconcile it with a non-falsy assertion (column 44)
```

blueyed:
Thanks!

PR #2353, opened 2019-05-09
[WIP] virtualtext improvements

blueyed:
1. Factor out neomake#utils#define_derived_highlight
2. improve default highlight groups

### DIFF
--- a/autoload/neomake/signs.vim
+++ b/autoload/neomake/signs.vim
@@ -258,32 +258,9 @@ function! neomake#signs#DefineHighlights() abort
     " Use background from SignColumn.
     let ctermbg = neomake#utils#GetHighlight('SignColumn', 'bg', 'Normal')
     let guibg = neomake#utils#GetHighlight('SignColumn', 'bg#', 'Normal')
-    let bg = 'ctermbg='.ctermbg.' guibg='.guibg
 
-    for [group, fg_from] in items({
-                \ 'NeomakeErrorSign': ['Error', 'bg'],
-                \ 'NeomakeWarningSign': ['Todo', 'fg'],
-                \ 'NeomakeInfoSign': ['Question', 'fg'],
-                \ 'NeomakeMessageSign': ['ModeMsg', 'fg']
-                \ })
-        let [fg_group, fg_attr] = fg_from
-        " NOTE: fg falls back to "Normal" always, not "SignColumn" inbetween.
-        let ctermfg = neomake#utils#GetHighlight(fg_group, fg_attr, 'Normal')
-        let guifg = neomake#utils#GetHighlight(fg_group, fg_attr.'#', 'Normal')
-
-        " Ensure that we're not using SignColumn bg as fg (as with gotham
-        " colorscheme, issue https://github.com/neomake/neomake/pull/659).
-        if ctermfg !=# 'NONE' && ctermfg ==# ctermbg
-            let ctermfg = neomake#utils#GetHighlight(fg_group, neomake#utils#ReverseSynIDattr(fg_attr))
-        endif
-        if guifg !=# 'NONE' && guifg ==# guibg
-            let guifg = neomake#utils#GetHighlight(fg_group, neomake#utils#ReverseSynIDattr(fg_attr).'#')
-        endif
-        exe 'hi '.group.'Default ctermfg='.ctermfg.' guifg='.guifg.' '.bg
-        if !neomake#utils#highlight_is_defined(group)
-            exe 'hi link '.group.' '.group.'Default'
-        endif
-    endfor
+    " Define NeomakeErrorSign, NeomakeWarningSign etc.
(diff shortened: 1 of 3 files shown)


petobens:
Hi! Is it possible to override this hl groups? Now my virtualtext looks like this (which is barely noticeable):
image

blueyed:
Yes, you can define the highlight groups, then it will not be overridden (IIRC it is even in the docs, but could be seen from the diff).
How did it look before ?  - I wonder since we should improve the defaults then maybe.

I assume signs are not affected, since you have contrast through SignColumn there then?

petobens:
Ohh indeed adding the hl groups seems to do the trick. Sorry for the noise

Before the fg text was red if I recall correctly.

blueyed:
Please post the output of `:hi …` for the group before and after.
I think we should do some detection here to ensure that it is not the `Normal` (nor CursorLine) bg maybe.

blueyed:
Does this help?

```diff
diff --git i/autoload/neomake/virtualtext.vim w/autoload/neomake/virtualtext.vim
index b97f2d70d..1b8124900 100644
--- i/autoload/neomake/virtualtext.vim
+++ w/autoload/neomake/virtualtext.vim
@@ -104,6 +104,10 @@ else
 endif
 
 function! neomake#virtualtext#DefineHighlights() abort
+    " Use background from SignColumn.
+    let ctermbg = neomake#utils#GetHighlight('CursorLine', 'bg', 'Normal')
+    let guibg = neomake#utils#GetHighlight('CursorLine', 'bg#', 'Normal')
+
     " NOTE: linking to SpellBad etc is bad/distracting (with undercurl).
     call neomake#utils#define_derived_highlights('NeomakeVirtualtext%s', ['NONE', 'NONE'])
 endfunction
```

petobens:
Before the change:
```
:hi NeomakeVirtualtextError
NeomakeVirtualtextError xxx links to NeomakeVirtualtextErrorDefault

:hi NeomakeVirtualtextErrorDefault
NeomakeVirtualtextErrorDefault xxx ctermfg=16 guifg=#24272e
```
after my change:
```
:hi NeomakeVirtualtextError
NeomakeVirtualtextError xxx ctermfg=168 guifg=#e06c75
```
I believe the problem indeed happens because I'm no setting a bg color for the Normal hl group: https://github.com/petobens/colorish/blob/6e2b67465f3e62d54f0e64af7bf4a3d618a6adc4/colors/onedarkish.vim#L300-L301

blueyed:
> because I'm no setting a bg color for the Normal hl group

I think that's normal.

> Again sorry for the noise

No, it's fine.
Can you try the patch from above without your manual setting?
It would be good to have better defaults in general.

petobens:
I tried with:
```diff
 function! neomake#virtualtext#DefineHighlights() abort
+    " Uses background from CursorLine.
+    let ctermbg = neomake#utils#GetHighlight('CursorLine', 'bg', 'Normal')
+    let guibg = neomake#utils#GetHighlight('CursorLine', 'bg#', 'Normal')
     " NOTE: linking to SpellBad etc is bad/distracting (with undercurl).
-    call neomake#utils#define_derived_highlights('NeomakeVirtualtext%s', ['NONE', 'NONE'])
+    call neomake#utils#define_derived_highlights('NeomakeVirtualtext%s', [ctermbg, guibg])
 endfunction
```
and it's indeed slightly better (but the foreground colors are still not good enough)

blueyed:
Can you also try with the following in the patch:
```
    let ctermbg = neomake#utils#GetHighlight('Normal', 'bg')
    let guibg = neomake#utils#GetHighlight('Normal', 'bg#')
```
(i.e. make it check only "Normal", not "CursorLine")
In the end it might make sense to check both there.

(But it fails short anyway if highlights are only a bit different)

Before this patch it was using just `Error`, while now it is using the `bg` of `Error` - does this match what you are seeing?  (check `:hi Error`)

What colorscheme are you using?

petobens:
```viml
    let ctermbg = neomake#utils#GetHighlight('Normal', 'bg')
    let guibg = neomake#utils#GetHighlight('Normal', 'bg#')
```
seems to do the trick:
```
:hi NeomakeVirtualtextError
NeomakeVirtualtextError xxx ctermfg=168 guifg=#e06c75
```
I'm using my own colorscheme onedarkish which you can find here: https://github.com/petobens/colorish

blueyed:
@petobens 
Looked into this a bit: it uses bg from Error by default, but swaps it for you for the signs, since it has the same bg there - but not so when setting it up for virtualtext.

I've noticed that you defined it explicitly by now (got me confused at first when using master ;)), so this is mainly just for info.

But I've started work on improving it in https://github.com/neomake/neomake/pull/2367, and it appears to work well there with your colorscheme also (before the explicit config therein).

Let me know what you think - would be good if you could try it, and provide feedback there.

PR #2355, opened 2019-05-10
fix echo messages too long when there are double width characters

lilydjwg:
#895 fixed that a character was cut in the middle, but it may still be too long.

### DIFF
--- a/autoload/neomake/utils.vim
+++ b/autoload/neomake/utils.vim
@@ -35,7 +35,7 @@ endfunction
 function! neomake#utils#truncate_width(string, width) abort
     let pos = a:width
     while 1
-        let s = matchstr(a:string, '.\{,' . (pos-1) . '}', 0, 1)
+        let s = matchstr(a:string, '.\{,'.string(pos-1).'}', 0, 1)
         let w = strwidth(s)
         if w <= a:width
             return s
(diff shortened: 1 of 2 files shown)


lilydjwg:
I've removed the `start` argument because it's not used. I'm also fixing test errors.

blueyed:
Thanks for this!

At first I'd rather wanted to keep the old function name and signature, but I guess we could bring it back later if needed.

I've pushed a small update: let me know if that looks good to you - it uses `😄` in the test for better visibility.

lilydjwg:
Thanks! The commits look good. (I tried with an emoji but I got a single-width one and then gave up.)

For the function name, a wide-char version of`strpart` basically is for tearing strings apart by character (Vim 8 has a `strcharpart` for this). What we need here is truncate a string to a given display width. These don't sound the same thing and a different name will avoid future confusion. (There are other aspects that this code doesn't deal with, e.g. combining characters, zero-width ones, etc) However I do think I've come up with a poor name. At least I forgot to put `str` somewhere.

BTW, how did you push commits to this pr? You pushed to `refs/pull/2355`?

blueyed:
> BTW, how did you push commits to this pr? You pushed to refs/pull/2355?

Used https://github.com/github/hub: `hub remote add -p lilydjwg`, then fetched, checked out, and pushed there.  But using `refs/pull` also should work, yeah.  You can (dis)allow this for PRs (checkbox on the right).

I think we can live with the function name for now.

Thanks again!

PR #2359, opened 2019-05-18
signs: change default for warning sign to "‼"

blueyed:
'DOUBLE EXCLAMATION MARK' (U+203C) is typically more visible than
'WARNING SIGN' (U+26A0), which would require the emoji variation
selector (U+fe0f) to render it in width 2, but that makes it wider than
the other defaults, and is not handled well in Neovim/Vim.

Ref: https://github.com/kovidgoyal/kitty/issues/1612
Ref: https://github.com/neovim/neovim/issues/7151


blueyed:
Another alternative might be "⚡", open for other suggestions, of course.

PR #2361, opened 2019-05-18
highlights: derive from ErrorMsg/WarningMsg instead of Error/Todo

blueyed:
`Error` and `Todo` are not defined by default actually.


blueyed:
> `Error` and `Todo` are not defined by default actually.

They are defined with/after `:syn on` - I've only checked with `vim -Nu NONE`.

PR #2367, opened 2019-05-19
WIP: improve heuristics for default highlight foreground

blueyed:


petobens:
I tried this with my colorscheme (https://github.com/petobens/colorish) without explicitly overrriding the hl groups and with the following `foo.py` file
```py
x = 2
i = y / x
z = i
print(z)
```
I get kind of proper hl for the first warning (it should probably be orange instead of red) but no for the following ones:
image

image

If I add this hls to my colorscheme:
```viml
call <sid>HL('NeomakeVirtualtextError', s:red, s:cursor_grey, '')
call <sid>HL('NeomakeVirtualtextWarning', s:dark_yellow, s:cursor_grey, '')
call <sid>HL('NeomakeVirtualtextInfo', s:red, s:cursor_grey, '')
call <sid>HL('NeomakeVirtualtextMessage', s:red, s:cursor_grey, '')
```
then I get:
image
image

petobens:
Is is possible to change the `g:neomake_virtualtext_prefix` sign conditional on if it's an error or a warning?

blueyed:
> Is is possible to change the `g:neomake_virtualtext_prefix` sign conditional on if it's an error or a warning?

No.  But there's a TODO / idea already for having the whole property / text being used there to be dynamic, which then could also provide the prefix.  I.e. a maker would return a special property that would be used there - or provide a callback to get the text for a given error (entry).  For this to work properly I'd like to first improve the internal list/error handling then though.

Thanks for trying / providing feedback already!
I will try your example later - for now you could look at / use `tests/vim/t-types.vim`, which makes it easier to compare different types, use e.g. `vim tests/vim/t-types.vim` and then `:source %`.

PR #2368, opened 2019-05-19
tests: do not use --embed with nvim in RPC based test

blueyed:
Ref: https://github.com/neovim/neovim/issues/10031#issuecomment-493755631

With `--embed` it would hang with output during autoload, likely waiting
for an hit-enter prompt (e.g. with `for i in range(0, 10) | echom i | endfor`
in `autoload/neomake/highlights.vim`).

TODO:

- [x] does not work with nvim-0.1.7 - should be done conditionally then, based on the Neovim version where this started working.

### DIFF
--- a/tests/compat.vader
+++ b/tests/compat.vader
@@ -78,7 +78,15 @@ Execute (neomake#compat#get_mode):
 
   if has('nvim')
     let nvim_exe = '/proc/'.getpid().'/exe'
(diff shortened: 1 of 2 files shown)


blueyed:
Oh well, that does not work on Nvim 0.1.7 and 0.3.1 apparently..

Issue #1919, opened 2018-04-07
The cargo maker doesn't mark warnings as warnings


MaikKlein:
### Expected behavior

The `cargo` marker doesn't mark warnings as warnings.
![](https://i.imgur.com/LGIHW6D.png)

Because of this `let g:neomake_warning_sign` won't be used.
### Steps to reproduce

[log file](https://gist.github.com/MaikKlein/0e2b3b833f15b8593b9d5f768911d7f8)

#### Neomake debug information

Async support: 1
Current filetype: rust
Windows: 0
[shell, shellcmdflag, shellslash]:['/usr/bin/fish', '-c', 0]

##### Enabled makers

For the current filetype ("rust", used with :Neomake):
 - cargo
   - append_file: 0
   - args: ['check', '--message-format=json', '--quiet']
   - auto_enabled: 1
   - cwd: '/home/maik/projects/sm'
   - exe: 'cargo'
   - process_output: function('neomake#makers#ft#rust#CargoProcessOutput')
NOTE: you can define g:neomake_rust_enabled_makers to configure it (or b:neomake_rust_enabled_makers).

Default maker settings:
 - output_stream: 'both'
 - remove_invalid_entries: 0
 - buffer_output: 1

For the project (used with :Neomake!):
None.
NOTE: you can define g:neomake_enabled_makers to configure it.

###### New-style (dict, overrides old-style)

```
g:neomake: {
  'automake': {
    'ignore_filetypes': ['startify'],
    'events': {
      'BufWritePost': {
        'delay': 0,
      },
    },
  },
}
b:neomake: unset
```

###### Old-style

```
g:neomake_error_sign = {'texthl': 'NeomakeErrorMsg', 'text': '▶'}
g:neomake_highlight_columns = 0
g:neomake_info_sign = {'texthl': 'NeomakeWarningMsg', 'text': 'W'}
g:neomake_logfile = '/tmp/neomake.log'
g:neomake_message_sign = {'texthl': 'NeomakeWarningMsg', 'text': 'W'}
g:neomake_place_signs = 1
g:neomake_warning_sign = {'texthl': 'NeomakeWarningMsg', 'text': 'W'}

```
  makeprg=make

#### :version

```
NVIM v0.2.2
Build type: Release
LuaJIT 2.0.5
[...]

PR #2392, opened 2019-06-15
Fix bug: rust mark warnings as errors

Luffbee:
Fix: #1919


blueyed:
Thanks for the PR - have not looked into it yet, but also created one: https://github.com/neomake/neomake/pull/2393

What's better? :)

blueyed:
Thanks for the PR again in the first place!
Let's continue it in https://github.com/neomake/neomake/pull/2393.

PR #2393, opened 2019-06-15
rust: cargo: fix type for entries with code

blueyed:
Also use str2nr to have "nr" 61 for "0061", and not 49.

Fixes https://github.com/neomake/neomake/issues/1919.

### DIFF
--- a/autoload/neomake/makers/ft/rust.vim
+++ b/autoload/neomake/makers/ft/rust.vim
@@ -130,7 +130,8 @@ function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
 
         let decoded = neomake#compat#json_decode(line)
         let data = get(decoded, 'message', -1)
-        if type(data) != type({}) || empty(data['spans'])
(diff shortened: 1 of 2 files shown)


autoload/neomake/makers/ft/rust.vim
```diff
@@ -135,16 +135,19 @@ function! neomake#makers#ft#rust#CargoProcessOutput(context) abort
...
+                    \ && index(['E', 'W'], code_dict['code'][0]) != -1
+            let error.type = code_dict['code'][0]
+            let error.nr = str2nr(code_dict['code'][1:])
+        else
+            let level = get(data, 'level', -1)
+            if level != -1
+                let error.type = toupper(level[0])
             else
-                let error.type = 'E'
+                let error.type = 'W'
```
blueyed: This fallback is not covered with tests.
Is it a valid use case?  (i.e. is there `"code": null`, without `"level"`?)

Issue #2340, opened 2019-03-25
The neomake eslint experience


cideM:
## The issue 
This post is meant as constructive criticism, because the out of the box experience for the uninitiated JS developer is pretty rough. Keep in mind that ALE just works, and so does Visual Studio Code with the eslint plugin.

1. Install `neomake` by e.g., adding it to the plugin list.,
2. Open `.js` file and run `:Neomake` -> Nothing happens. This is going to be the first hurdle since I assume people will have to first figure out that they can run `:NeomakeInfo`. But this is also out of scope for the eslint specific isssues, so moving on.
3.  `:NeomakeInfo` shows
```
   - WARNINGS:
     - maker's exe (eslint) is not executable.
```
4. Somehow the users now needs to understand that `neomake` assumes linters to be available globally because in many (most?) languages that's how it works. But we absolutely need the local eslint because a global eslint is pretty much useless (plugin version mismatch).
5. The https://github.com/neomake/neomake/issues/247 issue is lengthy with vastly different solutions. Let's assume that the user does something like the below so things just work for now.
```
let b:eslint_exe = substitute(system('npm bin') ,'\n', '', 'g') . '/eslint'
 
let b:neomake_eslint_exe = b:eslint_exe
```

6. Open `.js` file again, run `:Neomake` and see too many errors. The `cwd` is now set to the folder of the current file and at least in our codebase at work this breaks absolute imports like `import Foo from 'components/Foo'` which use the webpack and eslint resolve plugin and which works in ALE and VSC.
7. Empowered by the prior success of setting `exe` the user might now try to do the same with `cwd` but as far as I can see from `:h neomake` that's not actually possible because it's not part of the maker properties that can be set. 
8. Try to just set the entire maker as an object, realize that you now need to copy & paste the error format and args from the maker source file from the neomake repo. (which does work)

## What could help

I do understand that not everyone is a JS dev and that eslint is a bit of a special snowflake. But I really like neomake and I'd love to make it easier for JS devs. What would help is a `FAQ` entry which outlines the issue (neomake uses global binaries) and proposes a solution (with a link to an issue where people can find more help) which ideally doesn't require additional vimscript files and functions. **I also think that `cwd` should be easier to override or that the object/record way of creating makers can somehow be merged into the defaults for that maker rather than overriding it entirely.**

Thanks for neomake 🍬 🎊

davidosomething:
[...]

Issue #247, opened 2015-12-10
Switching exec for eslint


ahmedelgabri:
I want to use the local version of eslint `./node_modules/.bin/eslint` instead of the global one and this how my current setup looks like.

```vim
let g:neomake_open_list=2
let g:neomake_javascript_enabled_makers = ['standard']
let g:neomake_jsx_enabled_makers = ['standard']
if findfile('.eslintrc', '.;') !=# ''
  let g:neomake_javascript_enabled_makers = ['eslint']
  let g:neomake_jsx_enabled_makers = ['eslint']
  let g:neomake_javascript_eslint_maker = {
        \ 'exec': $PWD . '/node_modules/.bin/eslint'
        \ }
endif

autocmd! BufWritePost * Neomake
```

But it doesn't look like it's working although if I add `echo($PWD . '/node_modules/.bin/eslint')` the right path gets output. How can I fix this?

Yamakaky:
maybe `neomake_javascript_eslint_maker.bin = $PWD . '/node_modules/.bin/eslint'` or something like that ?

ahmedelgabri:
I don't think `neomake_javascript_eslint_maker.bin` is a valid option?

forbesmyester:
Put this in your .bashrc (or similar) 

$PATH=/node_modules/.bin:$PATH

You'll then realise that you don't ever need to install anything globally, which is a plain silly idea anyway. :smile:

ahmedelgabri:
Any help with this issue?

ahmedelgabri:
Turns out I had a typo! it's `exe` not `exec`, so this works
```vim
  let g:neomake_javascript_eslint_maker = {
  \ 'exe': $PWD .'/node_modules/.bin/eslint'
  \ }
```

But to get _everything_ working you need to do this which I find a bit odd, I'd expect overriding one property to be merged with the other default properties not to override everything.
```vim
  let g:neomake_javascript_eslint_maker = {
  \ 'exe': $PWD .'/node_modules/.bin/eslint',
  \ 'args': ['-f', 'compact'],
  \ 'errorformat': '%E%f: line %l\, col %c\, Error - %m,' .
  \ '%W%f: line %l\, col %c\, Warning - %m'
  \ }
```

benjie:
[...]

PR #2402, opened 2019-07-20
Documentation: add info about handling local over global executables

cideM:
This adds some documentation on an issue that typically arises with ESlint (see #2340 #247). I'm not well versed in documentation syntax for Vim so I mostly tried to stick with what was already there. I'll happily address any and all feedback! :pray:

### DIFF
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -1049,4 +1049,34 @@ Here are some tips to develop the 'errorformat' setting for makers:
    display the raw data.
 5. Pay attention to the "`valid`" property of entries.
 
+                                                     *neomake-faq-local-executables*
+                                                     *neomake-faq-eslint*
+
+6.4 How to prefer local over global executables?~
+
+Some programs such as eslint are typically found in both a global and a local
+location. The global location is usually included in your shell path, whereas
+the path to the locally installed executable for the current project is often
+not included in your path (run `$ echo $PATH` in your shell to see the current
+path)
+
+Neomake itself has no special logic to determine the correct path to any
+executable, meaning if for example eslint is only available globally, that's
+what Neomake will use. This often breaks eslint, since the plugins installed
+in your local project are not also installed globally.
+
+There are several ways of addressing this issue:
+
+- Add `./node_modules/.bin` to your shell path, *before* the entry for the
+  global executables. Keep in mind that running things like `$ eslint init`
+  will not use your global executable when ran in a directory that also has a
+  local executable.
+- Use a plugin such as https://github.com/benjie/local-npm-bin.vim
+- Set the executable path through |neomake-object-maker| or
+  |neomake-makers-properties| For certain configurations this can result in
+  issues with the current working directory (see
+  https://github.com/neomake/neomake/issues/2340#issue-424879342)
+
+See https://github.com/neomake/neomake/issues/247 for even more suggestions.
+
 vim: ft=help tw=78 isk+=<,>,\:,-,'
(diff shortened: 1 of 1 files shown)


doc/neomake.txt
```diff
@@ -1049,4 +1049,34 @@ Here are some tips to develop the 'errorformat' setting for makers:
...
+                                               *neomake-faq-local-executables*
+                                               *neomake-faq-eslint*
+
+6.4 How to prefer local over global executables?~
+
+Some programs such as eslint are typically found in both a global and a local
+location. The global location is usually included in your shell path, whereas
+the path to the locally installed executable for the current project is often
+not included in your path (run `$ echo $PATH` in your shell to see the current
+path)
```
blueyed:
```suggestion
path).
```

doc/neomake.txt
```diff
@@ -1049,4 +1049,34 @@ Here are some tips to develop the 'errorformat' setting for makers:
...
+
+There are several ways of addressing this issue:
+
+- Add `./node_modules/.bin` to your shell path, *before* the entry for the
+  global executables. Keep in mind that running things like `$ eslint init`
+  will not use your global executable when ran in a directory that also has a
+  local executable.
+- Use a plugin such as https://github.com/benjie/local-npm-bin.vim
+- Set the executable path through |neomake-object-maker| or
+  |neomake-makers-properties| For certain configurations this can result in
```
blueyed:
```suggestion
  |neomake-makers-properties|.
  For certain configurations this can result in
```

doc/neomake.txt
```diff
@@ -1049,4 +1049,34 @@ Here are some tips to develop the 'errorformat' setting for makers:
...
+  issues with the current working directory (see
+  https://github.com/neomake/neomake/issues/2340#issue-424879342)
```
blueyed:
```suggestion
  https://github.com/neomake/neomake/issues/2340#issue-424879342).
```

blueyed (approved):

doc/neomake.txt
```diff
@@ -1049,4 +1049,34 @@ Here are some tips to develop the 'errorformat' setting for makers:
...
+path)
+
+Neomake itself has no special logic to determine the correct path to any
+executable, meaning if for example eslint is only available globally, that's
+what Neomake will use. This often breaks eslint, since the plugins installed
+in your local project are not also installed globally.
+
+There are several ways of addressing this issue:
+
+- Add `./node_modules/.bin` to your shell path, *before* the entry for the
```
blueyed: This should use an absolute path likely (since Neomake might change the cwd before running the maker).
(This might have been the problem with your previous comment in https://github.com/neomake/neomake/issues/2340)

cideM: Good point, in my current `config.fish` I'm actually using `$PWD/node_modules/.bin`

blueyed: Yeah.. it should be probably set when `cd`ing to your project (e.g. via https://github.com/Tarrasch/zsh-autoenv).  Also https://github.com/embear/vim-localvimrc/ could be used to set this only in Vim/Neovim then.
Another option might be to use Neomake's hooks (autocommands to do this).

I am not sure if relative `$PATH` entries are meant to work in general.
It could make sense to use `exepath()` to get the full path of the executable before doing any `cd` for the job.  I'll create a new issue to track this.

cideM: I updated the description. I'm not really happy with the `$PATH` part to be honest since it has a lot of caveats, a lot of ways of achieving the goal and it depends on which shell the user uses. I therefore moved this suggestion to the bottom of the list.

blueyed:
Thanks for the update - still have to follow up there.

I just had an idea for a new setting, which would contain a path (or list of paths maybe), that would be looked up for executables if a given exe is not found.
This could then either be a buffer-local or global setting, and it could have defaults for different filetypes, similar to how `neomake#makers#ft#python#project_root_files` defines a list of files to detect Python root projects.
So for ft=javascript it could default to `['{{project_root}}/node_modules/.bin']` then, where `{{project_root}}` could be a placeholder.

blueyed:
See https://github.com/neomake/neomake/pull/2407.

PR #2406, opened 2019-07-22
Add stdin support for RuboCop

buehmann:
I was annoyed by RuboCop complaining about the temporary filename chosen by Neomake. (RuboCop includes a linter for the filenames as well.)

Adding support for stdin fixes this.


blueyed:
Awesome, thanks.

PR #2407, opened 2019-07-24
Support for extra bin dirs

blueyed:
TODO:

- [ ] doc

Ref: https://github.com/neomake/neomake/pull/2402#issuecomment-514455747

### DIFF
--- a/autoload/neomake.vim
+++ b/autoload/neomake.vim
@@ -845,6 +845,19 @@ function! neomake#create_maker_object(maker, ft) abort
                 let maker[key] = setting
             endif
         endfor
+
+        if !executable(maker.exe)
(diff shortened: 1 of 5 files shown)


teto:
I feel this is a bit overkill, can't the user play with its PATH instead ?

blueyed:
@teto 
Thanks for the feedback! :)

It is a common issue/request, and I've thought this might be a good solution to it finally.
While it is additional processing, it should not have an impact when not used, and I think it is good to e.g. have it for JavaScript files by default (like in this PR).
I might eventually add `{{project_root}}/.venv/bin` for Python, but the convention there is not so widespread as with JS.

PR #2413, opened 2019-08-14
Add MacOS support to MyOnBattery()

flowchartsman:
This PR allows the adaptive function `MyOnBattery()` to work on MacOS as well as on Linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,16 @@ to your plugin manager, e.g. after `call plug#end()` with vim-plug.)
 ### Advanced setup
 
 The author likes the following, which uses different modes based on if your
-laptop runs on battery (for Linux):
+laptop runs on battery (for MacOS or Linux):
 
 ```vim
 function! MyOnBattery()
-  return readfile('/sys/class/power_supply/AC/online') == ['0']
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,11 +58,16 @@ to your plugin manager, e.g. after `call plug#end()` with vim-plug.)
...
 
 The author likes the following, which uses different modes based on if your
-laptop runs on battery (for Linux):
+laptop runs on battery (for MacOS or Linux):
 
 ```vim
 function! MyOnBattery()
-  return readfile('/sys/class/power_supply/AC/online') == ['0']
+  if has('macunix')
+    return match(system('pmset -g batt'),"Now drawing from 'Battery Power'") != -1
```
blueyed: Just for reference: I've found https://github.com/imwally/battstat/blob/master/battstat#L38 - which just checks for "Battery Power" (with word boundaries).

blueyed:
```suggestion
    return match(system('pmset -g batt'), "Now drawing from 'Battery Power'") != -1
```

blueyed:
Thanks!

PR #2415, opened 2019-08-14
Updates golangci_lint error formats for lll

JackWink:
The long line linter doesn't include a column in the error output:

```
$ golangci-lint run --disable-all --enable=lll --out-format=line-number --print-issued-lines=false main.go
main.go:216: line is 124 characters (lll)
main.go:217: line is 124 characters (lll)
main.go:220: line is 124 characters (lll)
main.go:235: line is 122 characters (lll)
main.go:237: line is 134 characters (lll)
main.go:238: line is 128 characters (lll)
main.go:269: line is 122 characters (lll)
main.go:420: line is 148 characters (lll)
```
This commit adds an errorformat that matches lint output missing column information.

### DIFF
--- a/autoload/neomake/makers/ft/go.vim
+++ b/autoload/neomake/makers/ft/go.vim
@@ -79,6 +79,7 @@ function! neomake#makers#ft#go#golangci_lint() abort
         \ 'append_file': 0,
         \ 'cwd': '%:h',
         \ 'errorformat':
-            \ '%f:%l:%c: %m'
+            \ '%f:%l:%c: %m,' .
+            \ '%f:%l: %m'
         \ }
 endfunction


JackWink:
Yes and no, all the columns are marked as 0 in the JSON output.

```
$  golangci-lint run --disable-all --enable=lll main.go --out-format=json --print-issued-lines=false | jq
{
  "Issues": [
    {
      "FromLinter": "lll",
      "Text": "line is 124 characters",
      "Pos": {
        "Filename": "main.go",
        "Offset": 0,
        "Line": 216,
        "Column": 0
      },
      "SourceLines": [
        "[redacted]"
      ],
      "Replacement": null
    },
    ...
  ],
}
```

I'm not terribly familiar with vim's errorformat parsing, but another potential issue with JSON output is that it's all single-line. And this should probably be filled upstream, but `golangci_lint` doesn't seem to respect the `--print-issued-lines` flag when the output format is JSON.

blueyed:
Thanks for the info.
Handling JSON requires extra processing (cannot use "errorformat" directly with it (usually)).
So it's fine like that for now - especially given there's no extra benefit.

blueyed:
Thanks!

(I've missed that this affects a particular linter only, and that columns are still there otherwise.  All fine!)

PR #2435, opened 2019-10-19
python: mypy: filter success and failure messages

petobens:


petobens:
Using the following minimal vimrc:
```viml
let $DOTVIM = expand('$HOME/.config/nvim')

set runtimepath+=$DOTVIM/bundle/repos/github.com/petobens/neomake
filetype plugin indent on

let g:neomake_open_list = 2
let g:neomake_python_enabled_makers = ['flake8', 'mypy']
```
and running `:Neomake` on a `foo.py` file like:
```python
x=1
```
I get:
```
|| Success: no issues found in 1 source file
foo.py|1 col 2 warning| E225 missing whitespace around operator
```
This pr suppresses the first line.

blueyed:
Thanks, makes sense.
We could also suppress the summary line on failures maybe - but that might be good (informational) in general.  What do you think?

petobens:
Good point. Don't have a strong opinion on it though (they might be informative but also distracting).

petobens:
Actually I believe it makes sense to also suppress them and added the change to do so

blueyed:
Thanks!
Will test this later, and likely also add a test probably.

petobens:
Thank you!

PR #2438, opened 2019-11-01
Add new filetype makers for javascriptreact and typescriptreact

GerardKetuma:
Vim has recently added the filetypes javascriptreact and typescriptreact for .jsx and .tsx files respectively. This will add makers for those filetypes and they are simply copies of the jsx and tsx filetype makers.

Another way to do this will be to make them SupersetOf their jsx and tsx filetypes. But not sure of how the SupersetOf() function works since those filetypes are themselves super sets of another type.

https://github.com/vim/vim/issues/4830
https://github.com/vim/vim/commit/92852cee3fcff1dc6ce12387b234634e73267b22

### DIFF
--- a/autoload/neomake/makers/ft/javascriptreact.vim
+++ b/autoload/neomake/makers/ft/javascriptreact.vim
@@ -1,0 +1,18 @@
+function! neomake#makers#ft#javascriptreact#SupersetOf() abort
+    return 'javascript'
+endfunction
+
+function! neomake#makers#ft#javascriptreact#EnabledMakers() abort
+    return ['jshint', executable('eslint_d') ? 'eslint_d' : 'eslint']
+endfunction
+
+function! neomake#makers#ft#javascriptreact#javascriptreacthint() abort
+    return neomake#makers#ft#javascript#jshint()
+endfunction
+
+function! neomake#makers#ft#javascriptreact#stylelint() abort
+    return neomake#makers#ft#css#stylelint()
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks!
Maybe we can use symlinks?

GerardKetuma:
> Thanks!
> Maybe we can use symlinks?

There are slight modifications in these filetypes, so not exact copies. The function names had to change from jsx to javascriptreact and so forth.
How will the symlinks work in this case. If you can direct me, I should be able to modify this PR?

luxflux:
I'm interested in this as well, is there something I can help out with?

I don't think the symlink approach will work because the functions are renamed as well, or are the names irrelevant as long as the file has been loaded?

blueyed:
@luxflux 
Thanks for the feedback.  Can you try if the PR works for you?
If so I'm happy to merge as-is for now.

luxflux:
> Thanks for the feedback. Can you try if the PR works for you?

I just tried it and the `javascriptreact` works properly. Unfortunately, I don't have a typescript project at hand to test the `typescriptreact` part.

blueyed:
Thanks, please add it to https://github.com/neomake/neomake/wiki/Makers.

luxflux:
I just checked and I don't think we have to add it there as JSX and Typescript is already in the list - this PR essentially just fixes the behaviour for newer (neo)vim versions.

PR #2463, opened 2020-02-06
Add SwiftLint and Swift Maker

jandamm:
This Pull Request adds a new Maker for [SwiftLint](https://github.com/realm/SwiftLint).
It also adds a Maker for compiling a swift script with swift (instead of swiftc).

SwiftLint will automatically be enabled when a `.swiftlint.yml` is found.

`swiftfile` won't be enabled per default. In my own vimrc I'll replace `swiftc` with it though.

Rationale:
Having a `~/.swiftlint.yml` will enable SwiftLint but won't use this configuration but the default instead.
As this is also true for the existing `swiftpm`, I did ignore it.

Possible solutions:
- Replace/restrict `FindGlobFile` with a function which only searches up until the project root.
- Use the found `.swiftlint.yml`*

* This would result in having support for global a SwiftLint config. But as SwiftLint can merge multiple configurations this feature would be lost.

### DIFF
--- a/autoload/neomake/makers/ft/swift.vim
+++ b/autoload/neomake/makers/ft/swift.vim
@@ -1,14 +1,22 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#swift#EnabledMakers() abort
+    let l:list = ['swiftc']
     if !empty(s:get_swiftpm_config())
-        return ['swiftpm']
+        let l:list = ['swiftpm']
     endif
-    return ['swiftc']
+    if s:get_swiftlint_config()
+        let l:list = add(l:list, 'swiftlint')
+    endif
+    return l:list
 endfunction
 
 function! s:get_swiftpm_config() abort
     return neomake#utils#FindGlobFile('Package.swift')
+endfunction
+
+function! s:get_swiftlint_config() abort
+    return executable('swiftlint') && !empty(neomake#utils#FindGlobFile('.swiftlint.yml'))
 endfunction
 
 function! s:get_swiftpm_base_maker() abort
@@ -39,6 +47,20 @@ function! neomake#makers#ft#swift#swiftpmtest() abort
     let maker = s:get_swiftpm_base_maker()
     let maker.args = ['test']
     return maker
+endfunction
+
+function! neomake#makers#ft#swift#swiftlint() abort
+    return {
+        \ 'args': ['lint'],
+        \ 'append_file': 1,
+        \ 'errorformat':
+            \ '%E%f:%l:%c: error: %m,' .
+            \ '%W%f:%l:%c: warning: %m,' .
+            \ '%E%f:%l: error: %m,' .
+            \ '%W%f:%l: warning: %m,' .
+            \ '%Z%\s%#^~%#,' .
+            \ '%-G%.%#',
+        \ }
 endfunction
 
 function! neomake#makers#ft#swift#swiftc() abort
--- a/autoload/neomake/makers/ft/swift.vim
+++ b/autoload/neomake/makers/ft/swift.vim
@@ -49,6 +49,12 @@ function! neomake#makers#ft#swift#swiftpmtest() abort
     return maker
 endfunction
 
+function! neomake#makers#ft#swift#swiftfile() abort
+    let maker = s:get_swiftpm_base_maker()
+    let maker.append_file = 1
+    return maker
+endfunction
+
 function! neomake#makers#ft#swift#swiftlint() abort
     return {
         \ 'args': ['lint'],


autoload/neomake/makers/ft/swift.vim
```diff
@@ -1,16 +1,24 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#swift#EnabledMakers() abort
+    let list = ['swiftc']
```
blueyed: s/list/ret/

autoload/neomake/makers/ft/swift.vim
```diff
@@ -1,16 +1,24 @@
...
+    return list
 endfunction
 
 function! s:get_swiftpm_config() abort
     return neomake#utils#FindGlobFile('Package.swift')
 endfunction
 
+function! s:get_swiftlint_config() abort
+    return executable('swiftlint') && !empty(neomake#utils#FindGlobFile('.swiftlint.yml'))
+endfunction
```
blueyed: Isn't there also such a thing as a global config?

I think it makes sense to not require a local config for linters to be enabled.

autoload/neomake/makers/ft/swift.vim
```diff
@@ -41,6 +49,26 @@ function! neomake#makers#ft#swift#swiftpmtest() abort
...
+        \ 'args': ['lint'],
+        \ 'append_file': 1,
+        \ 'errorformat':
+            \ '%E%f:%l:%c: error: %m,' .
+            \ '%W%f:%l:%c: warning: %m,' .
+            \ '%E%f:%l: error: %m,' .
+            \ '%W%f:%l: warning: %m,' .
+            \ '%Z%\s%#^~%#,' .
+            \ '%-G%.%#',
+        \ }
```
blueyed: This appears to slightly extend (and maybe therefore only fix) the other errorformats being used.

Would be good to have some example output to test this with, and merge it with the base efm eventually.

autoload/neomake/makers/ft/swift.vim
jandamm: SwiftLint officially does not support global configs. It does have a default configuration which I think many wouldn't want to be enabled for every file they open. E.g. it marks every variable shorter than two letters as an error.
The convention is having a config file in your project or provide it via an argument.

I think the best solution would be to search for a config file then run `swiftlint`, check for ~/.swiftlint.yml if it exists run `swiftlintglobal` (with config provided).

blueyed: But wouldn't people not liking swiftlint then just not install it?

autoload/neomake/makers/ft/swift.vim
jandamm:
```
Linting Swift files at paths /Users/jandamm/Formula.swift
Linting 'Formula.swift' (1/1)
/Users/jandamm/Formula.swift:169:15: error: Force Cast Violation: Force casts should be avoided. (force_cast)
/Users/jandamm/Formula.swift:1:15: error: Identifier Name Violation: Variable name should be between 3 and 40 characters long: 'f' (identifier_name)
/Users/jandamm/Formula.swift:78:16: error: Identifier Name Violation: Variable name should be between 3 and 40 characters long: 's' (identifier_name)
/Users/jandamm/Formula.swift:143:1: warning: Identifier Name Violation: Variable name should be between 3 and 40 characters long: 'f1' (identifier_name)
/Users/jandamm/Formula.swift:144:1: warning: Identifier Name Violation: Variable name should be between 3 and 40 characters long: 'f2' (identifier_name)
/Users/jandamm/Formula.swift:91:1: warning: Line Length Violation: Line should be 120 characters or less: currently 133 characters (line_length)
/Users/jandamm/Formula.swift:137:4: warning: Switch and Case Statement Alignment Violation: Case statements should vertically align with their enclosing switch statement. (switch_case_alignment)
/Users/jandamm/Formula.swift:133:1: warning: Vertical Whitespace Violation: Limit vertical whitespace to a single empty line. Currently 2. (vertical_whitespace)
Done linting! Found 8 violations, 3 serious in 1 file.
```

I've seen [here](https://github.com/keith/swift.vim/blob/master/syntax_checkers/swift/swiftlint.vim#L24) that sometimes the columns seem to be missing.
Without the last two lines all lines are shown in.

autoload/neomake/makers/ft/swift.vim
jandamm: I have it installed because some projects I work on use it, but others don't.
I cannot speak for everyone but I wouldn't want the default rules enabled and I'm not sure many are using a global config (there is a pretty old [issue](https://github.com/realm/SwiftLint/issues/606) requesting this, though)

jandamm:
I think SW would match well.
If you're setting up SwiftLint in your IDE (which is recommended) an Error will fail your compilation, so I'm not sure it would be a `Style Error`.

autoload/neomake/makers/ft/swift.vim
jandamm: I've changed the error format. Should be good now 👍

autoload/neomake/makers/ft/swift.vim
jandamm: I've had a look at swift.vim.
There swiftlint is also only enabled when a `.swiftlint.yml` is present or the user requests it.
https://github.com/keith/swift.vim/blob/master/syntax_checkers/swift/swiftlint.vim#L14

What is your opinion? I would leave the current implementation as is. The user can always opt in and run/enable swiftlint for himself.

PR #2474, opened 2020-02-17
Add FAQ entry to help fix makeprg output

JudgeGregg:
- stdout and stderr are processed independently, which results in
errors appearing in the top-level directory rather than their
respective sub-directory

### DIFF
--- a/doc/neomake.txt
+++ b/doc/neomake.txt
@@ -1053,4 +1053,12 @@ Here are some tips to develop the 'errorformat' setting for makers:
    display the raw data.
 5. Pay attention to the "`valid`" property of entries.
 
+6.4 Neomake quickfix displays all errors in the top-level directory~
+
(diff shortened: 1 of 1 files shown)


blueyed:
Thanks for this, I've picked the linting fixes already - sorry for the trouble in that regard, and extra sorry that GitHub appears to have messed up the authorship when merging it (in https://github.com/neomake/neomake/pull/2479).

blueyed:
As for the change itself: it is not really limited to makeprg only, and I'd rather have a fix for this instead of documenting it.
But I also see that it's worth a FAQ given that there is no fix in sight.
For Vim it could be fixed (using a single callback in async mode), but not so easy for Neovim (https://github.com/neovim/neovim/issues/8823).

I'll think about it a bit and might take your patch, slightly amened then maybe in the end for now.

Thanks again already.

PR #2483, opened 2020-04-24
Add pyflakes errorformat for 2.2.0 output

karlch:
In pyflakes version 2.2.0 we get error messages in the form of
```
test.py:4:11 undefined name 'b'
```
This was not matched to the existing patterns as there is no trailing colon after the column output.

I am not sure if there is a reason for having the format with the existing colon, i.e. the pattern `'%E%f:%l:%c: %m,'`. If not, I could also remove this.

Thanks for this great tool I use all the time!


blueyed:
Thanks for the patch/PR.
I've gone ahead and fixed it finally, adding a comment.

PR #2484, opened 2020-04-28
vue: Use eslint-plugin-vue instead of eslint-plugin-html

yymm:
There is an vue official eslint plugin ( https://github.com/vuejs/eslint-plugin-vue ) , so I changed to use it.


blueyed:
Thank you.
Do you have a reference to what the previously used "html" plugin was/is?

wsdjeg:
it is  [eslint-plugin-html](https://github.com/BenoitZugmeyer/eslint-plugin-html)

Issue #2302, opened 2019-02-07
Pick up changes in g:python3_host_prog


teto:
### Expected behavior

Take into account changes in g:python3_host_prog after starting nvim.

I run neovim on nixos, which generates a specific config with a specific python interpreter g:python3_host_prog=/nix/store/3vwrd209k57ii4mn7rbc45ybkh9rsg8d-neovim-float/bin/nvim-python3.
When I start nvim, it gets this value loads neomake which picks up the previous value of g:python3_host_prog.
Now when I enter a nix-shell with a specific python project, I write a .nvimrc which contains let `g:python3_host_prog="python3"` so that neovim uses the python environment of this specific project (because it has the correct python libraries loaded). This is picked up by neovim (the new value of g:python3_host_prog appears in :checkhealth) but it doesn't seem to be picked up by neomake as
NeomakeInfo still returns information fetch  with the initial value of g:python3_host_prog (/nix/store/3vwrd209k57ii4mn7rbc45ybkh9rsg8d-neovim-float/bin/nvim-python3) : 
```
##### Enabled makers

For the current filetype ("python", used with :Neomake):
 - mypy
   - args: ['-m', 'mypy']
   - auto_enabled: 0
   - errorformat: '%E%f:%l: error: %m,%W%f:%l: warning: %m,%I%f:%l: note: %m'
   - exe: '/nix/store/3vwrd209k57ii4mn7rbc45ybkh9rsg8d-neovim-float/bin/nvim-python3'
   - version information (/nix/store/3vwrd209k57ii4mn7rbc45ybkh9rsg8d-neovim-float/bin/nvim-python3 --version): Python 3.7.2
NOTE: you can define g:neomake_python_enabled_makers to configure it (or b:neomake_python_enabled_makers).
```
I've tried running NeomakeClean to reset the makers to no avail.

### Steps to reproduce

I've tried to reproduce with this command:

`/nix/store/273i3m6i9w9am0wpbn4sycv3nld72flv-neovim/bin/nvim --cmd "let g:loaded_python_provider=1" --cmd "let g:python3_host_prog='/nix/store/3vwrd209k57ii4mn7rbc45ybkh9rsg8d-neovim-float/bin/nvim-python3'" -u minimal.vim --cmd "let g:python3_host_prog='python3'"`

but for some reasons NeomakeInfo didn't display the mypy maker which was showing the problem so I can't show you.

teto:
[...]

PR #2485, opened 2020-05-01
Introduce g:neomake_python_preferred for specifying py3/py2

Amar1729:
Neomake currently uses `python -V 2>&1` to determine the python version.

On systems where both `python -> python2` and `python3` executables exist in `$PATH`,  neomake will determine python 2.x as the current version (unless you have a virtualenv sourced, in which case the virtualenv's `python -> python3` shows up first in your `PATH`).

This primarily results in the optional checker `mypy` (used for checking python type annotations) think you are using python2. in python2, type annotations weren't supported by the syntax, so you get quite a few errors on each annotation you have in valid python3 code from `mypy`.

This is a dirty PR for my own sake that adds a global variable, `g:neomake_python_preferred`, to allow the user to just specify `python3`.

I've added comments that describe this code's edge cases - specifically, if the user sets `g:neomake_python_preferred` then that setting will override whichever python version is being used in a source virtualenv.

This is really the first time I've written anything real in vimscript so sorry if it's not great - if maintainers don't want to merge, hopefully it can at least serve as a reference for others who run into this admittedly niche problem.

### DIFF
--- a/autoload/neomake/makers/ft/python.vim
+++ b/autoload/neomake/makers/ft/python.vim
@@ -23,7 +23,35 @@ endfunction
 let neomake#makers#ft#python#project_root_files = ['setup.cfg', 'tox.ini']
 
 function! neomake#makers#ft#python#DetectPythonVersion() abort
-    let output = neomake#compat#systemlist('python -V 2>&1')
+    " lets use this as the preferred python version, so we don't nec. have to
+    " be inside a virtualenv for proper syntax checking
+    " let g:neomake_python_preferred = 'python3'
+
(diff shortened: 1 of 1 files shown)


Amar1729:
Linking this older issue https://github.com/neomake/neomake/issues/2302

The solution posted there is not robust, since `sudo pip3 install`ing anything is discouraged (so `python3 -m mypy won't work).

A halfway solution that "works" is setting `let g:neomake_python_mypy_args = []` or manually setting that array to the flags set in `python.vim`, but those solutions either get rid of some functionality or create unnecessary extra work.

PR #2487, opened 2020-05-15
Fix errorformat of markdownlint

### LINKED ISSUE #2480: markdownlinter jumping to wrong file/position
Consider the following minimal `init.vim` file:
```viml
let $DOTVIM = expand('$HOME/.config/nvim')

set runtimepath+=$DOTVIM/bundle/repos/github.com/neomake/neomake
filetype plugin indent on

let g:neomake_open_list = 2
let g:neomake_markdown_enabled_makers = ['markdownlint']
```
Now, as in the GIF do:
1. Open nvim and edit a `foo.md` file containing:
```md
# Foo

0. Bar
```
2. Run `:Neomake`
3. When the location list opens switch to it and press `<CR>` on the candidate.

After 3. a new file `foo.md:3` opens. I instead expect the cursor to jump to the existing `foo.md` file on line `3`

Peek 2020-03-28 12-27

jandamm:
Some error messages of markdownlint contain also a character. This
caused the character to be interpreted as line and the line to be part
of the filename.

Example:
```
README.md:1:81 MD013/line-length Line length [Expected: 80; Actual: 129]
```

Closes #2480

### DIFF
--- a/autoload/neomake/makers/ft/markdown.vim
+++ b/autoload/neomake/makers/ft/markdown.vim
@@ -29,8 +29,10 @@ endfunction
 function! neomake#makers#ft#markdown#markdownlint() abort
     return {
                 \ 'errorformat':
-                \ '%f:%l:%c %m,%f: %l: %m,' .
-                \ '%f:%l %m,%f: %l: %m'
+                \ '%f:%l:%c %m,' .
+                \ '%f: %l: %c: %m,' .
+                \ '%f:%l %m,' .
+                \ '%f: %l: %m'
                 \ }
 endfunction
 
--- a/autoload/neomake/makers/ft/markdown.vim
+++ b/autoload/neomake/makers/ft/markdown.vim
@@ -28,7 +28,9 @@ endfunction
 
 function! neomake#makers#ft#markdown#markdownlint() abort
     return {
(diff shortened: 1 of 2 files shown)


petobens:
Thank you!. @blueyed can we merge this one?

blueyed:
When exploding the list it shows that there is a duplicate entry now, which should not be necessary:
```diff
--- i/autoload/neomake/makers/ft/markdown.vim
+++ w/autoload/neomake/makers/ft/markdown.vim
@@ -29,8 +29,10 @@ endfunction
 function! neomake#makers#ft#markdown#markdownlint() abort
     return {
                 \ 'errorformat':
-                \ '%f:%l:%c %m,%f: %l: %m,' .
-                \ '%f:%l %m,%f: %l: %m'
+                \   '%f:%l:%c %m,' .
+                \   '%f: %l: %m,' .
+                \   '%f:%l %m,' .
+                \   '%f: %l: %m'
                 \ }
 endfunction

```
I don't know how the order should go.  Some example output might be helpful, and could be added to a test then also.

jandamm:
My bad with the duplicate in the errorformat.
I'm not sure about the `%f: %l: %m` format. I haven't seen it before but there is probably a reason why it has been added to neomake.

I've also added a test case for markdownlint.

PR #2496, opened 2020-07-08
Add basic support for glslc

gustafla:
Hello. I've been recently learning Vulkan and I added basic support for ft/glsl using glslc. I'm not sure about portability (using /dev/null because glslc doesn't have a syntax check only -flag), but the maker seems to run ok on my machine. I also considered using glslangValidator instead but couldn't get any errorformat working for it.

### DIFF
--- a/autoload/neomake/makers/ft/glsl.vim
+++ b/autoload/neomake/makers/ft/glsl.vim
@@ -1,0 +1,15 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#glsl#EnabledMakers() abort
+    return executable('glslc') ? ['glslc'] : []
+endfunction
+
+function! neomake#makers#ft#glsl#glslc() abort
+    return {
+        \ 'args': ['-c', '-o', g:neomake#compat#dev_null],
+        \ 'errorformat':
+            \ '%f:%l: %trror: %m,' .
+            \ '%f:%l: %tarning: %m,' .
+            \ '%-G%s',
+        \ }
+endfunction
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/glsl.vim
```diff
@@ -0,0 +1,17 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#glsl#EnabledMakers() abort
+    return executable('glslc') ? ['glslc'] : []
+endfunction
```
blueyed, line 5: `executable` gets check automatically, so it could/should just return the default makers, i.e. `['glslc']`.

autoload/neomake/makers/ft/glsl.vim
```diff
@@ -0,0 +1,17 @@
...
+    return {
+        \ 'args': ['-c', '-o', g:neomake#compat#dev_null],
+        \ 'errorformat':
+            \ '%f:%l: %trror: %m,' .
+            \ '%f: %trror: %m,' .
+            \ '%f:%l: %tarning: %m,' .
+            \ '%f: %tarning: %m,' .
+            \ '%-G%s',
+        \ }
+endfunction
```
blueyed, line 17: Do you have some example output, which could be added as a test then?
For examples search for `NeomakeTestsGetMakerWithOutput` in `tests`, e.g. https://github.com/blueyed/neomake/blob/8b7816744ed53f32f460a50dc5431f5d13d081c7/tests/ft_python.vader#L118-L129.
Happy to help if you do not want to do it yourself.

gustafla:
Closing because it's been a long time and I don't use this any more.

PR #2501, opened 2020-08-26
Added MIX_ENV to avoid compilation in dev environment

Gigitsu:
Using the default env `dev` when calling `mix compile`, prevents phoenix live reload to detect changes and recompile code on the fly.

The same happen if you change code while running `iex -S mix`, if you try to do `recompile` in iex shell the output is alwasy `:noop`

This is resolved in `elixir_ls` by using `test` as `MIX_ENV`.

### DIFF
--- a/autoload/neomake/makers/ft/elixir.vim
+++ b/autoload/neomake/makers/ft/elixir.vim
@@ -54,7 +54,7 @@ endfunction
 
 function! neomake#makers#ft#elixir#mix() abort
     return {
-      \ 'exe' : 'mix',
+      \ 'exe' : 'MIX_ENV=test mix',
       \ 'args': ['compile', '--warnings-as-errors'],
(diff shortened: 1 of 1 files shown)


autoload/neomake/makers/ft/elixir.vim
```diff
@@ -54,7 +54,7 @@ endfunction
 
 function! neomake#makers#ft#elixir#mix() abort
     return {
-      \ 'exe' : 'mix',
+      \ 'exe' : 'MIX_ENV=test mix',
```
blueyed, line 57: I think this won't work on Windows at least.

Environment variables can be set for jobs for newer Vim/Neovim (https://github.com/neomake/neomake/blob/4bc07569e982004c67089c94b4ccd066c57d0d51/autoload/neomake.vim#L48).
Does `mix` have an option for this maybe?

PR #2505, opened 2020-09-19
Protobuf Buf Linter

Jumziey:
Added protobuf buf linter. Removed old prototool linter which is deprecated.

About the testing... I couldn't really figure out how to run the linter in the test suite to validate the output of it. That could be nice for getting warnings if buf managed to screw up its json format in the future (just looking at the keys). It seems many other ft_ tests uses some magic files i could not figure out at all. I would love to know whats going on there. 

I also never written any vimscript nor used Vader before, so feel free to critique. 

@blueyed why does the Makefile use your personal fork of Vader that has no update since 2014 what i can see? I can see also you've worked on junegunn:s vader version as late as this year... so yea it's puzzling, would love to hear the reason behind it :)


blueyed:
@Jumziey the Vader testing is using a specific branch (which has been updated/rebased).

Jumziey:
> @Jumziey the Vader testing is using a specific branch (which has been updated/rebased).

Oh lol, the github interface actually hid the display-source-with-exceptions branch and only showed way older branches before i clicked "show more stale branches". Any particular reasoning for not using master? Just curious. 

Hadn't said it yet, but i think you're doing an awesome job with neomake! Ty

Jumziey:
Anything i can do to help "push" this along? :)

tests/ft_proto.vader
```diff
@@ -0,0 +1,30 @@
...
+  \ }']}
+
+  let entries = neomake#makers#ft#proto#BufProcessOutput(testInput)
+  let entry = entries[0]
+  
+  AssertEqual filename, entry.filename
+  AssertEqual start_line, entry.lnum
+  AssertEqual start_column, entry.col
+  AssertEqual type, entry.type
+  AssertEqual message, entry.text
```
blueyed, line 30: Should use `NeomakeTestsGetMakerWithOutput` for the test.
See the existing tests, e.g. tests/ft_python.vader.
This tests the whole maker then, not just the (internal) implementation.

blueyed (changes requested):
Thanks!

Test should get improved.
I can do it myself, let me know.

tests/ft_proto.vader
Jumziey: Great feedback, thank you. I just looked into it and it seems reasonable. I'll sort something out during the holidays.

Jumziey:
Just ping me if there's anything i can do to move this along :)

Jumziey:
If you'd look at this, I'd like to implement one or two more linters as well, might even try and support you a bit with issues etc if i have the opportunity to get warm in my clothes,  I have had great use of this plugin and would be ashame to loose you since you maintained it to work so well :)

kevinschoon:
@Jumziey I tried your branch and it works fine. The Buf command `buf check lint` was deprecated and should be updated to `buf lint`.

Thank you!

Jumziey:
Oh awesome, i can update that, it just been so long i forgot about this :joy:

Jumziey:
I've lost intrest, closing

PR #2507, opened 2020-10-27
added implementation for elm

ArnaudParan:
Hi guys, the current implementation did not work for me but that version that I corrected works like a charm. Does it fit with your way of doing things?


blueyed:
Thanks, do you have (raw) example output where it was failing before with?
Would be happy to add this as a test then (to get full coverage).

ArnaudParan:
> Thanks, do you have (raw) example output where it was failing before with?
> Would be happy to add this as a test then (to get full coverage).

Hi, sorry that I took time to reply to you and thanks for the interest.

Well, first, now elm-make is part of elm/core, and thus we could add functionality for the new syntax "elm make" which is the standard way to use the maker now, but even replacing that part, it seems that elm make changed a lot since the last update in neovim (my elm version is 0.19.1), because I get just no errors with the implementation which is present now. For example, if I alias elm make with elm-make or even create an elm-make which calls elm make, I get absolutely no errors when I should get them. For example, a code as simple as

`
concat : String -> String
concat s1 s2 =
  s1 + s2

`

which should produce an error in elm produces no error with the current implementation in neomake

image


with my implementation, there are errors printed (which there should be)

image

However, the weakness of my implementation is that the new elm make version does not deal with files which have unusual names, so when neomakes creates a file because you didn't save yet, there is an error from elm make

`
['{"type":"error","path":null,"title":"UNEXPECTED FILE NAME","message":["I am having trouble with this file name:\n\n    ",{"bold":false,"underline":false,"color":"RED","string":"frontend/.Test.elm@neomake_103840_3.elm"},"\n\nI found it
in your /home/arnaud/Programming/graphene/cta/bouygues-app/frontend/\ndirectory which is good, but I expect all of the files in there to use the\nfollowing module naming convention:\n\n    +--------------+--------------------------------
----------------------------------------------+\n    | Module Name  | File Path                                                                    |\n    +--------------+-------------------------------------------------------------------
-----------+\n    | Main         | /home/arnaud/Programming/graphene/cta/bouygues-app/frontend/Main.elm         |\n    | HomePage     | /home/arnaud/Programming/graphene/cta/bouygues-app/frontend/HomePage.elm     |\n    | Http.Helpers |
/home/arnaud/Programming/graphene/cta/bouygues-app/frontend/Http/Helpers.elm |\n    +--------------+------------------------------------------------------------------------------+\n\nNotice that the names always start with capital letter
s! Can you make your file\nuse this naming convention?\n\n",{"bold":false,"underline":true,"color":null,"string":"Note"},": Having a strict naming convention like this makes it a lot easier to find\nthings in large projects. If you see a
 module imported, you know where to look\nfor the corresponding file every time!"]}']
`
basically, the problem here is that .Test.elm@neomake_103840_3.elm is not a valid filename, however, at least, with the version I implemented, neomake works when we save the elm file in vim. With the old version, that does not work as I shown.

I don't know how you handle backwards compatibility, so as I did not want to break it, I non invasively added a maker instead of replacing the one which is already there.

I hope that I have been helpful. If what I say is unclear, I would be glad to add the details you ask me, but you can just try with the example of code I gave you, get the new elm, you can get it there [elm install page](https://guide.elm-lang.org/install/elm.html) and try the old neomake version and my version, you should be able to reproduce all the things said there

PR #2514, opened 2020-11-25
Sorbet linter

taq:
This is the Sorbet linter. To make sure it will not conflict with the Rubocop linter result, is a good idea install [rubocop-sorbet](https://github.com/Shopify/rubocop-sorbet).

### DIFF
--- a/autoload/neomake/makers/ft/ruby.vim
+++ b/autoload/neomake/makers/ft/ruby.vim
@@ -1,7 +1,7 @@
 " vim: ts=4 sw=4 et
 
 function! neomake#makers#ft#ruby#EnabledMakers() abort
-    return ['flog', 'mri', 'rubocop', 'reek', 'rubylint']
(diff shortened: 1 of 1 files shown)


blueyed:
> To make sure it will not conflict with the Rubocop linter result, is a good idea install [rubocop-sorbet](https://github.com/Shopify/rubocop-sorbet).

Should it maybe then not get used by default (if installed)?

Also it would be nice to have some tests with example output - please consider at least providing it here, I can look at adding the test then.

taq:
Hey. Yes, being optional is a good idea. Ruby 3 will allow to use type checking, but not everybody will use that.
What output you need? Some sample files with the Sorbet output as text?

blueyed:
> What output you need? Some sample files with the Sorbet output as text?

The raw ouput from some test/lint run would be enough - it does not need the input files, but would be mocked.
This is mainly to test the errorformat parsing/handling.

PR #2516, opened 2020-12-16
Add clj-kondo support

SevereOverfl0w:
I've gotten most of the way there, but I'm having some issues with stdin
support.  It seems that %t is  not being replaced with `-`, but instead
it is the full path to the file.

I had considered just doing

```vim
self.args = ['--filename', '%'] + self.args
```

But that's not done by any other maker, so I thought I should check
first.

Is there a way to do use these two features together that I've missed?

### DIFF
--- a/autoload/neomake/makers/ft/clojure.vim
+++ b/autoload/neomake/makers/ft/clojure.vim
@@ -1,0 +1,23 @@
+" vim: ts=4 sw=4 et
+
+function! neomake#makers#ft#clojure#EnabledMakers() abort
+    return ['clj_kondo']
+endfunction
+
+function! neomake#makers#ft#clojure#clj_kondo() abort
+    let maker = {
+        \ 'exe': 'clj-kondo',
+        \ 'args': ['--lint'],
+        \ 'errorformat':
+            \ '%f:%l:%c: Parse %t%*[^:]: %m,'.
+            \ '%f:%l:%c: %t%*[^:]: %m,'.
+            \ '%-Glinting took %.%#'
+        \ }
+
+    function! maker.supports_stdin(_jobinfo) abort
+        let self.args = ['--filename', '%'] + self.args
+        return 1
+    endfunction
+
+    return maker
+endfunction
--- a/tests/ft_clojure.vader
+++ b/tests/ft_clojure.vader
@@ -1,0 +1,29 @@
+Include: include/setup.vader
+
+Execute (clojure: clj_kondo: errorformat):
+  let maker = NeomakeTestsGetMakerWithOutput(neomake#makers#ft#clojure#clj_kondo(), [
+  \ 'file1.clj:2:1: error: unresolved symbol unknown',
+  \ 'linting took 12ms, errors: 1, warnings: 0',
+  \ ])
+  let maker.name = 'clojure'
+  new
+  file file1.clj
+  CallNeomake 1, [maker]
+  AssertEqualQf getloclist(0), [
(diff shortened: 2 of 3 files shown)


autoload/neomake/makers/ft/clojure.vim
```diff
@@ -0,0 +1,27 @@
...
+        \ 'append_file': 0,
+        \ 'errorformat': '%f:%l:%c: Parse %t%*[^:]: %m,%f:%l:%c: %t%*[^:]: %m',
+        \ 'output_stream': 'stdout',
+        \ }
+
+    function! maker.supports_stdin(_jobinfo) abort
+        let self.args += ['--filename', '%:p']
+        let self.tempfile_name = '-'
+        return 1
+    endfunction
```
blueyed, line 20: `supports_stdin` is not used with `append_file=0`.
(not sure if this is good, and should get documented at least probably).

I've pushed some suggested change, and also added tests - feel free to extend/revisit this/them yourself, of course.. :)
I suggest using `'args': ['--lint', '-', ]`

blueyed (commented):
Revisited and added tests.

autoload/neomake/makers/ft/clojure.vim
```diff
@@ -0,0 +1,23 @@
...
+endfunction
+
+function! neomake#makers#ft#clojure#clj_kondo() abort
+    let maker = {
+        \ 'exe': 'clj-kondo',
+        \ 'args': ['--lint'],
+        \ 'errorformat':
+            \ '%f:%l:%c: Parse %t%*[^:]: %m,'.
+            \ '%f:%l:%c: %t%*[^:]: %m,'.
+            \ '%-Glinting took %.%#'
```
blueyed, line 14: This was just reformatted, and I've added the last item to ignore the timing information - or do you think it is useful?

SevereOverfl0w: It's not that useful, no.  I just didn't know an easy way to strip it, I was worried I'd have to run it through a vimscript function or something to detect it.  Your solution looks good.

autoload/neomake/makers/ft/clojure.vim
SevereOverfl0w: > supports_stdin is not used with append_file=0.

This makes a lot of sense, spent a good bit of time banging my head on this :).  Docs would have been helpful!

>  I suggest using 'args': ['--lint', '-', ]

I was worried about this, because the user might disable stdin, or the file might be linted directly as it's been saved?  I was trying to support use with temp files simultaneously with stdin as I thought that was a thing.  If not, then I guess we could forgo the function and just do `'args': ['--lint', '-', '--filename', '%']`?

SevereOverfl0w:
All your changes look good to me.  I think it's good as is and supports both stdin & tempfiles.  I can't find any direct docs on tweaking args, so I'm assuming that's an "advanced" feature and you're expected to maintain some semblance of order if you poke at that stuff.

Marking as no longer a draft.

autoload/neomake/makers/ft/clojure.vim
blueyed: doc update: https://github.com/neomake/neomake/pull/2519

As for `args': ['--lint', '-', ]` this PR uses a different approach now, which should be fine(r).

blueyed:
Thanks, merging.
Please add the maker to https://github.com/neomake/neomake/wiki/Makers.

Issue #2515, opened 2020-11-27
No maker should be enabled by default.


ByLiZhao:
Some makers are enabled by default in Noemake. This means they will be running even if the user does not expect it. Some of these linters are buggy, polluting the user's location lists with false positive of warnings, even error messages from the linters themselves. The user is often left no option but modifying his `vimrc` file to explicitly disable them.  Forcing the user to disable something he may have not heard of doesn't seem right.

blueyed:
Good point.
However, changing this is not trivial, given that often users are required to enable makers then when it worked out of the box before.
The general idea is that installed makers should be used if it seems sensible by default.

But I can see that especially with C files this is not the case, and often you get warnings from clang when headers are not configured / picked up properly etc.
Do you have specific filetypes/makers where this is problematic for you?

ByLiZhao:
> Do you have specific filetypes/makers where this is problematic for you?

Yes. The other day when I opened a latex file with vim, the maker called `lacheck` spit dozens of warnings  into the location list. None of these warnings is helpful. I didn't know that I installed `lacheck` at the time, because it is installed as part of `TexLive` installation.  I have been using `TexLive` and `Neomake` for a long time, it was the first time that I noticed the existence of `lacheck`.

blueyed:
Do you have some example output for what is not considered useful?

`lacheck` was added in 2ee196220.  In 4b38e7366 only either chktex or lacheck would be used, but it was changed in 0e4d2c022 already then (likely due to a misunderstanding in https://github.com/neomake/neomake/pull/50#discussion_r23389282, #50).
/cc @natemara what do you think?

nate-onesignal:
I have not used this tool in years.

blueyed:
@nate-onesignal I see, sorry for the noise then - maybe you can still say something about using `lacheck` by default or not, and how useful it is in general?

ChasingZenith:
As for latex, I agree that no makers should be enabled by default.

Many issues have been raised before for latex filetype like #243 [#3470 in SpaceVim](https://github.com/SpaceVim/SpaceVim/issues/3470#issuecomment-660664167)

I think `lacheck` and `chtex` output too many irritating errors and nobody might need it unless some very good filters were used. Since `lacheck` and `chtex` are just some checkers for latex files, they do not compile the latex source file. They are just software that analyzes it, so they make many mistakes. Even if it output an `error`, the compilation by `latexmk` can succeed (in 99% cases).

[...]

PR #2520, opened 2020-12-20
makers: tex: do not enable "chktex" and "lacheck" by default

blueyed:
Ref: https://github.com/neomake/neomake/issues/2515#issuecomment-748581155


Big-Tree:
I just had this issue and also agree that those LaTeX makers should be disabled by default.  Anyone who writes LaTeX in vim likely uses [vimtex](https://github.com/lervag/vimtex) instead.

Issue #1971, opened 2018-06-08
Neomake javac maker not setting classpath for linting in gradle project


shoemakerdr:
### Expected behavior

Opening/saving a Java file that uses Gradle should correctly set classpath for linting/checking.

### Steps to reproduce

[This is the repo](https://github.com/shoemakerdr/tictactoe-java) I'm working with.

When I open one of the java files and try to save it or run `:Neomake`, I get a ton of `symbol not found` errors even though it works when I run gradle commands.

I'm also using [javacomplete2](https://github.com/artur-shaik/vim-javacomplete2), which is able to get the classpath fine.

Here's the relevant parts from the `neomake_logfile`:

```
14:46:36 [D      ] [-.-:1:1] automake: registered events: BufWritePost.
14:46:43 [D +7.23] [1.-:4:1] Running makers: javac (auto).
14:46:43 [D      ] Defining sign: sign define neomake_file_err texthl=error text=✖.
14:46:43 [D      ] Defining sign: sign define neomake_project_err texthl=error text=✖.
14:46:43 [D      ] Defining sign: sign define neomake_file_warn texthl=todo text=⚠.
14:46:43 [D      ] Defining sign: sign define neomake_project_warn texthl=todo text=⚠.
14:46:43 [D      ] Defining sign: sign define neomake_file_info texthl=todo text=ⓘ.
14:46:43 [D      ] Defining sign: sign define neomake_project_info texthl=todo text=ⓘ.
14:46:43 [D      ] Defining sign: sign define neomake_file_msg texthl=NeomakeMessageSign text=➤.
14:46:43 [D      ] Defining sign: sign define neomake_project_msg texthl=NeomakeMessageSign text=➤.
14:46:43 [D      ] [1.1:4:1] Setting neomake_automake_tick.
14:46:43 [V      ] [1.1:4:1] Starting async job: javac -Xlint -d . -d build/intermediates/classes/debug /Users/shoemakerdr/dev/java/tictactoe-java/src/main/java/tictactoe/Board.java.
14:46:43 [D      ] [1.1:4:1] cwd: /Users/shoemakerdr/dev/java/tictactoe-java.
14:46:43 [D      ] [1.1:4:1] Vim job: {'status': 'run', 'stoponexit': 'term', 'tty_out': '', 'exitval': 0, 'exit_cb': 0, 'tty_in': '', 'channel': 'channel 3 open', 'process': 22319}.
[...]

PR #2525, opened 2021-04-05
Gradle kotlin dsl support

MrcJkb:
This should add Kotlin DSL support for Gradle (by searching for `build.gradle.kts` files if `build.gradle` files are not found.

Before merging, please note that I have not been able to get Gradle autodetection working yet (neither on your `master` branch, nor on my `gradle-kotlin-dsl-support` branch), for projects using the Groovy DSL and the Kotlin DSL, respectively. This may be a bug, or I am missing something in the configuration of the plugin.

### DIFF
--- a/autoload/neomake/makers/ft/java.vim
+++ b/autoload/neomake/makers/ft/java.vim
@@ -321,7 +321,7 @@ endfunction " }}}2
 function! s:GradleOutputDirectory() abort
     let gradle_build_file_name = 'build.gradle'
     let gradle_build = s:findFileInParent(gradle_build_file_name, expand('%:p:h', 1))
-    if !filereadable(gradle)
+    if !filereadable(gradle_build)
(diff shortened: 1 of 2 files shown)


MrcJkb:
I noticed the gradle task in classpath.gradle is only compatible with very old gradle versions (prior to version 5.0.0). I updated it so that it is compatible with the latest version (6.8.3 as of writing).

It still doesn't recognize my gradle projects. I will have to implement look into this at another time.
Also, some regression tests may be necessary. Where can I find the end-to-end tests for the gradle project autodiscovery?

MrcJkb:
This may fix #1971 when completed.

MrcJkb:
I just noticed, the `Neomake!` command autocompletes to `Neomake! gradle` (and a bunch of others), ,but the `Neomake` command only autocompletes to `Neomake javac`.
 However, the `Neomake! gradle` command appears to fail quietly.

PR #2530, opened 2021-06-05
added maker for svelte

dbarrosop:
Enable `eslint` and `svelte-check` linters for [svelte](https://www.svelte.dev)

### DIFF
--- a/autoload/neomake/makers/ft/svelte.vim
+++ b/autoload/neomake/makers/ft/svelte.vim
@@ -1,13 +1,6 @@
 " vim: ts=4 sw=4 et
 function! neomake#makers#ft#svelte#EnabledMakers() abort
-    let makers = []
-    if executable('svelte-check')
-        call add(makers, 'svelte_check')
-    endif
-    if executable('eslint')
-        call add(makers, 'eslint')
-    endif
-    return makers
+    return ['svelte-check', 'eslint']
 endfunction
 
 function! neomake#makers#ft#svelte#svelte_check() abort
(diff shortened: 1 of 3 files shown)


autoload/neomake/makers/ft/svelte.vim
```diff
@@ -0,0 +1,28 @@
+" vim: ts=4 sw=4 et
+function! neomake#makers#ft#svelte#EnabledMakers() abort
+    let makers = []
+    if executable('svelte-check')
+        call add(makers, 'svelte_check')
+    endif
+    if executable('eslint')
+        call add(makers, 'eslint')
+    endif
```
blueyed, line 9: 1. Note that the executable check (based on `maker.exe`) gets done automatically, so this could be removed here.

2. Note also that it might not be wanted that makers get enabled/used by default (https://github.com/neomake/neomake/issues/2515).  I think it makes sense to have them here enabled by default though.

blueyed (commented):
Do you have some example output from svelte-check that could be added/used with a test then?

dbarrosop:
> Do you have some example output from svelte-check that could be added/used with a test then?

```
1632317135621 START "/Users/dbarroso/workspace/example-repo"
1632317136952 ERROR "src/lib/SomeComponent/index.svelte" 2:11 "'\"svelte\"' has no exported member named 'nMount'. Did you mean 'onMount'?"
1632317136952 ERROR "src/lib/SomeComponent/index.svelte" 8:2 "Cannot find name 'onMount'. Did you mean 'nMount'?"
1632317136952 COMPLETED 10 FILES 2 ERRORS 0 WARNINGS 1 HINTS
```

blueyed:
I've added some test, and fixed the returning of makers, which should be the name, not the exe.
Please check if it still works for you.
(I've merged master to make the test simpler, and retriggered the build for flaky tests)

dbarrosop:
yeah, it works fine. Thanks for the help here :)

PR #2542, opened 2021-10-12
Replace " with vimscript '' to avoid escape issues

StaticRocket:
Stumbled upon a weird issue with double quotes being ignored thanks to my stupid PowerShell environment at work.

I knew vimscript uses `"` for comments but apparently this has odd effects when used in a string as reported here:
https://stackoverflow.com/questions/24164365/vimscript-quotes-in-strings-when-used-as-expressions

Because the pylint maker uses a message template that contains characters that will be misinterpreted by PowerShell if not escaped correctly, when you change vim's default shell to PowerShell you can see that the `"`'s from
https://github.com/neomake/neomake/blob/053d6bd3605a64f05de11a179cbab423fe759ebe/autoload/neomake/makers/ft/python.vim#L70
are actually ignored and not passed through to the command:
```powershell
Neomake: pylint: unexpected output on stderr: pylint.exe : The command parameter was already specified.
At line:1 char:1
+ pylint --output-format=text --msg-template={path}:{line}:{column}:{C} ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [], ParameterBindingException
    + FullyQualifiedErrorId : ParameterSpecifiedAlready
```

It seems that changing all instances of `"` in command strings to `''` is more reliable than escaping with `\"`

---

The direction of this PR changed from a one off bug to an analysis of arg escape methods. This new mechanism should work for all OSs with the caveat that all arg strings that have `%f` substitutions will be wrapped in double quotes and all backslashes in the file path will be escaped.

### DIFF
--- a/autoload/neomake/compat.vim
+++ b/autoload/neomake/compat.vim
@@ -203,7 +203,7 @@ if neomake#utils#IsRunningWindows()
                 return argv
             endif
         endif
-        if &shell == 'powershell'
+        if &shell ==? 'powershell'
             return prefix.'{'.argv.'}'
         endif
         return prefix.argv
--- a/autoload/neomake/compat.vim
+++ b/autoload/neomake/compat.vim
@@ -203,6 +203,9 @@ if neomake#utils#IsRunningWindows()
                 return argv
             endif
         endif
+        if &shell == 'powershell'
+            return prefix.'{'.argv.'}'
+        endif
         return prefix.argv
     endfunction
 elseif has('nvim')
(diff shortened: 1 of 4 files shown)


blueyed:
I've not noticed that tests were running actually before already.
In general I think the change is fine, but it appears to be also a good test case / trigger for broken arg handling, which should become its own issue/PR maybe then.
I'll take you a look at/for other similar places that could be added here.  Let me know when you think this can be merged.

StaticRocket:
I'm closing this for now. I'll get you a proper fix Soon™

PR #2547, opened 2022-01-29
makers/nix: adapt error format to nix >= 2.4

symphorien:
Example error message
error: undefined variable 'bar'

       at /tmp/foo.nix:2:1:

            1| { foo }:
            2| bar
             | ^
            3|


blueyed:
Thanks!

PR #2563, opened 2024-04-01
feat(D): add D-Scanner

FelipeLema:


FelipeLema:
I don't know how to assemble the tests correctly... can I get some pointers?

FelipeLema:
got tests assembled ... but only passing in some environments? I'm defo lost now

FelipeLema:
how does one handle the `length` entry? I have all data to provide "end line" and "end column", but I dunno how to do it _across_ vim-and-neovim versions

